PR #7, opened 2024-06-06
feature: Added support for user parameters in sync rules.

Chriztiaan:
# Description
This PR introduces the ability to pass parameters from a client SDK to the sync rules.
Parameters are available in the sync rules under `user_parameters` (only supported `token_parameters` until now). Client PR [here](https://github.com/powersync-ja/powersync-js/pull/204) 

For example:
```
bucket_definitions:
  user_lists:
    parameters: select id as list_id from lists where owner_id = user_parameters.client_variable_test
    data:
      - SELECT * FROM lists where id = bucket.list_id
```

# Testing
Verified values pull through via self-host project.

### DIFF
--- a/packages/service-core/src/util/protocol-types.ts
+++ b/packages/service-core/src/util/protocol-types.ts
@@ -89,7 +89,12 @@ export const StreamingSyncRequest = t.object({
   /**
    * Data is received in a serialized BSON Buffer
    */
-  binary_data: t.boolean.optional()
+  binary_data: t.boolean.optional(),
+
+  /**
+   * Client parameters to be passed to the sync rules.
+   */
+  parameters: t.record(t.string).optional()
 });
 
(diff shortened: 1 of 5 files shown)


packages/service-core/src/util/protocol-types.ts
```diff
@@ -89,7 +89,12 @@ export const StreamingSyncRequest = t.object({
   /**
    * Data is received in a serialized BSON Buffer
    */
-  binary_data: t.boolean.optional()
+  binary_data: t.boolean.optional(),
+
+  /**
+   * Client parameters to be passed to the sync rules.
+   */
+  parameters: t.record(t.string).optional()
```
rkistner, line 97: Should probably allow arbitrary values here (`t.any`?)

Chriztiaan: Updated 👍

stevensJourney (approved):
LGTM

rkistner (approved):

PR #11, opened 2024-06-10
[Fix] WebSocket Close Bugs

stevensJourney:
# Overview

Recent alerts have shown uncaught exceptions for clients using the WebSockets connection method. Two main issues were identified.

## Write to closed WebSocket

In some circumstances the WebSocket might close unexpectedly. If the server tried to write data to this port an unhandled exception would be thrown. 

```
WebSocket is not open: readyState 2 (CLOSING)
```

An investigation led to this issue https://github.com/websockets/ws/issues/1515.

This PR adds better error and closed socket handling to the `WebSocketDuplexConnection` and `WebSocketServerTransport` implementations provided by RSocket. 

A unit test has been added to ensure no unhandled exceptions are thrown when the client closes the WebSocket. 

## Undefined frames
Some reports of the server failing to handle WebSocket frames were received.

image

This was reproducible by sending invalid data (not a valid RSocket frame buffer) over a WebSocket. Although not verified: this might be caused by a broken `Buffer` implementation on the client side which could happen in environments such as React Native. 

Additional guards have been added to the `WebSocketDuplexConnection` frame handling. Unit tests have also been added for invalid frames.

## Double close warning

Previously a double close warning for the RSocket server could be seen in logs. Currently we use a Fastify server which a WebSocket server manages `upgrade` events on. The standard RSocket `WebSocketServerTransport` registers a listener for the `close` event of the WebSocketServer and calls the `ServerCloseable` `close` wherein the server is again closed. This behaviour has been removed as the WebSocket server's closing is tied to the Fastify server's `close` event. 

## Additional Fixes

This PR includes some additional bug fixes:

- Included TypeScript `path` references for the `server` project.
- Added authentication errors to WebSocket responses 
- Added git ignored `.probes` folder for `service` folder

### DIFF
--- a/.changeset/happy-flies-tease.md
+++ b/.changeset/happy-flies-tease.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-core': patch
+---
+
+Fix missing authentication errors for websocket sync stream requests
--- a/.changeset/smooth-frogs-wait.md
+++ b/.changeset/smooth-frogs-wait.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-rsocket-router': patch
+---
+
+Fix issue where sending data during socket closing would throw an exception.
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,8 @@ dist
 npm-error.log
 .pnpm-debug.log
 .local-dev
+.probes
+
 
 packages/*/manifest.json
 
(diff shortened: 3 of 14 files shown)


rkistner:
Is it feasible to also get the changes upstream in `rsocket-websocket-server`, or will we have to continue maintaining forked code?

stevensJourney:
> Is it feasible to also get the changes upstream in `rsocket-websocket-server`, or will we have to continue maintaining forked code?

I can make a PR on their repository with these changes. I think making the changes here would be the fastest solution in the interim.

rkistner (approved):

PR #12, opened 2024-06-10
Incrementally calculate checksums per connection & include empty buckets

rkistner:
## Incremental checksums

Bucket operations are always additive (operations are only added at the end of the bucket, never removed or modified). This makes it easy to calculate checksums incrementally.

This changes the checksum calculation to be incremental per connection - we only get the checksum for operations since the last checkpoint from the database, instead of recalculating checksums for the entire checkpoint on every change. This doesn't reduce the latency on initial connection, but does significantly reduce the latency and database overhead for any changes processed after that.

Future improvements could cache checksums across different connections and/or users, in memory or in database storage.

## Empty buckets

This now also returns zero checksums for empty buckets to the client, instead of omitting the bucket. This does not affect synced data on the client, but helps with diagnosing sync issues.

### DIFF
--- a/packages/service-core/src/storage/BucketStorage.ts
+++ b/packages/service-core/src/storage/BucketStorage.ts
@@ -230,7 +230,16 @@ export interface SyncRulesBucketStorage {
     options?: BucketDataBatchOptions
   ): AsyncIterable<util.SyncBucketData>;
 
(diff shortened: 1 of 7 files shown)


stevensJourney (approved):

rkistner:
Superseded by #17.

PR #30, opened 2024-07-03
Expressions on parameters

rkistner:
This is an initial refactoring to allow some types of expressions on token_parameters and user_parameters.

Examples that this now supports (see tests for more):

```sql
SELECT id from users WHERE filter_param = token_parameters.some_param ->> 'description'
SELECT id from users WHERE filter_param = upper(token_parameters.some_param)
SELECT id from projects WHERE page = user_parameters.index + 1
```

These expressions are only supported for parameter queries, not for bucket parameters in data queries.

Additionally this:
 * Expands test coverage for sync rules.
 * Refactors internal naming to make concepts easier to follow.
 * Adds overview documentation on the implementation.

### DIFF
--- a/.changeset/cyan-doors-tap.md
+++ b/.changeset/cyan-doors-tap.md
@@ -1,0 +1,7 @@
+---
+'@powersync/service-sync-rules': minor
+'@powersync/service-core': patch
+'powersync-open-service': patch
+---
+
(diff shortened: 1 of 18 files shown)


Chriztiaan:
The sync rule readme helps a lot!

Chriztiaan (approved):

PR #36, opened 2024-07-10
Fix concurrency limit for websockets

rkistner:
#24 introduced concurrency limits for websockets, similar to what we had with http streams. However, the connection count does not always decrease when a connection is closed, leading to all clients eventually getting an error " [429] Maximum active concurrent connections limit has been reached".

This simplifies it a bit by just tracking the number of websocket clients on the websocket server, instead of attempting to track the individual streams. We could additionally limit the number of streams per client in the future.

### DIFF
--- a/packages/rsocket-router/src/router/ReactiveSocketRouter.ts
+++ b/packages/rsocket-router/src/router/ReactiveSocketRouter.ts
@@ -19,11 +19,7 @@ import { WebsocketServerTransport } from './transport/WebSocketServerTransport.j
 import { errors, logger } from '@powersync/lib-services-framework';
 
 export class ReactiveSocketRouter<C> {
-  protected activeConnections: number;
-
-  constructor(protected options?: ReactiveSocketRouterOptions<C>) {
-    this.activeConnections = 0;
-  }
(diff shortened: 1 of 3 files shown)


packages/rsocket-router/src/router/ReactiveSocketRouter.ts
```diff
@@ -60,11 +56,16 @@ export class ReactiveSocketRouter<C> {
       acceptor: {
         accept: async (payload) => {
           const { max_concurrent_connections } = this.options ?? {};
-          if (max_concurrent_connections && this.activeConnections >= max_concurrent_connections) {
-            throw new errors.JourneyError({
-              code: '429',
+          // wss.clients.size includes this connection, so we check for greater than
+          // TODO: Share connection limit between this and http stream connections
+          if (max_concurrent_connections && wss.clients.size > max_concurrent_connections) {
```
rkistner, line 61: Docs for `wss.clients`: https://github.com/powersync-ja/powersync-service/pull/36/files

The size of the set did seem to accurately reflect the number of concurrent connections in the testing I did.

stevensJourney (approved):

PR #37, opened 2024-07-11
Compacting buckets

rkistner:
Implement a `compact` command that compacts all buckets in the instance. See the new documentation for what exactly it does.

This is currently a separate manual command, and not run automatically. It can be run manually or on a schedule.

This does not implement defragmentation yet. For some use cases it may be beneficial to trigger defragmentation from the postgres side using a periodic query such as `update mytable set id = id where updated_at < now() - interval '1 week'`, after which the compact command will be able to efficiently compact historical operations.

The memory used by the command can be controlled using the `--max-old-space-size` config for NodeJS. 256MB should be enough for up to 1M unique rows per bucket.

This is the first implementation using MOVE and CLEAR operations. These operations have been defined in the protocol and implemented in all clients since the start, but we'll need proper integration testing for consistency.

This also adds:
1. Overview documentation for the protocol.
2. Documentation on formal bucket properties guaranteed by the protocol.
3. Tests validating the above properties for buckets and compacted buckets.

### DIFF
--- a/docs/sync-protocol.md
+++ b/docs/sync-protocol.md
@@ -151,6 +151,8 @@ Format:
 }
 ```
 
+A previous iteration of the protocol included an optional `data: '{"target": OpId}'` field. This was to indicate a checkpoint is not valid unless its op_id is greater than or equal to the target. Now, the server is responsible for invalidating checkpoints in those cases, by not sending a `CheckpointComplete` in that case.
+
 ## CLEAR
 
 Indicates a "reset point" of a bucket. All operations prior to this bucket must be removed.
--- a/docs/sync-protocol.md
+++ b/docs/sync-protocol.md
@@ -1,0 +1,170 @@
+# Sync Protocol
+
+This document is a work in progress - more details to be added over time.
+
+# Sync stream
+
+A sync stream contains 5 possible messages:
+
+## StreamingSyncCheckpoint
+
+Format:
+
+```ts
+{
+  checkpoint: {
+    last_op_id: OpId;
+    write_checkpoint?: OpId;
+    buckets: BucketChecksum[];
+  }
+}
+```
+
+This indicates a new checkpoint is available, along with checksums for each bucket in the checkpoint.
+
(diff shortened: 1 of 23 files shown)


docs/sync-protocol.md
```diff
@@ -0,0 +1,172 @@
...
+}
+```
+
+This indicates a new checkpoint is available, along with checksums for each bucket in the checkpoint.
+
+This is typically (but not necessarily) be the first message in the response, and is often followed by StreamingSyncData and StreamingSyncCheckpointComplete (both optional).
+
+## StreamingSyncCheckpointDiff
+
+This has the same conceptual meaning as a StreamingSyncCheckpoint. It is an optimization to only sent details for buckets that changed, instead of sending the entire checkpoint over.
```
stevensJourney, line 29:
```suggestion
This has the same conceptual meaning as a StreamingSyncCheckpoint. It is an optimization to only send details for buckets that changed, instead of sending the entire checkpoint over.
```

packages/service-core/src/storage/mongo/MongoCompactor.ts
```diff
@@ -0,0 +1,356 @@
...
+  constructor(private db: PowerSyncMongo, private group_id: number, options?: MongoCompactOptions) {
+    this.idLimitBytes = (options?.memoryLimitMB ?? DEFAULT_MEMORY_LIMIT_MB) * 1024 * 1024;
+    this.moveBatchLimit = options?.moveBatchLimit ?? DEFAULT_MOVE_BATCH_LIMIT;
+    this.moveBatchQueryLimit = options?.moveBatchQueryLimit ?? DEFAULT_MOVE_BATCH_QUERY_LIMIT;
+    this.clearBatchLimit = options?.clearBatchLimit ?? DEFAULT_CLEAR_BATCH_LIMIT;
+    this.maxOpId = options?.maxOpId;
+  }
+
+  /**
+   * Compact buckets by converting operatoins into MOVE and/or CLEAR operations.
```
stevensJourney, line 68: Nit: typo in `operatoins`

stevensJourney (approved):
This deserves more than a "looks good to me". I couldn't see any issues from my side.

PR #38, opened 2024-07-12
[Chore] Automatically trigger DockerHub release

stevensJourney:
# Overview

The release process for the open edition DockerHub image is currently manually triggering the `Docker Image Release` action after the Changesets `Version` PR has been merged. It's easy to forget to trigger this action. 

The Docker image is built (and versioned) from the `powersync-open-service` private package (located in the `service` folder).
This PR configures Changesets to tag the `powersync-open-service` (private) "package" when publishing. The publish action will then run after tags have been pushed. [Refference](https://github.com/changesets/changesets/blob/main/docs/versioning-apps.md)

This will also create Github releases for `powersync-open-service` which will notify watchers of the repo about Docker image releases.

Side note: Do we want to rename `powersync-open-service` to something else? Releases will have the title `powersync-open-service@[version]`.  We could potentially name it similar to the Docker image name: `@journeyapps/powersync-service` (note: the `@` is required in `package.json`)

For example it should look like this currently:
image

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,5 +7,9 @@
   "access": "restricted",
   "baseBranch": "origin/main",
   "updateInternalDependencies": "patch",
-  "ignore": []
+  "ignore": [],
+  "privatePackages": {
+    "tag": true,
+    "version": true
+  }
 }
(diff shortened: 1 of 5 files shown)


rkistner (approved):

benitav:
`@powersync/service-image` might make clearer what it's for? I also like something about matching the current image name, the @journeyapps just bothers me a little as it's inconsistent with our existing packages and I'm also thinking of potential future renames of the docker image, and then needing to keep these in sync. But don't have a particularly strong opinion here.

benitav (dismissed):

stevensJourney:
I've renamed the service private package to `@powersync/service-image`

benitav (approved):

PR #41, opened 2024-07-18
[Feature] Configurable metrics

stevensJourney:
# Description

## Metrics on Container

This moves the `Metrics` implementation from a class singleton to the framework `Container`. 

This allows for constructing different `Metrics` implementations which can be registered during project initialization. 

Code in the `@powersync/service-core` package can fetch the registered `Metrics` implementation easily without any massive refactoring - this approach is subject to change when a modules system is implemented. 

## Extendable Config

Extended Metrics implementations will require extended PowerSyncConfig which is collected by a `CompoundConfigCollector`. This class is now made extendable and generic. Downstream services can implement their own additional validation, decoding and resolving of config.


stevensJourney:
Closing this stale PR. superseded by https://github.com/powersync-ja/powersync-service/pull/221

PR #54, opened 2024-08-07
Add test client

rkistner:
Add an example NodeJS client that retrieves the raw operation data.

Apart from type definitions, the client uses minimal powersync dependencies, to attempt to be a standalone example.

See the readme for details.


stevensJourney (approved):
Looks good to me

PR #58, opened 2024-08-14
Add TableV2 / TypeScript schema generator

rkistner:
We have been recommending this syntax for a while - the generator should use this by default.

We can still give the option of the old syntax for apps not wanting to switch.

The generator here includes the entire "schema" file, including imports and exports.

### DIFF
--- a/packages/sync-rules/src/TsSchemaGenerator.ts
+++ b/packages/sync-rules/src/TsSchemaGenerator.ts
@@ -1,0 +1,50 @@
+import { ColumnDefinition, TYPE_INTEGER, TYPE_REAL, TYPE_TEXT } from './ExpressionType.js';
+import { SchemaGenerator } from './SchemaGenerator.js';
+import { SqlSyncRules } from './SqlSyncRules.js';
+import { SourceSchema } from './types.js';
+
+export class TsSchemaGenerator extends SchemaGenerator {
+  readonly key = 'ts';
+  readonly label = 'TypeScript';
+  readonly mediaType = 'application/typescript';
+  readonly fileName = 'schema.ts';
+
+  generate(source: SqlSyncRules, schema: SourceSchema): string {
+    const tables = super.getAllTables(source, schema);
+
+    return `import { column, Schema, TableV2 } from '@powersync/web';
+// OR: import { column, Schema, TableV2 } from '@powersync/react-native';
+
+${tables.map((table) => this.generateTable(table.name, table.columns)).join('\n\n')}
+
+export const AppSchema = new Schema({
+  ${tables.map((table) => table.name).join(',\n  ')}
+});
+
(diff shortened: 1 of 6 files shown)


DominicGBauer (dismissed):

packages/sync-rules/src/TsSchemaGenerator.ts
```diff
@@ -3,26 +3,53 @@ import { SchemaGenerator } from './SchemaGenerator.js';
 import { SqlSyncRules } from './SqlSyncRules.js';
 import { SourceSchema } from './types.js';
 
+export interface TsSchemaGeneratorOptions {
+  language?: 'ts' | 'js';
+  imports?: 'web' | 'react-native' | 'auto';
```
DominicGBauer, line 8: Is there a reason you're not using an enum?

packages/sync-rules/src/TsSchemaGenerator.ts
```diff
@@ -3,26 +3,53 @@ import { SchemaGenerator } from './SchemaGenerator.js';
...
+    if (lang == 'ts') {
+      return `export type Database = (typeof AppSchema)['types'];\n`;
+    } else {
+      return ``;
+    }
+  }
+
+  private generateImports() {
+    const importStyle = this.options.imports ?? 'auto';
+    if (importStyle == 'web') {
```
DominicGBauer, line 45: If we use enums we can also add some nice type safety.

I think a map here would be cleaner e.g.:
```
enum ImportStyle {
  Web = 'web',
  ReactNative = 'react-native',
  Auto = 'auto'
}
...

styleMap: Record<ImportStyle, string> = {
 [ImportStyle.Web]: `import { column, Schema, TableV2 } from '@powersync/web';`,
 [ImportStyle.ReactNative]: `import { column, Schema, TableV2 } from '@powersync/react-native';`
 [ImportStyle.Auto]: `import { column, Schema, TableV2 } from '@powersync/web';
// OR: import { column, Schema, TableV2 } from '@powersync/react-native';`
}

return styleMap[importStyle]
}
```

packages/sync-rules/src/TsSchemaGenerator.ts
rkistner: Laziness mostly

packages/sync-rules/src/TsSchemaGenerator.ts
rkistner: Note that string literal union types gives just as much type safety as enums, but I switched to proper enums here.

I'm still keeping the same style of if/else in the function - it's more flexible if we need to customize it more in the future.

DominicGBauer (dismissed):

DominicGBauer (approved):
Looks like a test failed

PR #64, opened 2024-08-22
Support client_id and User-Agent

rkistner:
1. Support client_id parameter.
    1. This specifically affects write checkpoints (limits write checkpoints to a specific client, rather than just an user).
    2. The id is generated by the client and not sure on its own, so we still scope per user_id.
2. Allow User-Agent and X-User-Agent headers.
    1. This allows us to identify which client is used.

### DIFF
--- a/packages/service-core/src/routes/endpoints/sync-stream.ts
+++ b/packages/service-core/src/routes/endpoints/sync-stream.ts
@@ -21,6 +21,9 @@ export const syncStreamed = routeDefinition({
   validator: schema.createTsCodecValidator(util.StreamingSyncRequest, { allowAdditional: true }),
   handler: async (payload) => {
     const system = payload.context.system;
+    const headers = payload.request.headers;
+    const userAgent = headers['x-user-agent'] ?? headers['user-agent'];
+    const clientId = payload.params.client_id;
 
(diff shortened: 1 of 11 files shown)


packages/service-core/src/routes/endpoints/sync-stream.ts
```diff
@@ -21,6 +21,9 @@ export const syncStreamed = routeDefinition({
   validator: schema.createTsCodecValidator(util.StreamingSyncRequest, { allowAdditional: true }),
   handler: async (payload) => {
     const system = payload.context.system;
+    const headers = payload.request.headers;
+    const userAgent = headers['x-user-agent'] ?? headers['user-agent'];
```
stevensJourney, line 25: Are there any limitations for adding this to the websocket method? I think we can access headers in the initial request or add this data to the request itself.

rkistner: I added this as a parameter to the setup metadata now.

stevensJourney: 👍

stevensJourney (approved):
Looks good to me

PR #49, opened 2024-07-25
[Chore] Neaten Routes

stevensJourney:
# Overview

The `@powersync/service-core` package currently exports various route definitions and some functions for registering the routes on a server. 

Routes are usually registered in the `server` runner inside the project `service` folder/package. The current configuration required duplication of the route registration between different projects. 

This PR introduces utility functions which perform the common route registration. The registration itself is still configurable if projects require custom functionality.

PR #65, opened 2024-08-26
[Fix] Add missing changeset

stevensJourney:
# Overview

https://github.com/powersync-ja/powersync-service/pull/49/files added a log entry for the number of currently connected WebSockets. This change is useful for debugging, but since no Changeset was added for this - no package version bump was done and the changes are not present downstream. The open edition has the change due to the Docker image copying the package from the repo source.


rkistner (approved):
Happy with the changeset.

I do want to add some additional logging before we roll out another release

PR #70, opened 2024-08-29
Fix `WebSocket is not open: readyState 2 (CLOSING)`

rkistner:
The issue was that the `removeAllListeners()` on close also removes the error handler. So if an error on the socket is triggered after closing, it ended up as an uncaught exception, crashing the process.

This only appears to happen when there is a combination of:
1. The socket is closed by the server, e.g. from a keepalive timeout.
2. A lot of data to be sent is buffered.

This may reduce the need for some of the workarounds implemented in https://github.com/powersync-ja/powersync-service/pull/11, although they probably don't hurt.

### DIFF
--- a/.changeset/tricky-parents-mate.md
+++ b/.changeset/tricky-parents-mate.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-rsocket-router': patch
+---
+
+Fix WebSocket is not open: readyState 2 (CLOSING)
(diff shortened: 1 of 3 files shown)


stevensJourney (approved):
Amazing find and fix!

Rentacookie:
Thank you Ralf 🙏

PR #77, opened 2024-09-04
More descriptive error message when SQL querying is not enabled

kobiebotha:


stevensJourney (commented):
This requires a Changeset in order for this to be released correctly. Could you please add a `patch` change for the `@powersync/service-core` package for this?

PR #90, opened 2024-09-25
Use generic schema definitions

rkistner:
This changes the schema provided over the API to primarily provide a _sqlite_ type, rather than a postgres type. It's more than just renaming the field - the conversion to the SQLite type now happens in the service (in the specific module), rather than the client.

Each column now primarily has:
1. `sqlite_type`: The type as it will behave in sync rules.
2. `internal_type`: The original type from the source database. This should not be used to check the type in code, but can be used for descriptions in auto-complete or generated schemas.

The old `pg_type` and `type` fields are still kept as is, to preserve backwards-compatibility. However, these do not need to be populated by other modules, and can be removed later.

This now also refactors internals to keep the original type (internal_type) available in the output schema after going through sync rules, allowing us to annotate the generated schema with comments like this:

```ts
const assets1 = new Table(
  {
    // id column (text) is automatically included
    name: column.text, // text
    count: column.integer, // int4
    owner_id: column.text // uuid
  },
  { indexes: {} }
);
```


modules/module-postgres/src/api/PostgresRouteAPIAdapter.ts
```diff
@@ -275,14 +275,14 @@ GROUP BY schemaname, tablename, quoted_name`
...
-    let schemas: Record<string, any> = {};
+    let schemas: Record<string, service_types.DatabaseSchema> = {};
 
     for (let row of rows) {
       const schema = (schemas[row.schemaname] ??= {
         name: row.schemaname,
         tables: []
       });
-      const table = {
+      const table: service_types.DatabaseSchema['tables'][0] = {
```
stevensJourney, line 285: Could we export this as a type instead of having to extract the type?

stevensJourney (approved):

stevensJourney (approved):

PR #108, opened 2024-10-17
New sync rule functions: json_keys and substring

rkistner:
The `substring` function roughly matches the [SQLite substring function](https://sqlite.org/lang_corefunc.html#substr) (notably different from the JavaScript equivalent). It is also similar to the Postgres substring function, but different when a negative `start` or `length` is specified.

The `json_keys` function has no direct equivalent in Postgres or SQLite. Postgres has a similar `json_object_keys`, but that returns each key as a separate row, while we return the keys as a JSON array.

Use cases:

substring:

```sql
-- Parameter query - filter by date value
-- substring(created_at, 1, 10) returns the date portion of the timestamp
select id as post_id from posts where substring(created_at, 1, 10) in request.jwt() -> 'sync_dates'
```

json_keys:
```sql
-- Data query - permissions_json: {[user_id]: 'read' | 'write' | 'admin'}
-- This is still limited - we can only operate on the keys, and not the values
select * from items where bucket.user_id in json_keys(permissions_json)

-- even better would be something like this, but that is significantly more complex to implement:
select * from items where (permissions_json ->> bucket.user_id) = 'read'
select * from items where 'read' in (permissions_json ->> bucket.user_id)
```

### DIFF
--- a/.changeset/thirty-llamas-carry.md
+++ b/.changeset/thirty-llamas-carry.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': minor
+---
+
+Support substring and json_keys functions in sync rules
--- a/packages/sync-rules/src/sql_functions.ts
+++ b/packages/sync-rules/src/sql_functions.ts
@@ -70,6 +70,61 @@ const lower: DocumentedSqlFunction = {
   detail: 'Convert text to lower case'
 };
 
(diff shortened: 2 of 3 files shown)


stevensJourney (approved):
This is awesome. It will help a lot with defining sync rules.

guillempuche:
Cooool!!!

PR #109, opened 2024-10-17
Watch Active Checkpoint

stevensJourney:
# Overview

This builds on-top of https://github.com/powersync-ja/powersync-service/pull/105. 

This adds the ability to set the storage's Write Checkpoint mode after the storage's initialisation. 

Setting the Write Checkpoint mode often depends on the active Sync rules. This adds a `watchActiveCheckpoint` method on the `BucketStorage` which should yield the active checkpoint and active Sync Rules.

This also adds plan parameters to the PowerSync config file.


rkistner:
The majority of the write checkpoint operations are specific to a specific set of sync rules (takes `sync_rules_id` somewhere in the parameters). Could we move those APIs to SyncRulesBucketStorage instead? (implemented in MongoSyncBucketStorage)

With that done, we could calculate the write checkpoint mode once when MongoSyncBucketStorage is instantiated, and don't need to watch for any changes.

packages/service-core/src/storage/MongoBucketStorage.ts
```diff
@@ -445,7 +457,8 @@ export class MongoBucketStorage
           operationType: 1,
           'fullDocument._id': 1,
           'fullDocument.last_checkpoint': 1,
-          'fullDocument.last_checkpoint_lsn': 1
+          'fullDocument.last_checkpoint_lsn': 1,
+          'fullDocument.content': 1
```
rkistner, line 461: This has the potential to add a lot of overhead - the rest of the change stream is optimized to get checkpoint updates with as little latency as feasible. It also shouldn't be needed - the content of a sync rules document never change.

stevensJourney:
> The majority of the write checkpoint operations are specific to a specific set of sync rules (takes `sync_rules_id` somewhere in the parameters). Could we move those APIs to SyncRulesBucketStorage instead? (implemented in MongoSyncBucketStorage)
> 
> With that done, we could calculate the write checkpoint mode once when MongoSyncBucketStorage is instantiated, and don't need to watch for any changes.

That does seem better. I've implemented that [here](https://github.com/powersync-ja/powersync-service/pull/110). I'll close this PR since it will no longer be required.

PR #110, opened 2024-10-21
[Modules] Move Write Checkpoint APIs

stevensJourney:
# Overview

This builds on-top of https://github.com/powersync-ja/powersync-service/pull/105. 

The Write Checkpoint mode can be dictated by the sync rules in some circumstances. 

This moves most of the Write Checkpoint APIs to the `SyncBucketStorage` interface. See [here](https://github.com/powersync-ja/powersync-service/pull/109#issuecomment-2426338728) for reference.

Moving the Write Checkpoint APIs to the  `SyncBucketStorage` allow the sync rules to be parsed and the write checkpoint mode to be correctly configured before Write Checkpoints are created (or attempted to be created).

This also adds plan parameters to the PowerSync config file.


rkistner (approved):
Thanks, this feels much cleaner now

PR #111, opened 2024-10-23
MySQL Alpha Support

Rentacookie:
### Alpha support for MySQL as a replication datasource

Replication for MySQL is handled by listening for changes on the [binlog](https://dev.mysql.com/doc/refman/8.4/en/binary-log.html) which is similar to Postgres' write ahead log.

We use a forked version of the Zongji MySQL Binlog listener library that has been customised to our needs:
https://github.com/powersync-ja/powersync-mysql-zongji

Demo app and backend is available here:
https://github.com/powersync-ja/self-host-demo/blob/main/demos/nodejs-mysql/README.md

Some additional configuration is necessary for your instance of MySQL to enable the binlog. A MySQL user should also be created with the required permissions for Powersync to perform the replication.
https://github.com/powersync-ja/self-host-demo/tree/main/services/mysql/init-scripts


modules/module-mysql/src/replication/zongji/zongji.d.ts
```diff
@@ -0,0 +1,119 @@
+declare module '@powersync/mysql-zongji' {
```
stevensJourney, line 1: Should we move this to our fork package?

packages/service-core/src/util/protocol-types.ts
```diff
@@ -88,7 +88,7 @@ export type StreamingSyncLine =
  */
 export type OpId = string;
 
-export interface Checkpoint {
+interface Checkpoint {
```
stevensJourney, line 91: Why this change?

modules/module-mysql/src/replication/zongji/zongji.d.ts
Rentacookie: I would be happy to - just not entirely sure what configuration is involved for that 😅 
I'm assuming this entails exporting it as the library's types or something?

stevensJourney: This file could be added to that package. It would require some minor modifications, like removing the `declare module ...` bit and exporting the individual declarations. We could then add an entry for `"types": "types/index.d.ts` to the package.json file. 
This is not a blocker for now though.

stevensJourney (approved):
Looks good from my side. We might need to cleanup some of the TODO comments and track those items externally.

stevensJourney (approved):
Looks good to me

modules/module-mysql/dev/config/sync_rules.yaml
```diff
@@ -0,0 +1,12 @@
+# See Documentation for more information:
+# https://docs.powersync.com/usage/sync-rules
+# Note that changes to this file are not watched.
+# The service needs to be restarted for changes to take effect.
+
+# Note that specifying the schema is currently required due to the default
+# schema being specified as `public`, but in mysql the schema is the database name
```
rkistner, line 7: Is this still the case, with `defaultSchema` now being the database name?

modules/module-mysql/test/src/util.ts
```diff
@@ -0,0 +1,62 @@
...
+  const version = await getMySQLVersion(connection);
+  if (gte(version, '8.4.0')) {
+    await connection.query('RESET BINARY LOGS AND GTIDS');
+  } else {
+    await connection.query('RESET MASTER');
+  }
+
+  // await connection.query(`DROP DATABASE IF EXISTS ${TEST_CONNECTION_OPTIONS.database}`);
+  //
+  // await connection.query(`CREATE DATABASE IF NOT EXISTS ${TEST_CONNECTION_OPTIONS.database}`);
```
rkistner, line 50: Can these comments be removed?

rkistner (commented):
Happy on the high-level structure, will see if I can still go through the details of the replication process at some point

modules/module-mysql/dev/config/sync_rules.yaml
Rentacookie: No you are right, no longer needed :)

modules/module-mysql/package.json
```diff
@@ -0,0 +1,51 @@
...
+      "require": "./dist/types/types.js",
+      "default": "./dist/types/types.js"
+    }
+  },
+  "dependencies": {
+    "@powersync/lib-services-framework": "workspace:*",
+    "@powersync/service-core": "workspace:*",
+    "@powersync/service-sync-rules": "workspace:*",
+    "@powersync/service-types": "workspace:*",
+    "@powersync/mysql-zongji": "0.0.0-dev-20241023144335",
```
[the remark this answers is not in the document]
Rentacookie: Definitely. I'm still struggling with an issue where the testing framework of that zongji lib causes the build to fail even though all the tests pass. That is currently the main blocker for getting out a proper released version.

PR #115, opened 2024-11-01
MySQL Type Consolidation

Rentacookie:
This changeset addresses some issues which were found during further testing of the MySQL alpha release:

**Type Discrepancies:**

Some differences in the values read from the MySQL replication stream and values read directly from the database during snapshots were detected. These mismatches were addressed by consolidating the type handling into in the powersync service only. Previously some type mappings were already happening upstream in the Zongji Binlog Listener.

**Duplicate ServerIds:**

MySQL expects replication clients to have unique ServerIds. Previously we were using the default value of 1. This proved to be a problem when deploying new sync rules since for a brief time both the old and the new replication were active causing one of the replication instances to throw an error. This was resolved by generating a random ServerId on startup based off of the current syncRuleId. Since we use GTID based replication and keep track of our position in the binlog ourself, it doesn't matter that the serverId changes regularly. 

**Enabled unit tests in CI:** 

Added the configuration so that the MySQL unit tests run as part of the build.

### DIFF
--- a/modules/module-mysql/test/src/MysqlTypeMappings.test.ts
+++ b/modules/module-mysql/test/src/MysqlTypeMappings.test.ts
@@ -1,0 +1,319 @@
+import { SqliteRow } from '@powersync/service-sync-rules';
+import { afterAll, describe, expect, test } from 'vitest';
+import { clearTestDb, TEST_CONNECTION_OPTIONS } from './util.js';
+import { eventIsWriteMutation, eventIsXid } from '@module/replication/zongji/zongji-utils.js';
+import * as common from '@module/common/common-index.js';
+import ZongJi, { BinLogEvent } from '@powersync/mysql-zongji';
+import { MySQLConnectionManager } from '@module/replication/MySQLConnectionManager.js';
+import { toColumnDescriptors } from '@module/common/common-index.js';
+
+describe('MySQL Data Types', () => {
+  const connectionManager = new MySQLConnectionManager(TEST_CONNECTION_OPTIONS, {});
+
+  afterAll(async () => {
+    await connectionManager.end();
+  });
+
+  async function setupTable() {
+    const connection = await connectionManager.getConnection();
+    await clearTestDb(connection);
+    await connection.query(`CREATE TABLE test_data (
+    tinyint_col TINYINT,
+    smallint_col SMALLINT,
+    mediumint_col MEDIUMINT,
+    int_col INT,
+    integer_col INTEGER,
+    bigint_col BIGINT,
+    float_col FLOAT,
+    double_col DOUBLE,
+    decimal_col DECIMAL(10,2),
+    numeric_col NUMERIC(10,2),
+    bit_col BIT(8),
+    boolean_col BOOLEAN,
+    serial_col SERIAL,
+    
+    date_col DATE,
+    datetime_col DATETIME(3),
+    timestamp_col TIMESTAMP(3),
+    time_col TIME,
+    year_col YEAR,
+
+    char_col CHAR(10),
+    varchar_col VARCHAR(255),
+    binary_col BINARY(16),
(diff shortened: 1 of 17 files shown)


modules/module-mysql/test/src/MysqlTypeMappings.test.ts
```diff
@@ -0,0 +1,302 @@
...
+    await setupTable();
+    await connectionManager.query(`
+INSERT INTO test_data (
+    tinyint_col,
+    smallint_col,
+    mediumint_col,
+    int_col,
+    integer_col,
+    bigint_col,
+ -- float_col,
```
rkistner, line 81: Why is float excluded?

modules/module-mysql/test/src/MysqlTypeMappings.test.ts
```diff
@@ -0,0 +1,302 @@
...
+    'TinyBlobData',           -- TINYBLOB
+    'BlobData',               -- BLOB
+    'MediumBlobData',         -- MEDIUMBLOB
+    'LongBlobData',           -- LONGBLOB
+    'TinyTextData',           -- TINYTEXT
+    'TextData',               -- TEXT
+    'MediumTextData',         -- MEDIUMTEXT
+    'LongTextData',           -- LONGTEXT
+    'value1',                 -- ENUM('value1', 'value2', 'value3')
+    'value1,value3'           -- SET('value1', 'value2', 'value3')
```
rkistner, line 158: Should we perhaps represent this as a JSON array, i.e. `["value1","value3"]`? Not 100% convinced that's the best approach, just wondering if that would be easier to work with. For example, our `IN` operator works with JSON arrays automatically.

modules/module-mysql/test/src/MysqlTypeMappings.test.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    const databaseRows = await getDatabaseRows(connectionManager, 'test_data');
+    const replicatedRows = await getReplicatedRows();
+
+    const expectedResult = {
+      tinyint_col: 127,
+      smallint_col: 32767,
+      mediumint_col: 8388607,
+      int_col: 2147483647,
+      integer_col: 2147483647,
```
rkistner, line 112: For postgres (and mongodb, I think), we use bigint for all integer types. It would be best to use the same here.

That specifically helps to keep the distinction between the "integer" and "real" types in the SQLite type system, which we also use for sync rules.

rkistner (commented):
The changes look great overall, just added some comments on specific type behavior.

modules/module-mysql/test/src/MysqlTypeMappings.test.ts
Rentacookie: I see - so we just convert all integer number types to Bigint?

rkistner: Correct.

modules/module-mysql/test/src/MysqlTypeMappings.test.ts
Rentacookie: With the rounding errors that creep in with the float mapping one cannot test for exact values.
I initially tested it by checking if the value was within a given precision range, but that didn't feel very useful so I ended up removing it.
Its due to these precision errors that MySQL recommends users should rather use the double type.

modules/module-mysql/test/src/MysqlTypeMappings.test.ts
Rentacookie: Oh yeah I think it would be. Let me take look what that mapping entails.

stevensJourney (approved):

rkistner (approved):

Rentacookie:
Thanks for the review guys 🙏

PR #129, opened 2024-11-13
[MongoDB] Skip current_data storage

rkistner:
Builds on #121.

The current_data bucket storage collection stores the following data:

 * `buckets`: The current set of buckets for each replicated row
 * `lookups`: Current parameter query lookups for each replicated row
 * `data`: A copy of the current data

Buckets and lookups are used to be able to do REMOVE operations (need to know in which buckets a row was before, so we know from which buckets we need to remove it from after an update).

The `data` bit is to support partial updates. For Postgres, this could be when we receive an update with unchanged TOAST columns. However, for MongoDB we get the full document on each update, so never need that. This adds an optimization for MongoDB to skip that persistence.

### DIFF
--- a/packages/service-core/src/storage/mongo/MongoBucketBatch.ts
+++ b/packages/service-core/src/storage/mongo/MongoBucketBatch.ts
@@ -33,6 +33,16 @@ const MAX_ROW_SIZE = 15 * 1024 * 1024;
 // In the future, we can investigate allowing multiple replication streams operating independently.
 const replicationMutex = new util.Mutex();
 
+export interface MongoBucketBatchOptions {
+  db: PowerSyncMongo;
+  syncRules: SqlSyncRules;
+  groupId: number;
+  slotName: string;
+  lastCheckpointLsn: string | null;
+  noCheckpointBeforeLsn: string;
+  storeCurrentData: boolean;
+}
+
 export class MongoBucketBatch extends DisposableObserver<BucketBatchStorageListener> implements BucketStorageBatch {
   private readonly client: mongo.MongoClient;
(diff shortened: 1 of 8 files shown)


packages/service-core/src/storage/mongo/MongoBucketBatch.ts
```diff
@@ -142,45 +147,52 @@ export class MongoBucketBatch extends DisposableObserver<BucketBatchStorageListe
...
-    // exceeding memory limits.
-    //
-    // A previous attempt tried to do batching by the results of the current_data query
-    // (automatically limited to 48MB(?) per batch by MongoDB). The issue is that it changes
-    // the order of processing, which then becomes really tricky to manage.
-    // This now takes 2+ queries, but doesn't have any issues with order of operations.
-    const sizeLookups: SourceKey[] = batch.batch.map((r) => {
-      return { g: this.group_id, t: r.record.sourceTable.id, k: r.beforeId };
-    });
+    let sizes: Map<string, number> | undefined = undefined;
```
Rentacookie, line 150: Looks like initializing `sizes` with undefined is not necessary and we can revert to the previous initialization?

rkistner: This function specifically uses `sizes = undefined` when `this.storeCurrentData == false` - it does have an effect in some places.

Rentacookie: Ah I see - hadn't gotten to the OperationBatch change yet 😅

Rentacookie (approved):

PR #135, opened 2024-11-14
Write checkpoint workaround for Aurora Postgres

rkistner:
On most Postgres versions, `pg_logical_emit_message()` returns the correct current LSN. However, on Aurora (Postgres compatible), it can return an entirely different LSN in some (but not all) cases, causing the write checkpoints to never be replicated back to the client. To cater for those cases, we now use `pg_current_wal_lsn()` to get the correct LSN.

Both queries are done in a single pgwire query call, so it is still a single round-trip with minimal overhead.

### DIFF
--- a/.changeset/tender-lions-hope.md
+++ b/.changeset/tender-lions-hope.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-core': patch
+---
+
+Workaround for Aurora Postgres write checkpoint bug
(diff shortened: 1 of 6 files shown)


fooware:
Discord thread for context:
https://discord.com/channels/1138230179878154300/1305574600373178478/1305574685261693061

The tests we have done is strongly indicating that this fix works for us when using Aurora serverless in RDS. However, since the reproduction rate earlier was somewhat random at about 70%, it could be a fluke. Very unlikely, but I want to mention it.

rkistner:
I rebased onto the latest main, and refactored the queries to not add additional overhead (so it doesn't negatively affect the cases that don't need this workaround).

stevensJourney (approved):

PR #136, opened 2024-11-25
MongoDB Bucket Storage Module

stevensJourney:
# Overview

This moves the Mongo DB `BucketStorage` implementation out of the `@powersync/service-core` package to a new MongoDB Storage module.

This PR also aims to move all MongoDB specific functionality to the MongoDB module. This includes:

- Migrations which are powered by MonogDB Migration stores. 
- MongoDB based exclusive locks.

Shared unit tests for the `BucketStorageFactory` are exported from the new `@powersync/service-core-tests` package. These tests are used for testing the MongoDB and Postgres bucket storage factories.


rkistner (commented):
Just some initial comments on locks and migrations.

packages/service-core/src/sync/sync.ts
```diff
@@ -318,9 +318,18 @@ async function* bucketDataBatch(request: BucketDataRequest): AsyncGenerator<Buck
...
+         * FIXME: ensure that the data is actually a string
+         */
         const response: util.StreamingSyncData = {
-          data: r
+          data: {
+            ...r,
+            data: r.data.map((entry) => ({
+              ...entry,
+              data: typeof entry.data == 'string' ? entry.data : JSONBig.stringify(entry.data)
+            }))
```
rkistner, line 331: Why were changes needed here?

rkistner (commented):
Overall the structure here looks good to me.

One thing I'm wondering about in regards to the module structure - would it perhaps work better to create separate modules for replication vs storage?

One specific place this has an effect is with tests: Currently we have for example the mysql module with a devDependency on the mongodb module, so that the mongodb storage can be used with mysql tests. But we probably want to test all combinations of postgres/mongodb/mysql replication with mongodb/postgres storage. The dependency structure between the modules may start getting weird if we implement that.

packages/service-core/package.json
```diff
@@ -29,7 +29,7 @@
     "@powersync/service-types": "workspace:*",
     "async": "^3.2.4",
     "async-mutex": "^0.5.0",
-    "bson": "^6.6.0",
+    "bson": "6.6.0",
```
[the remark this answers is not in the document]
stevensJourney: For context: The `service-core` package contained some `bson` helper functions which I wanted to use upstream in the Postgres Bucket storage, but I noticed when locally linking NPM packages that TypeScript would error if multiple versions of BSON were present in the project. Locking the version temporarily fixed that issue, but this should no longer be needed due to cleaning up and sharing more common bson functionality in the core package. I'll revert this.

libs/lib-services/src/locks/LockManager.ts
```diff
@@ -0,0 +1,22 @@
...
+   * Optionally retry and wait for the lock to be acquired
+   */
+  max_wait_ms?: number;
+};
+
+export type LockManager = {
+  init?: () => Promise<void>;
+  acquire: (options?: LockAcquireOptions) => Promise<string | null>;
+  refresh: (lock_id: string) => Promise<void>;
+  release: (lock_id: string) => Promise<void>;
```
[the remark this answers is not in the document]
stevensJourney: That is nicer. I'll implement that.

packages/service-core/src/sync/sync.ts
stevensJourney: I experienced an issue when testing the Postgres bucket storage over an HTTP stream. 

The `OplogEntry` here has a `data` member which is defined of the type ` data?: Record<string, SqliteJsonValue> | string | null;` and `SqliteJsonValue` supports `number | string | bigint | null`. In MongoDB the data is stored internally and returned (at this point) as a serialised JSON string which works fine. For Postgres I've stored the data as a `bytea` column containing serialised BSON assuming this might make blob/binary data storage simpler in future. The Postgres bucket storage deserialises the BSON to a ` Record<string, SqliteJsonValue>` compatible document which satisfies the `OplogEntry::data` typing, but causes issues in the following `JSON.stringify` call since `bigint` cannot be serialized. 

I've added the check here to continue using `JSON.stringify` for the final response. Alternatively we could use `JSONBig.stringify`, but I noticed the comment above mentions it is less efficient.

rkistner: The `OplogEntry` type does make things a little confusing, since it's used for both the bucket storage APIs, and the protocol, but differs a bit between them (currently data is always `string | null` in mongodb storage).

Storing as BSON in Postgres is an interesting idea. I think that would make sense if we start syncing rows directly as BSON in the future. However, since we always sync as a string right now, doing that `JSONBig` serialization for every client adds a _lot_ of overhead. So I'd recommend just storing it as a string for now, and re-evaluate this when we do add proper binary syncing support later.

modules/module-mongodb-storage/src/module/MongoStorageModule.ts
```diff
@@ -0,0 +1,37 @@
...
+   */
+  private resolveConfig(config: types.MongoStorageConfig) {
+    return {
+      ...config,
+      ...lib_mongo.normalizeMongoConfig(config)
+    };
+  }
+
+  async teardown(options: core.modules.TearDownOptions): Promise<void> {
+    // teardown is implemented in the storage engine
```
rkistner, line 35: I assume this means the MongoStorageProvider?

What is the interaction between this module and the StorageProvider - why the two teardown methods?

rkistner (approved):

Issue #72, opened 2024-09-01
support jwts signed using EdDSA algorithm


wildsylvan:
> EdDSA is known as a high performance signature algorithm with small key sizes and signatures. Using EdDSA signatures over RSA saves time, money and resources. EdDSA is also more secure than RSA.
> EdDSA is deterministic and does not depend on a random number generator for security. The requirement of a nonce and random number generator to create one has lead to known vulnerabilities in ECDSA implementations in the past. In addition, EdDSA does not require expensive point validation for corner cases. Consequently, EdDSA is more secure than ECDSA because it is simply easier to implement.
> As long as there are no other requirements that dictate the use of certain signature algorithms, switch to EdDSA if you can. It's fast, it's secure, and it's green.

https://github.com/powersync-ja/powersync-service/blob/a9ec391f2a66c987b084317a7cfa626334d882f3/packages/service-core/src/auth/KeySpec.ts#L5

LucDeCaf:
Hi @wildsylvan, [this PR](https://github.com/powersync-ja/powersync-service/pull/142) just got merged, which should allow you to use EdDSA in hopefully the next version of the service 👍

rkistner:
This is now available in the latest self-hosted and cloud versions. Thanks @LucDeCaf!

PR #142, opened 2024-11-29
Add EdDSA support for signing JWTs

LucDeCaf:
This adds support for the EdDSA signing algorithm in `core_service` as requested in [this issue](https://github.com/powersync-ja/powersync-service/issues/72).

I am not super familiar with the `core_service` architecture so feedback on what needs to be added/improved would be greatly appreciated.


rkistner (commented):
To also support EdDSA with a JWKS URI, could you also modify this bit to accept these keys?

https://github.com/powersync-ja/powersync-service/blob/a47bda4bf84c508d38be4a248e98991a1af7a334/packages/service-core/src/auth/RemoteJWKSCollector.ts#L76

That check is different from the ones in KeySpec, since we specifically do not want to allow the HS algorithms (symmetric keys) with JWKS URIs.

packages/service-core/test/src/auth.test.ts
```diff
@@ -331,12 +342,34 @@ describe('JWT Auth', () => {
...
+      .setAudience('tests')
+      .setExpirationTime('5m')
+      .sign(signKey);
+
+    const verified = (await store.verifyJwt(signedJwt, {
+      defaultAudiences: ['tests'],
+      maxAge: '6m'
+    })) as JwtPayload & { claim: string };
+
+    expect(verified.claim).toEqual('test-claim');
```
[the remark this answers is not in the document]
LucDeCaf: Whoops 😅

packages/service-core/test/src/auth.test.ts
```diff
@@ -331,12 +342,34 @@ describe('JWT Auth', () => {
...
-    expect(key.kid).toEqual(publicKey.kid!);
+    expect(key.kid).toEqual(publicKeyRSA.kid!);
+  });
+
+  test('signing with EdDSA', async () => {
+    const keys = await StaticKeyCollector.importKeys([publicKeyEdDSA]);
+    const store = new KeyStore(keys);
+    const signKey = (await jose.importJWK(publicKeyEdDSA)) as jose.KeyLike;
+
+    const signedJwt = await new jose.SignJWT({ claim: 'test-claim' })
```
rkistner, line 359: The test is failing here - it currently attempts to sign using the _public_ key, while it should sign using the _private_ key.

rkistner: Thanks! Looks good now. I'll test properly next week, then we can merge.

rkistner (approved):
Thanks, looks great!

PR #147, opened 2024-12-02
Sync rules function to convert uuid to base64

js2702:
In our project we use base64 encoded uuids on the client, for faster indexes and joins. 

This PR adds a function to convert the underline uuid bytes into base64, alternatively there could be a function to convert to just bytes and then use the base64 sync rule function. Let us know what you think is best.

This is our Discord discussion: https://discord.com/channels/1138230179878154300/1312036536849666118

### DIFF
--- a/.changeset/swift-ways-live.md
+++ b/.changeset/swift-ways-live.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': minor
+---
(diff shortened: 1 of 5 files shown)


rkistner (approved):
Thanks, this is useful! Will likely have a release out tomorrow

PR #152, opened 2024-12-06
[Postgres] Improve timeouts and table snapshots

rkistner:
Builds on #150.

This configures both TCP keepalive and connection timeouts, to detect broken snapshot connections in around 30 seconds instead of 6 minutes.

This now also uses `DECLARE CURSOR` and `FETCH` to fetch data in batches, instead of one large stream. This may reduce the chance of these connections hanging.

### DIFF
--- a/modules/module-postgres/src/replication/WalStream.ts
+++ b/modules/module-postgres/src/replication/WalStream.ts
@@ -352,10 +352,19 @@ WHERE  oid = $1::regclass`,
               logger.info(`${this.slot_name} Skipping ${table.qualifiedName} - snapshot already done`);
               continue;
             }
-            await this.snapshotTable(batch, db, table);
+            let tableLsnNotBefore: string;
+            await db.query('BEGIN');
+            try {
+              await this.snapshotTable(batch, db, table);
 
-            const rs = await db.query(`select pg_current_wal_lsn() as lsn`);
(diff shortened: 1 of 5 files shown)


modules/module-postgres/src/replication/WalStream.ts
```diff
@@ -342,25 +342,35 @@ WHERE  oid = $1::regclass`,
...
             await touch();
           }
         }
-        await batch.commit(startLsn);
+
+        // Always commit the initial snapshot at zero.
+        // This makes sure we don't skip any changes applied before starting this snapshot,
+        // in the case of snapshot retries.
+        // We could alternatively commit at the replication slot LSN.
+        await batch.commit(ZERO_LSN);
```
rkistner, line 373: This was not an issue before. However, changing the snapshot to execute in a transaction, introduced consistency issues - streaming replication after the snapshot could skip changes that shouldn't be skipped.

I'm not sure why changing between using/not using a transaction has an effect on this, but the real issue was doing a commit with the wrong LSN.

There is a small chance that there are more scenarios triggering the same consistency issue since #150 (only when initial replication is interrupted and retried).

stevensJourney (approved):
Looks good to me.

stevensJourney (approved):

PR #163, opened 2024-12-11
[Postgres + MongoDB] Resumable replication

rkistner:
## Problem Statement

When updating sync rules, we perform initial replication from the source database with the new sync rules. This involves doing an initial snapshot, then streaming incremental changes.

For large source databases, initial replication can take a long time. Historically we did not implement resuming for the initial snapshot queries - only for the initial replication stream. So if the replication process/container crashed/restarted in that process, we had to restart the snapshot from scratch.

We have already implemented partial mitigations for Postgres source databases:
 * #150 added support to skipping tables already replicated.
 * #152 improved table snapshots to be faster when retrying - avoiding writing the same rows again.

## The fix

This implements the same mitigations for MongoDB source databases, and additionally attempts to chunk initial snapshot queries using the primary key, allowing us to fully resume initial replication where we left off. This is basically repeated queries of the form `SELECT * from <table> WHERE id > :lastId LIMIT 10000`, instead of just using a single `SELECT * FROM <table>` query.

For Postgres, this is limited to tables with a single primary key column of specific types (text/varchar/uuid/int2/int4/int8). We can add more supported types over time, and potentially support compound primary keys, if we do proper testing. However, these should already cover a large percentage of replicated tables. Tables not covered from this would restart replication when interrupted.

For MongoDB, we support any collection.

### Edge cases

Since the snapshot query is no longer performed at a single point-in-time, there are some edge cases we need to consider, when rows are added, updated or removed while snapshotting. Most of these are handled with the proper consistency purely by resuming streaming replication after the snapshot. However, Postgres has a particular test case we need to be careful with:
1. Replicate snapshot for chunk A.
2. Update a row, moving it from chunk B to chunk B.
3. Replicate snapshot for chunk B.

This means the row is not covered by the snapshot query. It is still covered by streaming replication, but the change event may exclude TOAST values.

To handle this edge case, we detect rows with missing TOAST columns, and re-replicate those.

### Limitations

Some notable limitations currently:
1. Resumable replication is not implemented for MySQL yet.
2. ~Progress inside a table is not implemented for Postgres storage yet.~
3. ~Resuming MongoDB snapshots require storing the snapshot LSN, which isn't implemented for Postgres storage yet.~

### Logging

This improves replication logging a bit, by:
1. Pre-calculating table sizes, and logging this when we start replicating.
2. Using a consistent log prefix as part of the logger instance when replicating.

In the future we can use this to expose snapshot progress via the diagnostics API, but for now logs can be used to see that.

### MongoDB snapshot resume token

Previously, we stored the `clusterTime` when starting a MongoDB snapshot, then resumed streaming from that point afterwards. Now, this is replaced by an actual resume token. This allows for better detection of replication issues after the initial snapshot, such as the oplog window being too small, or switching source databases.

### Migrations

For Postgres storage, this has a new migration for the additional snapshot metadata. No migration is needed for MongoDB storage.


packages/service-core/src/storage/bson.ts
```diff
@@ -10,7 +10,10 @@ type NodeBuffer = Buffer<ArrayBuffer>;
  */
 export const BSON_DESERIALIZE_INTERNAL_OPTIONS: bson.DeserializeOptions = {
   // use bigint instead of Long
-  useBigInt64: true
+  useBigInt64: true,
+  // We cannot use promoteBuffers: true, since that also converst UUID to Buffer
+  // Instead, we need to handle bson.Binary when reading data
+  promoteBuffers: false
```
rkistner, line 16: This does not actually change anything - just a note to not attempt to change that in the future.

modules/module-postgres/src/replication/SnapshotQuery.ts
```diff
@@ -0,0 +1,206 @@
...
+      }
+      yield chunk;
+    }
+  }
+}
+
+/**
+ * This performs a snapshot query using a list of primary keys.
+ */
+export class IdSnapshotQuery {
```
Rentacookie: Looks like this class is missing the `implements SnapshotQuery`

modules/module-postgres/src/replication/WalStream.ts
```diff
@@ -442,60 +494,147 @@ WHERE  oid = $1::regclass`,
...
+  ): Promise<storage.SourceTable> {
+    await db.query('BEGIN');
+    try {
+      let tableLsnNotBefore: string;
+      await this.snapshotTable(batch, db, table, limited);
+
+      // Get the current LSN.
+      // The data will only be consistent once incremental replication
+      // has passed that point.
+      // We have to get this LSN _after_ we have started the snapshot query.
```
Rentacookie: Getting the LSN after the snapshot query is throwing me a bit. Does this work because at this point because we haven't replicated anything from the wal yet? Would `select pg_current_wal_lsn() as lsn` have returned a different LSN if done before the snapshot query?

modules/module-postgres/src/replication/WalStream.ts
```diff
@@ -442,60 +494,147 @@ WHERE  oid = $1::regclass`,
...
-          afterReplicaId: getUuidReplicaIdentityBson(record, table.replicaIdColumns)
+      // Important: flush before marking progress
+      await batch.flush();
+      if (limited == null) {
+        let lastKey: Uint8Array | undefined;
+        if (q instanceof ChunkedSnapshotQuery) {
+          lastKey = q.getLastKeySerialized();
+        }
+        if (lastCountTime < performance.now() - 10 * 60 * 1000) {
+          totalEstimatedCount = await this.estimatedCountNumber(db, table);
```
Rentacookie, line 635: Do we expect this count to change while the snapshot is in progress for the table?
Just wondering since this is all happening in the same transaction 🤔

rkistner: We use the default "Read Committed" isolation level for the transaction, which means the data can change between queries within that transaction. [isolation level docs](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-READ-COMMITTED)

I should probably add some comments regarding that.

That said, updating the count is not super important - it's not often that you'll get major changes while replicating. Not sure if it's actually worth the extra effort of re-counting while replicating.

modules/module-postgres/src/replication/WalStream.ts
rkistner: AFAIK, `pg_current_wal_lsn()` is independent of the transaction you're running it in.

There are basically two relevant LSNs here:
1. A: The LSN _before_ the snapshot starts. We don't explicitly record this on the PowerSync side, but it is implicitly recorded in the replication slot.
2. B: The LSN _after_ the table snapshot is complete, which is what is referred to here.

When we do the snapshot queries, the data that we get back could match the state anywhere between A and B. To actually have a consistent state on our side, we need to:
1. Complete the snapshot.
2. Wait until logical replication has caught up with all the change between A and B.

The comment is actually not fully correct - it says after we have _started_ the snapshot query, but it should be after we have _finished_ the snapshot.

modules/module-postgres/src/replication/WalStream.ts
Rentacookie: Ah of course! Was mostly curious. 😄

Rentacookie (dismissed):
This is some seriously impressive stuff - nice going Ralf 🎉

modules/module-postgres/src/replication/WalStream.ts
Rentacookie: Ah i see now - I initially thought that this specific LSN was the `before` snapshot LSN 😅

PR #166, opened 2024-12-18
iif function based in the iif function of sqlite

js2702:
When implementing the base64 uuid functionality we noticed that UUID foreign keys that were null were being transformed as empty string in the client.

This is because of the implementation of base64 function, which returns an empty string when passing it null. 

I guess changing this would be a breaking change, so with this function we should be able to handle null values without breaking existing functionality.

We followed the behavior of the SQLite IIF function described [here](https://www.sqlite.org/lang_corefunc.html#iif), which is essentially an if-else.

Alternatively it might make sense to change the base64 and hex functions so that they behave like the other functions, where they return null when inputted null. For example the length function.

### DIFF
--- a/packages/sync-rules/src/sql_functions.ts
+++ b/packages/sync-rules/src/sql_functions.ts
@@ -255,18 +255,7 @@ const ifnull: DocumentedSqlFunction = {
 const iif: DocumentedSqlFunction = {
   debugName: 'iif',
   call(x: SqliteValue, y: SqliteValue, z: SqliteValue) {
-    let isTrue: boolean;
-    if (Number.isNaN(x)) {
-      isTrue = false;
-    } else if (typeof x == 'number') {
-      isTrue = x !== 0;
-    } else if (typeof x == 'bigint') {
-      isTrue = x !== 0n;
-    } else {
-      isTrue = false;
-    }
-
-    return isTrue ? y : z;
+    return sqliteBool(x) ? y : z;
   },
   parameters: [
     { name: 'x', type: ExpressionType.ANY, optional: false },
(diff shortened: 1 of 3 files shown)


packages/sync-rules/src/sql_functions.ts
```diff
@@ -251,6 +251,34 @@ const ifnull: DocumentedSqlFunction = {
...
+    let isTrue: boolean;
+    if (Number.isNaN(x)) {
+      isTrue = false;
+    } else if (typeof x == 'number') {
+      isTrue = x !== 0;
+    } else if (typeof x == 'bigint') {
+      isTrue = x !== 0n;
+    } else {
+      isTrue = false;
+    }
```
rkistner, line 267: Could you use the `sqliteBool` function here? That would make sure it's consistent with the boolean handling we have elsewhere.

js2702: Just did that, didn't see that when implementing the function.

rkistner:
Thanks, this is definitely a useful function!

> Alternatively it might make sense to change the base64 and hex functions so that they behave like the other functions, where they return null when inputted null. For example the length function.

We're currently following the SQLite behavior for the hex function, which returns an empty text value for null input. I agree that it's not ideal for scenarios like this, but we're striving to keep compatibility with existing SQLite functions as far as possible.

For the base64 function our behavior is actually different from the official [base64 extension](https://www.sqlite.org/base64.html) - that one throws an error on null input (despite what the docs say). I don't think it's worth making the breaking changes to get compatibility in this case, and it wouldn't help for the use case here anyway.

rkistner (approved):

davidmartos96:
@rkistner Did the base64 spec changed in some update? The docs in the link you posted mention that it will return NULL when the input is NULL.

> If the argument to base64() is NULL, then NULL is returned.

https://www.sqlite.org/base64.html

PR #169, opened 2024-12-31
Handle MySQL servers with non-UTC timezone

rkistner:
Tests currently fail if the MySQL server is configured with a timezone other than UTC.

The issue is that MySQL always returns timestamps in the session timezone, without any timezone indicator. So there is no way to know what we should configure the client `timezone` with if we don't know the server timezone.

The current tests also stored timestamps in the default/server timezone, but expected them to be returned in UTC timezone.

This changes:
1. Fix the tests to use `+00:00` timezone in literals.
2. Always set the timezone on the mysql connection used for queries: `SET time_zone = "+00:00"`. This is somehow not needed for zongji. This is currently set on every query - not sure if there is a more efficient way to do this with a pool.


rkistner:
Superseded by #170.

PR #170, opened 2024-12-31
MySQL replication fixes

rkistner:
Fixes some issues with MySQL replication:

### Timestamps: Timezone handling

Timezone in timestamps were not handled correctly if the MySQL server is configured with a timezone other than UTC. The issue is that MySQL always returns timestamps in the session timezone, without any timezone indicator. So there is no way to know what we should configure the client `timezone` with if we don't know the server timezone.

The current tests also stored timestamps in the default/server timezone, but expected them to be returned in UTC timezone.

This changes:
1. Fix the tests to use `+00:00` timezone in literals.
2. Always set the timezone on the mysql connection used for queries: `SET time_zone = "+00:00"`. This does not appear to be needed for zongji.

### Timestamps: Updates broken

When updating a row containing an update, the row was converted in-place to the SQLite format, then converted again. The first pass converted the timestamp to a string, then the second pass failed since it expected a Date. This results in an error `row[key].toISOString is not a function`.

This fixes the conversion to not modify the original row, and also avoids doing the conversion twice.

### Timestamps: Handling invalid dates

With MySQL, it is possible to disable strict mode and then persist a timestamp such as `2024-00-00 00:00:00`. This is not a valid Date in JavaScript, and results in an error `RangeError: Invalid time value` when calling `toISOString()`.

This fixes the issue by converting these to null.

### Initial replication: Improved error handling

The implementation of snapshot handling did not handle errors properly in all cases (such as the above issue) - it resulted in an unhandled promise rejection, which could cause either the process to crash, or replication to stall.

This changes the snapshot query to using a Readable stream, which does not have this issue.

### Resuming replication

When replication is resumed after a restart, the list of tables were not loaded, resulting in an empty list of tables in `includeSchema`, and no changes replicated. This now loads the list of tables when resuming replication.

### Quoting table names

This now properly quotes table names with backticks in snapshot queries, avoiding potential issues when a table name contains special characters for some reason. Unlike with Postgres, this is not required matching the table name case, but is still better to do either way.

### DIFF
--- a/modules/module-mysql/src/common/mysql-to-sqlite.ts
+++ b/modules/module-mysql/src/common/mysql-to-sqlite.ts
@@ -99,6 +99,7 @@ export function toColumnDescriptorFromDefinition(column: ColumnDefinition): Colu
 }
 
 export function toSQLiteRow(row: Record<string, any>, columns: Map<string, ColumnDescriptor>): sync_rules.SqliteRow {
+  let result: sync_rules.DatabaseInputRow = {};
   for (let key in row) {
     // We are very much expecting the column to be there
     const column = columns.get(key)!;
@@ -107,17 +108,17 @@ export function toSQLiteRow(row: Record<string, any>, columns: Map<string, Colum
       switch (column.typeId) {
         case mysql.Types.DATE:
           // Only parse the date part
-          row[key] = row[key].toISOString().split('T')[0];
+          result[key] = row[key].toISOString().split('T')[0];
           break;
         case mysql.Types.DATETIME:
         case ADDITIONAL_MYSQL_TYPES.DATETIME2:
         case mysql.Types.TIMESTAMP:
         case ADDITIONAL_MYSQL_TYPES.TIMESTAMP2:
-          row[key] = row[key].toISOString();
+          result[key] = row[key].toISOString();
           break;
         case mysql.Types.JSON:
           if (typeof row[key] === 'string') {
-            row[key] = new JsonContainer(row[key]);
+            result[key] = new JsonContainer(row[key]);
           }
           break;
         case mysql.Types.BIT:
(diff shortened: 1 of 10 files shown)


modules/module-mysql/src/common/mysql-to-sqlite.ts
```diff
@@ -107,17 +108,35 @@ export function toSQLiteRow(row: Record<string, any>, columns: Map<string, Colum
       switch (column.typeId) {
         case mysql.Types.DATE:
           // Only parse the date part
-          row[key] = row[key].toISOString().split('T')[0];
+          {
+            const date = row[key] as Date;
+            if (isNaN(date.getTime()) && false) {
```
stevensJourney, line 113: Is the `&& false` intentional/necessary? I might be missing some JS quirk here though.

rkistner: Nope, this was to disable the logic for testing - thanks for catching.

modules/module-mysql/test/src/mysql-to-sqlite.test.ts
```diff
@@ -222,24 +223,62 @@ INSERT INTO test_data (
...
+      for (let i = 0; i < expectedResults.length; i++) {
+        expect(databaseRows[i]).toMatchObject(expectedResults[i]);
+        expect(replicatedRows[i]).toMatchObject(expectedResults[i]);
+      }
+    } finally {
+      connection.release;
+    }
+  });
+
+  test.skip('Date types mappings - null', async () => {
```
stevensJourney, line 263: should this still be skipped?

modules/module-mysql/src/common/mysql-to-sqlite.ts
rkistner: Also, there isn't a test for that case yet (a date value of e.g. `2025-00-00`), since the behavior is different between initial replication versus streaming replication (one triggers this condition, the other returns a different date).

modules/module-mysql/test/src/mysql-to-sqlite.test.ts
rkistner: Removed the test now - it wasn't actually testing anything useful. I initially thought the issue was with updating a timestamp to null, but it was actually with making any update when there is a timestamp column.

stevensJourney (approved):

modules/module-mysql/src/replication/BinLogStream.ts
```diff
@@ -291,38 +295,32 @@ AND table_type = 'BASE TABLE';`,
...
-          connection.resume();
-          Metrics.getInstance().rows_replicated_total.add(1);
-        })
-        .on('end', async function () {
-          await batch.flush();
-          resolve();
-        });
+    // MAX_EXECUTION_TIME(0) hint disables execution timeout for this query
+    const query = connection.query(`SELECT /*+ MAX_EXECUTION_TIME(0) */ * FROM ${escapeMysqlTableName(table)}`);
+    const stream = query.stream();
```
Rentacookie, line 300: This is a nice simplification to the streaming process! 🥳

PR #171, opened 2025-01-06
MongoDB: Fix replication of undefined values

rkistner:
When replicating Postgres rows, in some edge cases we have toasted (missing) values in rows. We use `undefined` in our internal "current_data" storage to indicate these values are missing, and wait until we have all values before processing the row in sync rules.

With MongoDB replication, we always get the full document - we don't run into the case above. However, we had a bug where `undefined` values were kept as-is when replicating, triggering that logic. This caused rows with undefined values to not be replicated, without any indication why.

Our tests appeared to check for this case, but didn't in reality. MongoDB deprecated the undefined value in BSON, making it very difficult to actually insert an undefined value in the database. Our tests actually only tested for `null`, and never for `undefined`. The tests now use a nasty workaround using `mapReduce` to insert an `undefined` value.

This only changes the behavior for top-level `undefined` values. Behavior for `undefined` in nested objects and arrays is unchanged:

```js
// top-level: converted to null
{value: undefined} -> {value: null}
// array: converted to null
{array: [undefined] -> {array: "[null]"}
// nested document: treated as missing key
{nested: {value: undefined} -> {nested: "{}"}
```

The the [Discord thread](https://discord.com/channels/1138230179878154300/1324331616973361202/1324430130558533643) reporting this issue.

### DIFF
--- a/.changeset/sweet-chairs-build.md
+++ b/.changeset/sweet-chairs-build.md
@@ -1,0 +1,7 @@
+---
+'@powersync/service-module-mongodb': patch
+'@powersync/service-core': patch
+'@powersync/service-image': patch
+---
+
+MongoDB: Fix replication of undefined values causing missing documents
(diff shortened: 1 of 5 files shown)


stevensJourney (approved):
LGTM

PR #173, opened 2025-01-07
Added heartbeat to Mysql Binlog Listener connection

Rentacookie:
It was found that under some circumstance mysql replication stops without any errors, and only starts up correctly again when the powersync instance is restarted. We suspect that this may be happening to due an error on the replication connection that is not picked up by the Binlog listener. 

This adds a heartbeat mechanism on the listener's replication connection to ensure we timeously pick up when the connection has stalled.

### DIFF
--- a/.changeset/grumpy-cameras-breathe.md
+++ b/.changeset/grumpy-cameras-breathe.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-module-mysql': patch
+---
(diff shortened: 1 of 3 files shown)


rkistner (dismissed):

rkistner:
Note: still needs a changeset.

rkistner (approved):

PR #178, opened 2025-01-10
Postgres Bucket Storage

stevensJourney:
# Overview

In some circumstances configuring a Postgres server for bucket storage might be simpler than a MongoDB replica set. 

This PR introduces a new `@powersync/service-module-postgres-storage` module which registers the ability to use a Postgres database for sync bucket storage.

See the [readme](https://github.com/powersync-ja/powersync-service/blob/pg-bucket-storage/modules/module-postgres-storage/README.md#L1) for some additional context.

# Testing

The Postgres storage implementation currently passes all unit tests. Tests include the shared storage unit tests from `@powersync/service-core-tests` and the replicator tests for all the current replication modules where Postgres is used in addition to MongoDB as a storage provider. 

This has also been tested in an application demo environment in the new [Postgres Bucket storage self host demo](https://github.com/powersync-ja/self-host-test/blob/pg-storage/demos/nodejs-postgres-bucket-storage/README.md#L1) from https://github.com/powersync-ja/self-host-demo/pull/45. 

Basic performance comparisons between MongoDB were made for testing purposes only. Running locally on an M1 Mac, an initial replication of 2 million rows was timed. Initial replication took 4 minutes and 20 seconds for Postgres and 3 minutes 59 seconds for MongoDB.

Memory allocations were monitored using a debugger while doing initial replication and streaming data to clients. No noticeable memory leaks were found. 

## Future work

Some of the storage implementation classes duplicate minor implementation between them. These could be improved by using abstract classes at some stage.

There is room for performance optimisations for both reads and writes. E.g. when using the `test-client` to `fetch-operations` for 2.5 million rows of a single global bucket on an M3 Macbook it takes about 25 seconds for MongoDB to respond with all the data, while Postgres takes 75 seconds.


modules/module-postgres-storage/src/migrations/scripts/1684951997326-init.ts
```diff
@@ -0,0 +1,158 @@
...
+     * [For comparison MongoDB took 5min 22 seconds to replicate the same data].
+     *
+     * The time to fetch operations for the 2.5mil rows is 2min, 35 seconds with indexes versus
+     * 2min 21 seconds without indexes (probably in the margin of error).
+     * [For comparison MongoDB took 1 minute for the same data].
+     *
+     * Not including them since they impact initial replication more than providing any noticeable benefit.
+     */
+    // await db.sql`CREATE INDEX idx_bucket_data_composite ON bucket_data (group_id, bucket_name, op_id); `.execute();
+    // await db.sql`CREATE INDEX idx_bucket_data_group_id ON bucket_data (group_id);`.execute();
```
rkistner, line 59: The unique constraint above already creates an index:

```sql
SELECT indexname, indexdef FROM pg_indexes WHERE tablename = 'bucket_data';

 indexname |                                             indexdef                                              
-----------+---------------------------------------------------------------------------------------------------
 unique_id | CREATE UNIQUE INDEX unique_id ON powersync.bucket_data USING btree (group_id, bucket_name, op_id)
```

This index would make both indexes mentioned in the comments redundant.

It would however recommend that you use a primary key constraint rather than just an unique index. If I understand correctly it doesn't have a significant impact on how the data is actually stored or queried, but it is best practice to have a primary key on each table.

The same applies to `custom_write_checkpoints` and `current_data`.

modules/module-postgres-storage/src/storage/batch/PostgresBucketBatch.ts
```diff
@@ -0,0 +1,888 @@
...
+
+/**
+ * Intermediate type which helps for only watching the active sync rules
+ * via the Postgres NOTIFY protocol.
+ */
+const StatefulCheckpoint = models.ActiveCheckpoint.and(t.object({ state: t.Enum(storage.SyncRuleState) }));
+type StatefulCheckpointDecoded = t.Decoded<typeof StatefulCheckpoint>;
+
+// The limits here are not as strict as MongoDB
+const MAX_ROW_SIZE = 40_000_000;
```
rkistner, line 39: I'd recommend keeping the same limit as we have for MongoDB for the initial release, then re-evaluate what we'd like here - this can potentially have big impacts on memory usage in many parts of the service. And it's easy to increase this later, but we can't decrease it without breaking backwards-compatibility.

modules/module-postgres-storage/src/migrations/scripts/1684951997326-init.ts
stevensJourney: Thanks. That is a good point, I've updated the indexes and primary keys for tables.

modules/module-postgres-storage/src/storage/batch/PostgresBucketBatch.ts
stevensJourney: 👍  Updated the value now.

modules/module-postgres-storage/src/migrations/scripts/1684951997326-init.ts
```diff
@@ -0,0 +1,139 @@
...
+        id BIGSERIAL PRIMARY KEY,
+        state TEXT NOT NULL,
+        snapshot_done BOOLEAN NOT NULL DEFAULT FALSE,
+        last_checkpoint BIGINT,
+        last_checkpoint_lsn TEXT,
+        no_checkpoint_before TEXT,
+        slot_name TEXT,
+        last_checkpoint_ts TIMESTAMP WITH TIME ZONE,
+        last_keepalive_ts TIMESTAMP WITH TIME ZONE,
+        keepalive_op TEXT,
```
rkistner, line 59: I think this should be BIGINT instead of TEXT. No idea why I had it as a string in MongoDB storage, but I think that's wrong (it should match checkpoints/op_ids).

modules/module-postgres-storage/src/migrations/scripts/1684951997326-init.ts
```diff
@@ -0,0 +1,139 @@
...
+        data TEXT,
+        target_op bigint
+      )
+    `.execute();
+
+    await db.sql`CREATE TABLE instance (id TEXT PRIMARY KEY) `.execute();
+
+    await db.sql`
+      CREATE TABLE sync_rules (
+        id BIGSERIAL PRIMARY KEY,
```
rkistner, line 50: The size of this should match group_id used in other places (plain `SERIAL` which is equivalent to `INTEGER` instead of `BIGSERIAL`).

This also creates a second sequence, `sync_rules_id_seq`, in addition to the `sync_rules_id_sequence` explicitly defined. It would be better to either:
1. Change this to a plain INTEGER field, with either no default, or a default of `nextval('sync_rules_id_sequence')`, OR:
2. Remove the explit `sync_rules_id_sequence`, and just rely on the auto-incrementing id behavior of `SERIAL`.

modules/module-postgres-storage/src/storage/batch/PostgresBucketBatch.ts
```diff
@@ -0,0 +1,891 @@
...
+      if (no_checkpoint_before_lsn > this.no_checkpoint_before_lsn) {
+        this.no_checkpoint_before_lsn = no_checkpoint_before_lsn;
+
+        await db.sql`
+          UPDATE sync_rules
+          SET
+            no_checkpoint_before = ${{ type: 'varchar', value: no_checkpoint_before_lsn }},
+            last_keepalive_ts = ${{ type: 1184, value: new Date().toISOString() }}
+          WHERE
+            id = ${{ type: 'int8', value: this.group_id }}
```
rkistner, line 424: There's currently a mix of `int8` and `int4` used for this id in code - I think just stick to `int4`.

libs/lib-postgres/src/locks/PostgresLockManager.ts
```diff
@@ -0,0 +1,126 @@
...
+  protected async _acquireId(): Promise<string | null> {
+    const now = new Date();
+    const expiredTs = new Date(now.getTime() - this.timeout);
+    const lockId = uuidv4();
+
+    try {
+      // Attempt to acquire or refresh the lock
+      const res = await this.db.query(sql`
+        INSERT INTO
+          locks (name, lock_id, ts)
```
rkistner, line 62: [Postgres Advisory Locks](https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS) could be useful for this. It has the advantage that the lock will automatically be released when the connection is closed, e.g. due to the process crashing. That also means that you won't have any issues with refreshing the lock.

The disadvantage is that the lock is tied to a specific Postgres connection, so it may be more tricky to manage with connection pooling.

Either way, this is a bigger change, and we can investigate it after the initial release.

libs/lib-postgres/src/locks/PostgresLockManager.ts
```diff
@@ -0,0 +1,126 @@
...
+        ON CONFLICT (name) DO UPDATE
+        SET
+          lock_id = CASE
+            WHEN locks.ts <= $4 THEN $2
+            ELSE locks.lock_id
+          END,
+          ts = CASE
+            WHEN locks.ts <= $4 THEN $3
+            ELSE locks.ts
+          END
```
rkistner, line 78: The WHEN conditions should always be true due to the WHERE clause. Can we unconditionally set these two columns?

.github/workflows/test.yml
```diff
@@ -104,6 +104,18 @@ jobs:
...
+        run: |
+          docker run \
+            --health-cmd pg_isready \
+            --health-interval 10s \
+            --health-timeout 5s \
+            --health-retries 5 \
+            -e POSTGRES_PASSWORD=postgres \
+            -e POSTGRES_DB=powersync_storage_test \
+            -p 5431:5432 \
+            -d postgres:${{ matrix.postgres-version }}
```
rkistner, line 117: Is it feasible to use a single Postgres server for both? (Not important if that's too much effort)

libs/lib-postgres/src/db/connection/DatabaseClient.ts
```diff
@@ -0,0 +1,250 @@
...
+  query(...args: pgwire.Statement[]): Promise<pgwire.PgResult>;
+  async query(...args: any[]): Promise<pgwire.PgResult> {
+    await this.initialized;
+
+    const { schemaStatement } = this;
+    if (typeof args[0] == 'object' && schemaStatement) {
+      args.unshift(schemaStatement);
+    } else if (typeof args[0] == 'string' && schemaStatement) {
+      args[0] = `${schemaStatement.statement}; ${args[0]}`;
+    }
```
rkistner, line 99: It would be great if we could run the schemaStatement once per connection, instead of on every query. That's probably tricky with the pgwrire connection pool though.

Perhaps just add a comment explaining what is happening here for now?

libs/lib-postgres/src/db/connection/DatabaseClient.ts
```diff
@@ -0,0 +1,250 @@
...
+  connectionCreated?: (connection: pgwire.PgConnection) => Promise<void>;
+};
+
+export const TRANSACTION_CONNECTION_COUNT = 5;
+
+export class DatabaseClient extends AbstractPostgresConnection<DatabaseClientListener> {
+  closed: boolean;
+
+  protected pool: pgwire.PgClient;
+  protected connections: ConnectionSlot[];
```
rkistner, line 30: This appears to use a combination of a pgwire pool, and a custom connection pool. Where should one be used versus the other?

.github/workflows/test.yml
stevensJourney: I tested with a single server and that resulted in an endless "loop" of processing keep_alive operations. I think we might be able to work around that specific issue by potentially throttling keep_alive events, but that seems a bit hacky and might have other consequences that I'm not aware of. I'd like to address after the initial release if possible.

modules/module-postgres-storage/README.md
```diff
@@ -0,0 +1,88 @@
+# Postgres Sync Bucket Storage
+
+This module provides a `BucketStorageProvider` which uses a Postgres database for persistence.
+
+## Configuration
+
+Postgres storage can be enabled by selecting the appropriate storage `type` and providing connection details for a Postgres server.
+
+The storage connection configuration extends the configuration for a Postgres replication source, thus it accepts and supports the same configurations fields.
```
rkistner, line 9: "extends the configuration" feels weird here - can we just say it uses the same configuration fields?

modules/module-postgres-storage/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+# Connection settings for sync bucket storage
+storage:
+  type: postgresql
+  # This accepts the same parameters as a Postgres replication source connection
+  uri: !env PS_STORAGE_SOURCE_URI
+```
+
+**IMPORTANT**:
+A separate Postgres server is currently required for replication connections (if using Postgres for replication) and storage. Using the same server might cause unexpected results.
```
rkistner, line 30: A separate server or separate database?

Is this just because we haven't done enough testing, or are there actual issues when using the same server?

libs/lib-postgres/src/db/connection/DatabaseClient.ts
stevensJourney: Yeah it is tricky to implement this once with the connection pool. I've moved the method comment closer to this implementation. 
The connection slots, which provide access to a single connection for a period of time only set this once.

modules/module-postgres-storage/README.md
```diff
@@ -0,0 +1,88 @@
...
+    # Maximum estimated byte size of operations in a single batch.
+    # Defaults to 5 megabytes.
+    max_estimated_size: 5000000
+    # Maximum number of records present in a single batch.
+    # Defaults to 2000 records.
+    # Increasing this limit can improve replication times for large volumes of data.
+    max_record_count: 2000
+    # Maximum byte size of size of current_data documents we lookup at a time.
+    # Defaults to 50 megabytes.
+    max_current_data_batch_size: 50000000
```
rkistner, line 87: I'd leave these out of the documentation to start with, until we can provide better guidelines on when it should be modified and what reasonable ranges are.

libs/lib-postgres/src/db/connection/DatabaseClient.ts
stevensJourney: The connection slots are currently used for providing transaction and notification functionality which I believe is not possible with the pgwire pool. The pgwire connection pool is used initial for basic individual statements, but it could be replaced with the custom pool if it's overkill to have 2 pools.

modules/module-postgres-storage/src/types/models/BucketParameters.ts
```diff
@@ -0,0 +1,17 @@
...
+import { pgwire_number } from '../../utils/ts-codec.js';
+import { bigint, jsonb } from '../codecs.js';
+import { SQLiteJSONRecord } from './SQLiteJSONValue.js';
+
+export const BucketParameters = t.object({
+  id: bigint,
+  group_id: pgwire_number,
+  source_table: t.string,
+  source_key: framework.codecs.buffer,
+  lookup: framework.codecs.buffer,
```
rkistner, line 12: Could you add comments to document the expected format that this is stored, and confirm it does work as expected?

I have specific concerns about bigint handling. Looking through the codecs, it looks like a bigint will be serialized as a string, and then stay a string when deserialized. But I'm also not sure how ts-codec handles arbitrary record types like these.

modules/module-postgres-storage/README.md
stevensJourney: A separate server is required. Using different databases on the same server still results in endless keepalive events.

modules/module-postgres-storage/src/storage/PostgresBucketStorageFactory.ts
```diff
@@ -0,0 +1,526 @@
...
+    }
+
+    const bucketData = await this.db.sql`
+      SELECT
+        --- This can differ from the octet_length
+        sum(pg_column_size(data)) AS operations_size_bytes
+      FROM
+        bucket_data
+      WHERE
+        group_id = ${{ type: 'int4', value: active_sync_rules.id }}
```
rkistner, line 120: These queries can be quite slow when the tables are large. Is there a faster way to get a size estimate for the entire table?

modules/module-postgres-storage/README.md
stevensJourney: 👍  removed this

modules/module-postgres-storage/README.md
stevensJourney: 👍 Updated the wording.

modules/module-postgres-storage/src/migrations/scripts/1684951997326-init.ts
stevensJourney: I've implemented `1` with no default.

modules/module-postgres-storage/src/storage/batch/PostgresBucketBatch.ts
stevensJourney: Thanks. Good catch. This one got a bit confusing since pgwire would return `bigint` type values for `integer` columns. I've changed this to `int4` with the relevant `number` Typescript types. I've used a codec to workaround pgwire returning a bigint.

libs/lib-postgres/src/locks/PostgresLockManager.ts
stevensJourney: Good catch. Updated this.

modules/module-postgres-storage/src/types/models/BucketParameters.ts
stevensJourney: Yeah, I understand how `ts-codec` can be a bit tricky to follow sometimes. I'm currently using `ts-codec` for:

**Decoding columns from pgwire**
I’m using ts-codec mainly to decode certain column values returned from pgwire. While pgwire already performs some conversion for common SQLite types, adding an extra decoding layer with ts-codec makes it easier to work with JS-native values. For instance, columns like `jsonb` are returned as JsonContainer/string instead of parsed objects. Similarly, fields like `group_id`, which are expected to be `Number`, are returned as `Bigint`. By automatically converting these, the resulting code feels much cleaner and more maintainable.

**JSON bulk inserts**
For persisted batch operations, JSON query parameters are used for bulk inserts. These payloads often need to be encoded into JSON-compatible values. Collections like `bucket_data`, `current_data`, and `bucket_parameters` all rely on encoded/decoded types, and `ts-codec` helps define these interfaces while keeping everything in sync when changes are made. I do manually encode some types to tap into intermediate values (e.g., for size estimation), but the final types are always validated using `ts-codec` to ensure consistency.

**Regarding the bigint codec**
You’re absolutely correct that the current `bigint` codec serializes values as strings. However, when decoding, those strings are reliably converted back into `Bigint`. You can see this [here](https://github.com/powersync-ja/powersync-service/blob/a06c715f9a2bd89016f029df83543f54f48af2bf/modules/module-postgres-storage/src/types/codecs.ts#L26). 

I have verified that the codec(s) work as expected. I can add some more comments to the types.

modules/module-postgres-storage/src/storage/PostgresBucketStorageFactory.ts
stevensJourney: I see I misread the MongoDB implementation. I assumed this was only for the active sync rules. If we want the entire table we can use `pg_total_relation_size`. Updated this to use `pg_total_relation_size` now.

libs/lib-postgres/src/locks/PostgresLockManager.ts
stevensJourney: Added a note to investigate after the initial release.

modules/module-postgres-storage/src/types/models/BucketParameters.ts
stevensJourney: I added some comments about the use of ts-codec [here](https://github.com/powersync-ja/powersync-service/blob/79da9db294fcb0b96d5052f78caaac0f6d3b4090/modules/module-postgres-storage/src/types/codecs.ts#L6). 
I've also rechecked all the decode operations. I made one cleanup for optimising the `Buffer` decoder/validator: the core framework implementation assumes a base64 encoded form while pgwire returns a `Uint8Array` and we encode to hex strings. This did not cause any issues previously since the decode operation did a `Buffer.from` operation and the encoder was not explicitly used.

modules/module-postgres-storage/src/storage/PostgresBucketStorageFactory.ts
```diff
@@ -0,0 +1,499 @@
...
+    });
+    this.slot_name_prefix = options.slot_name_prefix;
+
+    this.db.registerListener({
+      connectionCreated: async (connection) => this.prepareStatements(connection)
+    });
+    this.notificationConnection = null;
+  }
+
+  async [Symbol.dispose]() {
```
rkistner, line 74: This should use `Symbol.asyncDispose` - `Symbol.dispose` is for synchronous disposing.

modules/module-postgres-storage/src/storage/PostgresSyncRulesStorage.ts
```diff
@@ -0,0 +1,664 @@
...
+    const results = await this.db.sql`
+      WITH
+        filter_data AS (
+          SELECT
+            FILTER ->> 'bucket' AS bucket_name,
+            (FILTER ->> 'start')::BIGINT AS start_op_id,
+            (FILTER ->> 'end')::BIGINT AS end_op_id
+          FROM
+            jsonb_array_elements(${{ type: 'jsonb', value: rangedBatch }}::jsonb) AS FILTER
+        )
```
rkistner, line 629: This approach was mentioned as a performance issue for `getBucketDataBatch`, but not here? Is it just because these queries are not run as often?

Probably fine for the intial release, but we should properly investigate later.

modules/module-postgres-storage/src/storage/PostgresCompactor.ts
```diff
@@ -0,0 +1,366 @@
...
+          source_table,
+          table_name,
+          row_id,
+          source_key,
+          bucket_name
+        FROM
+          bucket_data
+        WHERE
+          group_id = ${{ type: 'int4', value: this.group_id }}
+          AND bucket_name LIKE COALESCE(${{ type: 'varchar', value: bucketLower }}, '%')
```
rkistner, line 129: Not sure if this will efficiently use the primary key index for filtering bucket names if specified. Not important for intial release.

modules/module-postgres-storage/src/storage/batch/PostgresBucketBatch.ts
```diff
@@ -0,0 +1,891 @@
...
+      return null;
+    }
+
+    logger.debug(`Saving ${record.tag}:${record.before?.id}/${record.after?.id}`);
+
+    if (!sourceTable.syncData && !sourceTable.syncParameters) {
+      return null;
+    }
+
+    logger.debug(`Saving ${record.tag}:${record.before?.id}/${record.after?.id}`);
```
rkistner, line 111: These lines are duplicated

modules/module-postgres-storage/src/storage/PostgresSyncRulesStorage.ts
```diff
@@ -0,0 +1,664 @@
...
+      WHERE
+        group_id = ${{ type: 'int4', value: this.group_id }}
+        AND lookup = ANY (
+          SELECT
+            decode((FILTER ->> 0)::text, 'hex') -- Decode the hex string to bytea
+          FROM
+            jsonb_array_elements(${{
+        type: 'jsonb',
+        value: lookups.map((l) => storage.serializeLookupBuffer(l).toString('hex'))
+      }}) AS FILTER
```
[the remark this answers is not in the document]
rkistner: Found my own answer to this. `FILTER` or `FILTER.value` would be `'"abc"'` here, instead of `'abc'`. To "parse" the JSON value, `->> 0` works on JSONB. If you're using the plain JSON type instead, use `#>> '{}'` instead (that difference confused me when testing in a postgres shell).

modules/module-postgres-storage/src/storage/PostgresBucketStorageFactory.ts
```diff
@@ -0,0 +1,499 @@
...
+    });
+  }
+
+  async slotRemoved(slot_name: string): Promise<void> {
+    const next = await this.getNextSyncRulesContent();
+    const active = await this.getActiveSyncRulesContent();
+
+    // In both the below cases, we create a new sync rules instance.
+    // The current one will continue erroring until the next one has finished processing.
+    // TODO: Update
```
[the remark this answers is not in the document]
stevensJourney: Not entirely sure. The Mongo implementation has the same [comment](https://github.com/powersync-ja/powersync-service/blob/d006e90fd91ce61e6a6bc89a4a89536344008512/modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts#L109) without much history. Can remove this.

modules/module-postgres-storage/src/storage/PostgresSyncRulesStorage.ts
```diff
@@ -0,0 +1,664 @@
...
+              table_name,
+              replica_id_columns
+            )
+          VALUES
+            (
+              ${{ type: 'varchar', value: uuid.v4() }},
+              ${{ type: 'int4', value: group_id }},
+              ${{ type: 'int4', value: connection_id }},
+              --- The objectId can be string | number, we store it as a string and decode when querying
+              ${{ type: 'varchar', value: objectId.toString() }},
```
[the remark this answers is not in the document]
stevensJourney: This one is slightly different from the previous note. In this case it was difficult to store either a number or string directly in a column, so I opted to store in a TEXT column. I wasn't sure if that would cause issues or not. I've updated this now to be stored in a JSON column, which should be able to preserve the original type.

modules/module-postgres-storage/src/storage/PostgresSyncRulesStorage.ts
```diff
@@ -0,0 +1,664 @@
...
+     * ORDER BY
+     * b.bucket_name ASC,
+     * b.op_id ASC
+     * LIMIT
+     * $4;
+     * ```
+     * Which might be better for large volumes of buckets, but in testing the JSON method
+     * was significantly slower than the method below. Syncing 2.5 million rows in a single
+     * bucket takes 2 minutes and 11 seconds with the method below. With the JSON method
+     * 1 million rows were only synced before a 5 minute timeout.
```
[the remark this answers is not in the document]
stevensJourney: Yeah that was very interesting. It seemed like the JOIN would scale better if more buckets are present, but when testing it seemed to be a large bottleneck for cases with small amounts of buckets.

modules/module-postgres-storage/src/storage/PostgresSyncRulesStorage.ts
stevensJourney: I suspect we probably could get some performance boost from changing this to the multiple "OR" method, but I haven't done those levels of optimisation yet.

modules/module-postgres-storage/src/storage/PostgresBucketStorageFactory.ts
stevensJourney: That is better. I've updated the `BucketStorageFactory` interface to use `AsyncDisposable` and all the relevant usages to use `await using`

modules/module-postgres-storage/src/storage/batch/PostgresBucketBatch.ts
stevensJourney: Removed now

modules/module-postgres-storage/src/storage/PostgresBucketStorageFactory.ts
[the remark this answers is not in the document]
stevensJourney: Removed these comments. Could not find a reason for it.

rkistner (dismissed):

rkistner (approved):

Chriztiaan (approved):

PR #186, opened 2025-01-28
[Postgres Storage] Ability to use the same server

stevensJourney:
# Overview

This PR adds support for using the same Postgres server as both the replication source and sync bucket storage. The same database is also supported, but this should be avoided if possible. This change aims to simplify self-hosting for certain users.

Currently, using the same server is not recommended due to the risk of WAL feedback loops. The Postgres replication module consumes a replication slot using the `pgoutput` plugin and emits `keepalive` events via `pg_logical_emit_message`. On older versions of Postgres, the contents of these `keepalive` events cannot be read, so the module assumes that any empty replication slot event is a `keepalive` event.

A feedback loop occurs as follows:

- The replicator writes data to the cluster, advancing the WAL LSN (Write-Ahead Log Log Sequence Number).
- This WAL progression generates a replication slot event.
- The event is misinterpreted as a keepalive event, prompting another write to the cluster.
- The cycle repeats.

Starting with Postgres 14, the `pgoutput` plugin provides access to the contents of logical replication messages. This allows the module to properly distinguish between keepalive events (which we emit) and other external WAL events, preventing feedback loops.

To maintain compatibility, an explicit error will be thrown if the same Postgres cluster is used for both the replication source and sync bucket storage on versions below 14. This ensures:

Postgres < 14 remains supported if separate clusters are used.
Postgres 14+ is required if using the same cluster for both purposes.

## Additional Items

Some minor Postgres cleanup is also included here. This includes:

- Cleanup for some Postgres persisted batch queries

Some bugs were identified:

During testing a bug was observed with the Postgres active checkpoint notifications. A crash could occur if a client subscribed to the notification stream after a connection had timed out. This issue has been resolved by cleaning up the listener flow and connection configuration.


modules/module-postgres-storage/src/storage/batch/PostgresBucketBatch.ts
```diff
@@ -529,31 +529,28 @@ export class PostgresBucketBatch
       const current_data_lookup = new Map<string, CurrentDataDecoded>();
       for await (const currentDataRows of db.streamRows<CurrentData>({
         statement: /* sql */ `
-          WITH
```
stevensJourney, line 532: This makes the query a bit easier to read, doesn't really have much of an impact on sync performance.

modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
```diff
@@ -84,6 +84,23 @@ export class MongoBucketStorage
...
+    let id = 'N/A';
+    try {
+      const result = await this.client
+        .db('local')
+        .collection('system.replset')
+        .findOne({}, { projection: { _id: 1 } });
+      id = result?._id.toString() ?? id;
+    } catch (ex) {
+      logger.warn(`Could not query MongoDB replica set id`, ex);
+    }
```
rkistner, line 97: You can use this approach to get the replicaset name, which doesn't require any priviledges:

https://github.com/powersync-ja/powersync-service/blob/08cf53247e36089be3dbf833812a389f38b338fe/modules/module-mongodb/src/replication/replication-utils.ts#L12-L20

I'd also recommend not catching the error - depending on what we use this for it can cause more issues if we silently return `N/A` rather than throwing an error. (I'm thinking use cases like detecting when the source db was switched).

rkistner (commented):
Apart from the mongodb system identifier, I'm happy with the changes. It looks like the MongoDB one is not _really_ used, but if it's there I think we need to make sure the implementation is working properly.

modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
stevensJourney: Thanks! I've updated to use the linked approach for fetching the replica set name. 
It seems like the replica set _name_ might not have the same cluster/server wide uniqueness that the Postgres system identifier has (I believe it can be manually specified when creating a replica set). The replica set Id could be more similar, but that does not seem to be available with this method. Other commands for fetching the replica set id seem to require elevated levels of permissions. 
I suppose the replica set name is okay for now.

libs/lib-postgres/src/db/connection/ConnectionSlot.ts
```diff
@@ -78,19 +76,6 @@ export class ConnectionSlot extends framework.DisposableObserver<ConnectionSlotL
     }
   }
 
-  registerListener(listener: Partial<ConnectionSlotListener>): () => void {
-    const dispose = super.registerListener(listener);
-    if (this.connection && this.hasNotificationListener()) {
-      this.configureConnectionNotifications(this.connection);
```
stevensJourney, line 84: This could cause an uncaught exception if a client subscribed to the active checkpoint stream after the underlaying Postgres connection was internally closed after a timeout. Always `listen`ing to the notifications on connect should solve that issue.

modules/module-mongodb-storage/src/storage/implementation/MongoBucketBatch.ts
```diff
@@ -654,6 +655,10 @@ export class MongoBucketBatch
       return false;
     }
 
+    if (!createEmptyCheckpoints && this.persisted_op == null) {
```
stevensJourney, line 658: The default for `createEmptyCheckpoints` is true (to match the previous behaviour), which should result in this not being reached in most circumstances.

modules/module-postgres-storage/README.md
```diff
@@ -31,7 +31,13 @@ storage:
...
+
+Separate Postgres servers are required for replication connections **if using PostgreSQL versions below 14**.
+
+| PostgreSQL Version | Server configuration                                                                                                                                                                                                     |
+| ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| Below 14           | Separate servers are required for the source and sync bucket storage. Replication will be blocked if the same server is detected.                                                                                        |
+| 14 and above       | The source database and sync bucket storage database can be on the same server. Using the same database (with separate schemas) is supported but may lead to higher CPU usage. Using separate servers remains an option. |
```
stevensJourney, line 40: Using the same database for the storage will result in many `begin` and `commit` replication slot events being emitted whenever data is written to the tables in the storage schema. These events don't contain any data operations between them, but we do still have to process them. 

Committing these empty transactions would result in a feedback loop where commits generate more transactions etc... The `createEmptyCheckpoints` setting should prevent committing these empty transactions.  This should avoid critical CPU usage, but generally the usage could increase just from using the same database. 

The behaviour of `createEmptyCheckpoints` is configured to only be `false` if using the same Postgres database is used for both the source and storage.

rkistner (approved):

PR #192, opened 2025-02-03
Bucket priorities

simolus3:
This implements the bucket priorities proposal. When buckets with different priorities are involved:

1. We only guarantee consistency within each priority (clients see changes from higher-priority buckets before changes from lower-priority buckets, even if there is a causal relationship between them).
2. (not a concern for the sync service): For the highes priority, clients are allowed to upload local changes before they have received a complete checkpoint.

Priorities are currently declared with a static literal column named `_priority` in a data query, e.g.

```yaml
bucket_definitions:
  global_todos:
    parameters: SELECT 0 as _priority;
    data:
      - SELECT * FROM todos
  global_lists: # implicit default priority 1
    data:
      - SELECT * FROM lists
```

At the moment, the implementation simply groups buckets into their priorities and then synchronizes each priority in a batch (instead of using a batch for all buckets like before).
An interesting improvement might be to interrupt lower-priority sync work when higher-priorities have new data. For instance, in a flow like:

1. New checkpoint sent to client.
2. No changed `todos` to send to client, immediately send partial complete message for priority 0.
3. Start syncing a large amount of `lists` rows.
4. Before being done with step 3, a new checkpoint with new `todos` comes in.

Here, we might want to stop sending rows for 4 (it's not that they're lost, the client has received the operations and we can later resume from that state) so that we can send the new checkpoint first.


.changeset/tall-peas-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/service-sync-rules': minor
```
simolus3, line 2: I've set this to minor because breaking bumps it to 1.0.0 - there are breaking changes.

packages/service-core/src/sync/sync.ts
```diff
@@ -193,27 +205,53 @@ async function* streamResponseInner(
...
-      tracker,
-      user_id: syncParams.user_id
-    });
+    // TODO: This is not really how bucket priorities are supposed to be implemented - we might want to listen for new write
+    // checkpoints while we're serving this one too. When we're syncing low-priority buckets of this checkpoint, we could interrupt
+    // this batch and serve the high-priority buckets of the new checkpoint first?
+    bucketsToFetch.sort((a, b) => a.priority - b.priority);
+    let firstBucketInSamePriority = 0;
+    let currentPriority = bucketsToFetch[0]!.priority;
+    const lowestPriority = bucketsToFetch.at(-1)!.priority;
```
rkistner, line 225: I didn't double-check, but I think `bucketsToFetch` could be empty, which would break this.

packages/service-core/src/sync/sync.ts
```diff
@@ -193,27 +205,53 @@ async function* streamResponseInner(
...
+      if (bucketsToFetch[i].priority == currentPriority) {
+        continue;
+      }
+
+      // This incrementally updates dataBuckets with each individual bucket position.
+      // At the end of this, we can be sure that all buckets have data up to the checkpoint.
+      yield* bucketDataWithPriority(i);
+      firstBucketInSamePriority = i;
+      currentPriority = bucketsToFetch[i].priority;
+    }
```
rkistner, line 252: The logic could perhaps be more clear if you group by priority first.

`Map.groupBy` would be great for that, but would need an upgrade to Node 22 first.

rkistner (commented):
Looks good overall! Added some minor comments on the implementation.

If I understand the implementation correctly, it will effectively change nothing if bucket priorities are not specified, which is great.

For specifying bucket priorities, could you also add support for this syntax?

```yaml
bucket_definitions:
  bulk_data:
    priority: 2
    parameters: select id as project_id from projects where ...
```

I think this would be simpler for the majority of cases, leaving the `as _priority` form only for cases where developers need more control.

rkistner:
After looking at the consistency implications in the client implementation, I think it would be better to make the default priority 3 instead of 1.

If the client gets a partial checkpoint, we only sync PUT operations, ignoring REMOVEs, which changes the consistency properties a little. I think that's a decent trade-off in many cases, but the developer needs to be aware of it.

Now imagine a developer starts with only having a priority 1 bucket (the default). Now they want to sync some bulk data as well, so they add a new bucket with priority 2. The issue is that this now affected the consistency properties of _bucket 1_, despite not touching it in the sync rules.

While if we instead have the default as priority 3 (the lowest), it means the developer will have to explicitly modify the original bucket to assign a higher priority, making the implications more obvious.

Thoughts?

simolus3:
> Now imagine a developer starts with only having a priority 1 bucket (the default). Now they want to sync some bulk data as well, so they add a new bucket with priority 2. The issue is that this now affected the consistency properties of _bucket 1_, despite not touching it in the sync rules.

I agree that making the lowest priority the default is a good choice to make sure changing the priority of some buckets doesn't affect unrelated buckets 👍

package.json
```diff
@@ -40,5 +41,13 @@
...
     "vitest": "^2.1.1",
     "ws": "^8.2.3"
+  },
+  "pnpm": {
+    "ignoredBuiltDependencies": [
+      "esbuild"
+    ],
+    "onlyBuiltDependencies": [
+      "esbuild"
+    ]
```
rkistner, line 51: What does this do?

packages/service-core/src/sync/sync.ts
```diff
@@ -103,138 +108,222 @@ async function* streamResponseInner(
...
   binary_data: boolean | undefined;
-  tracker: RequestTracker;
-  signal: AbortSignal;
+  /** Signals that the connection was aborted and that streaming should stop ASAP. */
+  abort_connection: AbortSignal;
+  /**
+   * Signals that higher-priority batches are available. The current batch can stop at a sensible point.
+   * This signal also fires when abort_connection fires.
+   */
+  abort_batch: AbortSignal;
```
rkistner, line 318: I realize in theory these two _could_ be different, but in practice they both just abort the query between batches, and I'm not sure that would ever change. Should we perhaps just make it one signal here?

packages/service-core/src/sync/sync.ts
```diff
@@ -103,138 +108,222 @@ async function* streamResponseInner(
...
+        };
+        yield checksum_line;
+      }
+      lastChecksums = checksumMap;
+      lastWriteCheckpoint = writeCheckpoint;
+
+      // Start syncing data for buckets up to the checkpoint. As soon as we have completed at least one priority and
+      // at least 1000 operations, we also start listening for new checkpoints concurrently. When a new checkpoint comes
+      // in while we're still busy syncing data for lower priorities, interrupt the current operation and start syncing
+      // the new checkpoint.
```
rkistner, line 236: > As soon as we have completed at least one priority

Wouldn't that negate most of the advantages here? I.e. if we're busy syncing bulk data with a low priority, that may never be interrupted for a new checkpoint with higher-priority buckets.

packages/service-core/src/sync/sync.ts
simolus3: I'm fine with just using a single signal, but I want to point out a minor practical difference: `abort_connection` would also drop data _after_ we've waited for a chunk in `getBucketDataBatch` to be emitted (check at the beginning for the `for await` loop in `bucketDataBatch`). We will never ignore rows we already have when `abort_batch` is aborted though, we just check that signal to not ask for more data.

There's an inherent race condition between a new chunk in `getBucketDataBatch()` and `watchWriteCheckpoint`, but that's by design because we want to interrupt syncs where possible.
We have the choice of either emitting that chunk (current implementation, no wasted resources, but higher latency until the clients receive the next checkpoint after the interruption) or throwing it away (in which case we could use a single abort signal).

packages/service-core/src/sync/sync.ts
simolus3: That's a good point, I think we should just check for 1000 rows synced then.

My original idea (which this check doesn't implement accurately) was that we shouldn't interrupt high-priority sync operations when we receive a new checkpoint that might only cover lower-priority buckets.
I'm not sure if we should still check this separately now or if it doesn't matter in practice because we expect high-priority buckets to be generally smaller (and would still have the 1000 row limit before we listen for a new checkpoint).

packages/service-core/src/sync/sync.ts
rkistner: Ok, that makes sense. Happy to keep as-is then.

rkistner (approved):
Happy to merge this.

As mentioned offline, there is a fix in https://github.com/powersync-ja/powersync-service/pull/197/commits/f60c7059a8a6379cd0c4b8f0fb31a67aa8a931bd that should be included before releasing this.

PR #193, opened 2025-02-04
Fix -> operator on missing values to return null

rkistner:
Using the `->` operator with a missing value, for example `select '{}' -> 'test'`, would return `undefined` instead of `null`. We use SQLite types, so there should never be an `undefined` value.

This was specifically an issue when used together with `json_each`, for example `SELECT value as project_id FROM json_each(request.jwt() -> 'project_ids')`. This query would throw an error `Expected json_each to be called with a string, got undefined`, instead of treating it like an empty array like with `null` values.

This fixes `select '{}' -> 'test'` to return `null` correctly.

This also adds tests for all the following forms:

```sql
select '{"t": null}' -> 't' -- null in PowerSync, 'null' in SQLite
select '{}' -> 't' -- null in PowerSync and SQLite, fixed by this PR

select '{"t": null}' ->> 't' -- 'null' in PowerSync, null in SQLite
select '{}' ->> 't' -- null in PowerSync and SQLite

select json_extract('{"t": null}', '$.t') -- 'null' in PowerSync, null in SQLite
select json_extract('{}', '$.t') -- null in PowerSync and SQLite
```

Notice the difference in handling of `{"t": null}` values between PowerSync and SQLite. That is not affected by this PR though, and fixing it to match SQLite would be a breaking change, so keeping it as-is for now.

### DIFF
--- a/packages/sync-rules/src/sql_functions.ts
+++ b/packages/sync-rules/src/sql_functions.ts
@@ -865,7 +865,7 @@ export function jsonExtract(sourceValue: SqliteValue, path: SqliteValue, operato
     value = value[c];
   }
   if (operator == '->') {
-    // -> must always stringify, expect when it's null
+    // -> must always stringify, except when it's null
     if (value == null) {
       return null;
     }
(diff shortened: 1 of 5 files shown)


packages/sync-rules/src/sql_functions.ts
```diff
@@ -865,7 +865,10 @@ export function jsonExtract(sourceValue: SqliteValue, path: SqliteValue, operato
     value = value[c];
   }
   if (operator == '->') {
-    // -> must always stringify
+    // -> must always stringify, expect when it's null
```
stevensJourney, line 868: Nit: Should this be `except`?

stevensJourney (approved):

stevensJourney (approved):

PR #195, opened 2025-02-05
Upgrade typescript and target ES version

simolus3:
I've extracted this into a separate PR to reduce the size of the bucket priorities PR. This changes our packages to target ES2024-compatible runtimes, and also updates TypeScript to 5.7 for ES2024 support.

A change is that `Buffer` is a generic type in 5.7, that change was added to express the API differences between shared and non-shared array buffers. It looks like we have to consistently use `Buffer<ArrayBuffer>` instead of the raw `Buffer` type to avoid compilation errors.


rkistner (approved):

simolus3:
Do we need a version bump for this?

rkistner:
I think a changeset should be done either way, to ensure new versions any changed packages are published in the next release, even though we're not releasing yet. Should probably be a minor version bump for packages depending on the new node version.

rkistner (approved):

PR #197, opened 2025-02-07
Optimize incremental sync: Phase 1

rkistner:
This refactors large parts of the sync flow to allow better optimizations.

Refactoring:
1. `watchWriteCheckpoint` is moved from `BucketStorageFactory` to `SyncRulesBucketStorage`, making it operate on a single sync-rules instance/version only. This simplifies the implementation and will make it much easier to do optimizations here, since we don't have to deal with sync rule changes.
    * When the active sync rules change, this is detected, and this stops the stream, which closes the connection to the client. The client then has to reconnect to sync against the new sync rules version.
    * This is different from before where we kept connections open during sync rule changes. However, since we have to re-sync all data anyway when sync rules change, this is not a significant difference.
2. In `sync.ts`, the bucket parameter state management is moved to a separate class, keeping the main loop simpler.
3. For looking up bucket parameters, this is now split into two steps:
    1. Pre-compute the static buckets and lookup values. This is now only done once per connection, instead of for each new checkpoint.
    2. Query the "dynamic buckets" (parameter queries on tables). This is now only done if there are dynamic buckets.
    3. This does not have a big impact by itself, but will allow big optimizations in the next phase.

The biggest actual optimization in this phase is removing `fullDocument: 'uploadLookup'` from the changestream for watching checkpoints. Since this changeStream could be updated _very_ frequently, this can add significant overhead on the bucket storage database. We now instead merge the incremental updates ourselves.

Builds on some API refactoring in #192.

---

I added a new test-client command to test concurrent connections:
```
pnpm concurrent-connections -n 20 -c ../service/powersync.yaml
```

This just opens connections and keep them open to simulate sync load. No direct reporting of performance is done yet, but you can view the timestamp differences between the first `checkpoint_diff` and the last `checkpoint_complete` operations to get an idea of performance.


modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
```diff
@@ -289,19 +268,22 @@ export class MongoBucketStorage
...
 
-    return this.makeActiveCheckpoint(doc);
+    // It is important that this instance is cached.
+    // Not for the instance construction itself, but to ensure that internal caches on the instance
+    // are re-used properly.
+    if (this.activeStorageCache?.group_id == content.id) {
+      return this.activeStorageCache;
+    } else {
+      const instance = this.getInstance(content);
+      this.activeStorageCache = instance;
```
stevensJourney, line 284: Should we `dispose` the previous `activeStorageCache` before assigning a new reference?

modules/module-mongodb/test/src/change_stream_utils.ts
```diff
@@ -147,14 +147,15 @@ export async function getClientCheckpoint(
...
 
   const timeout = options?.timeout ?? 50_000;
-  let lastCp: ActiveCheckpoint | null = null;
+  let lastCp: ReplicationCheckpoint | null = null;
 
   while (Date.now() - start < timeout) {
-    const cp = await bucketStorage.getActiveCheckpoint();
-    lastCp = cp;
-    if (!cp.hasSyncRules()) {
+    const storage = await storageFactory.getActiveStorage();
```
stevensJourney, line 153: Should this be `using storage`?

modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
rkistner: I'm still trying to wrap my head around disposing of SyncRulesBucketStorage. Especially when used in the sync API, the lifetime of the instance is not well-defined. Even when the instance is not in the `activeStorageCache` anymore, there could still be references to it being used.

In the case of the sync APIs, it is fine if references are still being used after disposing, since disposing only affect listeners, and listeners are not relevant for the sync API.

@stevensJourney What is the practical effect of disposing when used for replication? What will happen if we never dispose the instance?

modules/module-mongodb/test/src/change_stream_utils.ts
rkistner: Similar to my above comment, the lifecycle around disposing these is difficult. But in this specific case, disposing it would be wrong, since the same instance would still be cached for the next call.

modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
stevensJourney: Discussed offline: My primary concern was with the custom write checkpoint module, which registered listeners on the sync bucket storage. I initially thought the callbacks for those listeners might hold references to the sync bucket storage, potentially preventing garbage collection. However, after further review, it appears that we don't track any references in those listeners.

stevensJourney (approved):
Looks good to me. Havent tested anything.

simolus3 (dismissed):
I've mostly looked at the sync changes, and they look good to me 👍

stevensJourney (approved):

PR #198, opened 2025-02-10
Exit on sync rule validation error

rkistner:
Currently, if sync rules fail to validate (e.g. due to a SQL syntax error), a message is logged, and the error is otherwise ignored. This makes it easy to miss the issue, and not understand why the new sync rules are applied.

See the issue logged here: https://github.com/powersync-ja/powersync.dart/issues/234

This changes the validation behavior to one of two options:
1. `exit_on_error: true` (the default for self-hosting): The replication process will exit with an error on startup when the sync rules do not pass validation. This should be easy to detect during self-hosting.
2. `exit_on_error: false` (will be used for cloud version): This loads the new sync rules file even if validation failed. The replication process will catch the error, and report it (1) in the logs every 30 seconds or so, and (2) in the diagnostics API, visible on the cloud dashboard.

The option is in the sync_rules section:
```yaml
sync_rules:
  path: sync-rules.yaml
  exit_on_error: false
```

This new default `exit_on_error: true` behavior may be considered a breaking change for self-hosted, but would only apply when there is already a sync rules issue (and will now properly surface the issue), so I think it's worth making this breaking change.


stevensJourney (approved):
Looks good to me :)

PR #214, opened 2025-02-24
Fix missing completion when no buckets are in checkpoint

simolus3:


packages/service-core/src/sync/sync.ts
```diff
@@ -198,7 +207,7 @@ async function* streamResponseInner(
           user_id: syncParams.user_id,
           // Passing undefined will emit a full sync complete message at the end. If we pass a priority, we'll emit a partial
           // sync complete message.
-          forPriority: !isLast ? priority : undefined
+          forPriority: !isLast ? priority : null
```
rkistner, line 210: The comment above mentions "Passing undefined will...". I assume that also applies to `null`?

simolus3: It does, I've updated the comment.

rkistner (approved):

PR #215, opened 2025-02-24
Fix truncating collections from MongoDB source database

rkistner:
Specific scenario is when all of the following conditions are met:
1. We're replicating from an Atlas Flex instance.
2. There were changestream events recorded while the PowerSync service is paused.
3. We're only replicating from a single database.

Then we've observed an ns with for example `{db: '67b83e86cd20730f1e766dde_ps'}`, instead of the expected `{db: 'ps'}` on the change stream.

This caused issues on multiple levels internally:
1. We only used the collection name as the "relation_id" for the internal "source table" (bug 1).
2. The storage module then detected the equivalent of a renamed collection, since it saw the same relation_id, but a different schema.
3. We then truncate the table, but never did a snapshot of the "new" collection (bug 2).
4. This resulted in the collection being missing, apart from documents specifically updated again.

This fixes the issue on multiple levels:
1. Remove the `relation_id` for MongoDB source databases, and only rely on the db and collection names directly.
     1. This was simpler than consistently migrating the `relation_id`.
     2. We should probably do the same for MySQL, but that would require more testing.
     3. This means we can now correctly handle multiple collections of the same name in different dbs.
     4. This would avoid the "truncating" issue from above.
2. If we do see a new collection, snapshot it. While we don't expect this to happen, it will allow for more graceful handling of edge cases if we run into any again.
3. Implement a workaround for the incorrect `ns` from Flex instances.
4. Add proper logging for operations like truncating collections, for if we do run into scenarios like these again.

---

There is another report with similar symptoms, but not using a Flex instance. See [this thread](https://discord.com/channels/1138230179878154300/1343243467350081566).

### DIFF
--- a/.changeset/rare-birds-peel.md
+++ b/.changeset/rare-birds-peel.md
@@ -1,0 +1,9 @@
+---
+'@powersync/service-module-postgres-storage': patch
+'@powersync/service-module-mongodb-storage': patch
+'@powersync/service-module-mongodb': patch
+'@powersync/service-core': patch
+'@powersync/service-module-mysql': patch
+---
+
+Improve handling of some edge cases which could trigger truncating of synced tables.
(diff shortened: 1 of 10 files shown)


stevensJourney (approved):
LGTM. Tested MongoDB locally and verified with a Flex instance.

PR #218, opened 2025-02-26
Configurable service limits

rkistner:
Add configurable parameters, with defaults matching the previous behavior:
1. Max number of buckets and max number of parameter query results. Currently both limited to 1000 to match previous behavior, but parameter query results can have a higher limit in the future.
2. Max number of connections (default 200).
3. Max data fetch concurrency (default 10).
4. Max storage db pool size (default 8).

The general idea is to allow serving more connections per process/container, for use cases where the number of buckets and data volumes are smaller.


stevensJourney (approved):
Happy with the changes

stevensJourney (approved):

PR #221, opened 2025-03-03
Metrics refactor

Rentacookie:
Refactored metrics to use a MetricsEngine which forms part of the service context.
Metrics and their creation are abstracted away behind interfaces decoupled from Opentelemetry.
Metrics can be created and accessed via the MetricsEngine.

Moved the metrics initialisation code out of the service and into service core. That way the initialisation code can be re-used by the powersync cloud edition.


modules/module-postgres/src/module/PostgresModule.ts
```diff
@@ -29,9 +30,7 @@ export class PostgresModule extends replication.ReplicationModule<types.Postgres
     });
   }
 
-  async initialize(context: system.ServiceContextContainer): Promise<void> {
-    await super.initialize(context);
-
+  async onInitialized(context: system.ServiceContextContainer): Promise<void> {
```
Rentacookie, line 33: I made this change after discovering a bug where the initialize code would still run even if the module was not supposed to be active. This is because the escape clause is in the parent `initialize` function, so the standard initialize would still run after the `super.initialize(context)` was called. 
The main initialize method now instead calls `onInitialized` only when the module is actually supposed to be active.

modules/module-postgres/src/module/PostgresModule.ts
```diff
@@ -46,13 +45,11 @@ export class PostgresModule extends replication.ReplicationModule<types.Postgres
...
-    if (context.metrics) {
-      jpgwire.setMetricsRecorder({
-        addBytesRead(bytes) {
-          context.metrics!.data_replicated_bytes.add(bytes);
-        }
-      });
-    }
+    jpgwire.setMetricsRecorder({
+      addBytesRead(bytes) {
+        context.metricsEngine.getCounter(ReplicationMetric.DATA_REPLICATED_BYTES).add(bytes);
```
Rentacookie, line 50: Moving the initialized MetricsEngine into the ServiceContext solved the bug where the replicated bytes metric was always 0. Previously the metrics in the context were not initialized yet at this point.

dylanvorster (dismissed):
skimmed, but happy with the patterns and architecture used. Nice work!

modules/module-mongodb/test/src/env.ts
```diff
@@ -7,5 +7,5 @@ export const env = utils.collectEnvironmentVariables({
   CI: utils.type.boolean.default('false'),
   SLOW_TESTS: utils.type.boolean.default('false'),
   TEST_MONGO_STORAGE: utils.type.boolean.default('true'),
-  TEST_POSTGRES_STORAGE: utils.type.boolean.default('true')
+  TEST_POSTGRES_STORAGE: utils.type.boolean.default('false')
```
stevensJourney, line 10: I kind of like the exposure of testing the Postgres storage with this module. This can also be disabled locally by specifying `false` in the `.env` file.

packages/service-core/src/metrics/open-telemetry/util.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export interface RuntimeMetadata {
+  [key: string]: string | number | undefined;
+}
+
+export function createOpenTelemetryMetricsFactory(context: ServiceContext): MetricsFactory {
+  const { configuration, lifeCycleEngine, storageEngine } = context;
+  const configuredExporters: MetricReader[] = [];
+
+  if (env.METRICS_PORT) {
```
stevensJourney, line 18: Could we move this to the collected configuration?

stevensJourney (dismissed):
Left some minor comments. Overall this looks brilliant.

modules/module-mongodb/test/src/env.ts
Rentacookie: Thanks for picking this up - this must have slipped in with a merge.

dylanvorster (approved):

PR #133, opened 2024-11-14
Module Replication: MongoDB & MySQL support

rkistner:
This is a major refactor to support modular replication, and adds initial support for MongoDB and MySQL source databases.

Opening the PR to start the process of getting this into the stable builds.

PR #226, opened 2025-03-11
[Postgres] Fix replication slot name

rkistner:
This appears to be an oversight from the refactoring in https://github.com/powersync-ja/powersync-service/pull/133 - the `slot_name_prefix` was not respected anymore. We specifically use this in the cloud version to include the instance id in the slot name, making it easier to identify powersync instances connected to a specific postgres instance.

The actual structure behind this config option is still not ideal, but this does not make it any worse.


stevensJourney (approved):
🙏

PR #228, opened 2025-03-13
[Postgres] Optimize transaction replication throughput

rkistner:
Previously, if we got 100k transactions each with a single operation, we'd commit/flush that 100k times to the bucket storage. Flushing is slow - in the case of a storage cluster under load, this can take 100-500ms per flush, resulting in throughput of less than 10 transactions per second.

Lucky for us, Postgres already chunks messages together in the replication stream. So now we look ahead in the current chunk to see if there are any more commit messages. If there are, we only flush/commit on the _last_ one. This means if we get many transactions in a single chunk, they are all batched together in a single flush to the bucket storage.

For reference, we already have similar behavior in our MongoDB replication implementation.

In theory we could also batch transactions across multiple replication chunks, but I don't expect significant further gains from that, and it could increase complexity and memory usage.

This also adds Postgres 17 to the test matrix (released Sept 2024).

### DIFF
--- a/.changeset/chilly-flowers-fix.md
+++ b/.changeset/chilly-flowers-fix.md
@@ -1,0 +1,7 @@
+---
+'@powersync/service-module-postgres': minor
+'@powersync/service-core': minor
+'@powersync/service-image': minor
+---
+
+[Postgres] Only flush once per replicated chunk, increasing transaction replication throughput.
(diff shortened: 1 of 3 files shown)


stevensJourney (approved):
Looks like a nice optimisation to me :)

PR #230, opened 2025-03-14
Optimize write checkpoints lookups

rkistner:
Builds on #200.

This now uses a change stream to get write checkpoint updates, instead of polling. This is the last big remaining optimization required to efficiently scale incremental updates on a large number of concurrent connections.

Write checkpoints are different from bucket and parameter data, since they do not follow any op_id sequence, so we use a different approach from those. Write checkpoint write load is also expected to be much lower than bucket and parameter data, so the additional changestream should not have much overhead.

This shares a single change stream between all connections, similar to sync rules. This cannot use the same change stream as sync rules, since we need `fullDocument: 'updateLookup'` here, and don't want that overhead for sync rules updates. Note that change streams can be multiplexed over the same connection, so this should not increase our overall connection count.

This optimization is currently only implemented for MongoDB storage, not for Postgres storage.

This also fixes an issue with custom write checkpoints with Postgres storage (custom write checkpoints failed to persist).


stevensJourney (approved):
Looks good to me. Thanks for fixing the custom write checkpoint issue in the Postgres storage.

PR #234, opened 2025-03-25
MySQL: Reworked Varchar/Varbinary column identification

Rentacookie:
MySQL uses the same type code for Varbinary and Varchar columns. This then requires additional interrogation of the column metadata to correctly determine the type. Previously we just checked the binary flag in the column metadata to determine if a column was in fact Varbinary. But it turns out Varchar columns with binary encoding set will also have the binary flag set, leading to mis-identifcation. Instead we now check the characterSet for the column for the binary characterSet which seems to only be set for actual Varbinary columns

### DIFF
--- a/modules/module-mysql/test/src/mysql-to-sqlite.test.ts
+++ b/modules/module-mysql/test/src/mysql-to-sqlite.test.ts
@@ -40,6 +40,7 @@ describe('MySQL Data Types', () => {
 
     char_col CHAR(10),
     varchar_col VARCHAR(255),
+    varchar_binary_encoding_col VARCHAR(255) COLLATE utf8mb4_bin,
     binary_col BINARY(16),
     varbinary_col VARBINARY(256),
     tinyblob_col TINYBLOB,
@@ -145,6 +146,7 @@ INSERT INTO test_data (
 INSERT INTO test_data (
     char_col,
     varchar_col,
+    varchar_binary_encoding_col,
     binary_col,
     varbinary_col,
     tinyblob_col,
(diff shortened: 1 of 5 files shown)


modules/module-mysql/test/src/mysql-to-sqlite.test.ts
```diff
@@ -40,6 +40,7 @@ describe('MySQL Data Types', () => {
 
     char_col CHAR(10),
     varchar_col VARCHAR(255),
+    varchar_binary_encoding_col VARCHAR(255) COLLATE utf8mb4_bin,
```
Rentacookie, line 43: Specifying COLLATE for a column without specifying the character set defaults the column characterset to the one for COLLATE:
https://dev.mysql.com/doc/refman/5.7/en/charset-column.html

rkistner, line 43: `binary_encoding` doesn't sound quite correct here. As I understand it, this is utf8 encoding, but comparison is byte-wise.

rkistner: Could we also add a test for something like `VARCHAR(255) CHARACTER SET binary`?

Rentacookie: Thanks - the edge case `VARCHAR(255) CHARACTER SET binary` highlighted a problem that there it not enough info in the metadata to correctly deduce the type. 
Looks like the only way is to query the table schema directly.

Rentacookie: Turns out `VARCHAR(255) CHARACTER SET binary` is essentially just a `Varbinary` column, therefore we parse it as one.

rkistner (approved):

PR #239, opened 2025-04-04
Dont interrupt sync for unrelated checkpoint

simolus3:
This restores support for interrupting checkpoints (the functionality was disabled in https://github.com/powersync-ja/powersync-service/pull/236 due to a flaw where new checkpoints on unrelated bucket would interrupt sync).

To restore this functionality, we're expanding the work we're doing while the previous sync iteration is active: Instead of just waiting for a new checkpoint from storage, we're also building the new checkpoint line preemptively. If the new checkpoint does not result in a `checkpoint` line for the connection, we do nothing and keep waiting for the next checkpoint. 

This should be safe because the stateful logic to build checkpoint lines does not update its internal state if no line ends up being built.

### DIFF
--- a/packages/service-core/src/sync/BucketChecksumState.ts
+++ b/packages/service-core/src/sync/BucketChecksumState.ts
@@ -6,7 +6,6 @@ import * as util from '../util/util-index.js';
 import { ErrorCode, logger, ServiceAssertionError, ServiceError } from '@powersync/lib-services-framework';
 import { JSONBig } from '@powersync/service-jsonbig';
 import { BucketParameterQuerier } from '@powersync/service-sync-rules/src/BucketParameterQuerier.js';
-import { BucketSyncState } from './sync.js';
 import { SyncContext } from './SyncContext.js';
 import { getIntersection, hasIntersection } from './util.js';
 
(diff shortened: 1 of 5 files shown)


packages/service-core/src/sync/BucketChecksumState.ts
```diff
@@ -63,6 +63,15 @@ export class BucketChecksumState {
     }
   }
 
+  /**
+   * Build a new checkpoint line for an underlying storage checkpoint update if any buckets have changed.
+   *
+   * When this returns null, the call is idempotent (meaning that no internal state is updated until a line
+   * is generated).
```
rkistner, line 70: Is there a way to restructure this function so we can better "guarantee" this behavior? E.g. splitting it into one idempotent function, and separate one to update the state.

I went through the function again, and there is quite a lot happening before null is returned, including updating `this.bucketDataPositions`, which feels risky.

rkistner (dismissed):

packages/service-core/src/sync/sync.ts
```diff
@@ -187,7 +207,7 @@ async function* streamResponseInner(
         // 3. However, the new checkpoint does not contain any new data for this connection, so nothing further is synced.
```
stevensJourney, line 207: I might be missing something here, but is this comment still valid? It seems like we are no longer disabling interrupting checkpoints.

simolus3: Good catch, the comment should be removed.

PR #244, opened 2025-04-19
Fix deleteMany for bucket_parameters to use an index

rkistner:
The `deleteMany` for clearning bucket_parameters data used a filter like this:

```js
     "key": {
        "$gte": {
          "g": 3,
          "t": MinKey(),
          "k": MinKey()
        },
        "$lt": {
          "g": 3,
          "t": MaxKey(),
          "k": MaxKey()
        }
      }
```

This followed the same pattern as the _id filters on other collections. However, there is no direct index on `key`, only this index:

```js
await database.bucket_parameters.createIndex(
      {
        'key.g': 1,
        lookup: 1,
        _id: 1
      },
      { name: 'lookup1' }
    );
```

So in this case we can just use a simpler filter on `key.g`, to achieve the same thing but using an index.

### DIFF
--- a/modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
@@ -571,7 +571,7 @@ export class MongoSyncBucketStorage
     );
     await this.db.bucket_parameters.deleteMany(
       {
-        key: idPrefixFilter<SourceKey>({ g: this.group_id }, ['t', 'k'])
+        'key.g': this.group_id
       },
       { maxTimeMS: lib_mongo.db.MONGO_CLEAR_OPERATION_TIMEOUT_MS }
     );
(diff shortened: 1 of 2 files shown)


kobiebotha (approved):

modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
```diff
@@ -571,7 +571,7 @@ export class MongoSyncBucketStorage
     );
     await this.db.bucket_parameters.deleteMany(
       {
-        key: idPrefixFilter<SourceKey>({ g: this.group_id }, ['t', 'k'])
+        'key.g': this.group_id
```
Rentacookie, line 574: Would a simpler filter + index like this not perhaps help with the clearing of the `bucket_data` collection. 
I had a look and even though the index there is being used it still looked to be rather slow.

rkistner: Adding another index on bucket_data can significantly affect the performance for all writes there, and even make deletes slower because of that. The current filter is slightly complex, but it does efficiently use the _id index. I suspect the performance bottleneck is to a large extend just an overall write throughput bottleneck (although I haven't tested that properly).

PR #250, opened 2025-04-27
Update normalizeMongoConfig function and its tests to allow replica URI

whygee-dev:


libs/lib-mongodb/src/types/types.ts
```diff
@@ -33,29 +42,36 @@ export function baseUri(options: BaseMongoConfig) {
...
     throw new ServiceError(ErrorCode.PSYNC_S1105, `MongoDB connection: database required`);
   }
 
-  delete uri.userinfo;
-
   const lookupOptions: LookupOptions = {
     reject_ip_ranges: options.reject_ip_ranges ?? []
   };
-  const lookup = makeHostnameLookupFunction(uri.host ?? '', lookupOptions);
+  const lookup = makeHostnameLookupFunction(uri.hosts[0] ?? '', lookupOptions);
```
rkistner, line 71: This will need to be extended to validate all hosts instead of just the first one, but we can handle that after this PR is merged.

libs/lib-mongodb/package.json
```diff
@@ -31,6 +31,7 @@
     "@powersync/lib-services-framework": "workspace:*",
     "bson": "^6.10.3",
     "mongodb": "^6.14.1",
+    "mongodb-connection-string-url": "^3.0.2",
```
rkistner, line 34: We can remove uri-js below now - this was the only place it was used in this package.

rkistner (commented):
Thanks for the changes and the comprehensive tests!

For this repo we do require a signed CLA for any pull requests to be able to merge it - are you able to do that?

rkistner (approved):
Thanks!

I fixed an issue with the lookup function (see the tests on github actions failing before) - the rest looks good to me.

stevensJourney (approved):
Looks good to me

PR #251, opened 2025-04-30
[MongoDB] Fix resume token handling when no events are received

rkistner:
Our current implementation used resume tokens received from change stream events to resume the stream after a restart.

However, if the source collections do not receive writes, we do not receive any actual events, and the token is never advanced. If the source cluster receives many other oplog events, resuming the stream can become very slow after a while, or time out completely. Once it starts timing out, we don't have a way to recover other than restarting replication from scratch. The timeout typically shows up as an error "connection 2 to 159.41.94.47:27017 timed out". The error can be simulated by reducing `socketTimeoutMS` significantly (say 50ms).

MongoDB does however provide new resume tokens even when there are no events:
https://www.mongodb.com/docs/manual/changeStreams/#std-label-change-stream-resume-token

> The [$changeStream](https://www.mongodb.com/docs/manual/reference/operator/aggregation/changeStream/#mongodb-pipeline-pipe.-changeStream) aggregation stage includes a resume token on the cursor.postBatchResumeToken field.
> The [getMore](https://www.mongodb.com/docs/manual/reference/command/getMore/#mongodb-dbcommand-dbcmd.getMore) command includes a resume token on the cursor.postBatchResumeToken field.

The MongoDB driver exposes these as `stream.resumeToken`. There are some implementation details on exactly when this is updated, but in general it is safe to use that when `stream.tryNext()` returned null.

This will avoid the case of the resume token falling to far behind, at least in the case where the service is running. For self-hosted cases, when the service is stopped for a while (or when there were connection errors for a while), it is still possible to run into the timeout. This doesn't implement automatic recovery from the timeout yet (since it is difficult to know whether the timeout is permanent or can just be retried), but does improve the error message.

### DIFF
--- a/modules/module-mongodb/test/src/resume_token.test.ts
+++ b/modules/module-mongodb/test/src/resume_token.test.ts
@@ -1,0 +1,37 @@
+// Write tests for parseResumeTokenTimestamp
+import { parseResumeTokenTimestamp } from '@module/common/MongoLSN.js';
+import { mongo } from '@powersync/lib-service-mongodb';
+import { describe, expect, it } from 'vitest';
+
+describe('parseResumeTokenTimestamp', () => {
+  it('parses a valid resume token (1)', () => {
+    const timestamp = parseResumeTokenTimestamp({ _data: '826811D298000000012B0429296E1404' });
+    expect(timestamp.t).toEqual(1745998488);
+    expect(timestamp.i).toEqual(1);
+  });
+
+  it('parses a valid resume token (2)', () => {
+    const timestamp = parseResumeTokenTimestamp({
+      _data:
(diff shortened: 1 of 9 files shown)


modules/module-mongodb/test/src/resume_token.test.ts
```diff
@@ -0,0 +1,37 @@
+// Write tests for parseResumeTokenTimestamp
```
Rentacookie, line 1: Unused comment here ;)

packages/service-errors/src/codes.ts
```diff
@@ -253,6 +253,22 @@ export enum ErrorCode {
    */
   PSYNC_S1344 = 'PSYNC_S1344',
 
+  /**
+   * Failed to read MongoDB Change Stream to to a timeout.
```
Rentacookie, line 257:
```suggestion
   * Failed to read MongoDB Change Stream due to a timeout.
```

Rentacookie (dismissed):
This must have been challenging to debug 😬

Rentacookie (approved):

PR #252, opened 2025-04-30
Fix missing parameter storage metrics

Rentacookie:
The storage metric `powersync_parameter_storage_size_bytes` was missing when there were not any parameter queries in the sync rules.

### DIFF
--- a/packages/service-core/src/storage/storage-metrics.ts
+++ b/packages/service-core/src/storage/storage-metrics.ts
@@ -34,7 +34,7 @@ export function initializeCoreStorageMetrics(engine: MetricsEngine, storage: Buc
   let cacheTimestamp = 0;
 
   const getMetrics = () => {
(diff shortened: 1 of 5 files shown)


packages/service-core/src/storage/storage-metrics.ts
```diff
@@ -47,21 +47,21 @@ export function initializeCoreStorageMetrics(engine: MetricsEngine, storage: Buc
...
-      return metrics.operations_size_bytes;
+      return metrics.operations_size_bytes ?? 0;
     }
   });
 
   parameter_storage_size_bytes.setValueProvider(async () => {
     const metrics = await getMetrics();
     if (metrics) {
-      return metrics.parameters_size_bytes;
+      return metrics.parameters_size_bytes ?? 0;
```
Rentacookie, line 64: As best I could tell, `metrics.parameters_size_bytes` returns `NaN` when there are no parameter queries in the sync rules.

rkistner (approved):

PR #254, opened 2025-05-05
Health Check Improvements

stevensJourney:
# Overview

The PowerSync service uses a probing system to communicate health check statuses. Internally, the probe state is updated when the service becomes ready or active. This state is maintained in memory and can be exposed externally via the filesystem or HTTP endpoints.

Previously, probe state was always exposed via the filesystem by writing timestamps to .probes/startup, .probes/ready, and .probes/poll. This required the service to run with write access to the filesystem, which isn't always feasible or recommended. There was no way to disable filesystem probes.

In addition, the same probe state was exposed via HTTP routes, but only if the RouterEngine started an HTTP server. This occurred only in API and UNIFIED modes. In SYNC mode, the only way to access probe state externally was through the filesystem.

## What's new
This PR introduces configurable probe types through the PowerSync configuration file. You can now enable or disable probe outputs via YAML:

yaml
Copy
Edit
service:
  healthchecks:
    probe_types:
      - filesystem
      - http
The specified probe_types determine how probe state is exposed. Selecting http will now start an HTTP server even when running in SYNC mode.

**Backwards compatibility**
If no healthchecks configuration is provided, the service retains its previous behavior.

## Implementation

Internally this moves some  configuration from the `service` runner to a new shared module. This should remove some minor   duplication throughout our hosted codebases.

### DIFF
--- a/modules/module-core/src/CoreModule.ts
+++ b/modules/module-core/src/CoreModule.ts
@@ -13,6 +13,8 @@ export class CoreModule extends core.modules.AbstractModule {
   }
 
   public async initialize(context: core.ServiceContextContainer): Promise<void> {
+    this.configureTags(context);
+
     if ([core.system.ServiceContextMode.API, core.system.ServiceContextMode.UNIFIED].includes(context.mode)) {
       // Service should include API routes
       this.registerAPIRoutes(context);
@@ -27,8 +29,11 @@ export class CoreModule extends core.modules.AbstractModule {
     await this.configureMetrics(context);
   }
 
+  protected configureTags(context: core.ServiceContextContainer) {
+    core.utils.setTags(context.configuration.metadata);
+  }
+
   protected registerAPIRoutes(context: core.ServiceContextContainer) {
-    // Register API routes
     context.routerEngine.registerRoutes({
       api_routes: [
         ...core.routes.endpoints.ADMIN_ROUTES,
@@ -41,7 +46,7 @@ export class CoreModule extends core.modules.AbstractModule {
   }
 
   /**
-   * Registers the HTTP server which will handle routes once the router engine is started
+   * Configures the HTTP server which will handle routes once the router engine is started
    */
   protected configureRouterImplementation(context: core.ServiceContextContainer) {
     context.lifeCycleEngine.withLifecycle(context.routerEngine, {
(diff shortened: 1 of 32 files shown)


packages/types/src/config/PowerSyncConfig.ts
```diff
@@ -207,6 +222,16 @@ export const powerSyncConfig = t.object({
     })
     .optional(),
 
+  service: t
+    .object({
+      health_checks: t
+        .object({
+          probe_modes: t.array(t.Enum(ProbeType)).optional()
```
stevensJourney, line 229: After some consideration. I think boolean flags here might be more convenient. This would allow devs to use a base YAML config which conditionally configures the probe modes with environment variables.

```yaml
service:
    health_checks:
       probes:
          http: true
          filesystem: !env PS_ENABLE_FILESYSTEM_PROBES

```

stevensJourney: I've updated this to use booleans

modules/module-core/src/CoreModule.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    if (context.serviceMode == core.system.ServiceContextMode.API) {
+      /**
+       * In the pure API mode we don't currently have any other code which touches the probes.
+       * This configures a timer which will touch every 5 seconds.
+       */
+      let timer: NodeJS.Timeout | null = null;
+      context.lifeCycleEngine.withLifecycle(null, {
+        start: () => {
+          setInterval(() => {
```
rkistner, line 129: Copilot's comment appears correct here - the result of `setInterval` should be assigned to `timer`.

stevensJourney: Good point. Fixed this now :)

rkistner (dismissed):
I like the new CoreModule structure here.

One thing that isn't clear from the docs - if you specify `use_http` and run with `-r sync`, would that start an http server just for the probes?

stevensJourney:
> I like the new CoreModule structure here.
> 
> One thing that isn't clear from the docs - if you specify `use_http` and run with `-r sync`, would that start an http server just for the probes?

Yes, starting in that option will register the probe routes on the `RouterEngine`. The router engine will then start an HTTP server. Only probe routes will be available on this router/server.

rkistner (approved):

stevensJourney:
> One request - can we make this warning more friendly:
> 
> > warn: No routes have been registered. Router Engine will not continue.
> 
> I expect that running with `-r sync` without http probes would be common/expected usage, so shouldn't generate a warning.

I've updated this to an `info` level log which logs `Router Engine will not start an HTTP server as no routes have been registered.`

PR #255, opened 2025-05-06
Fix and test "has_more" for data batches

rkistner:
## Primary issue

With Postgres bucket storage, there are scenarios when a data line would have `has_more: false`, even though the bucket has more data. This leads to a checksum failure for the bucket, and the entire bucket re-downloaded.

This only occurred when there are multiple different buckets returned in a single `getBucketDataBatch()` call, with the first bucket returned not having this issue. This means the client did make progress on each sync attempt, and did not repeatedly run into the same checksum failure. So in most cases, users would not directly see the issue, but sync would take longer than it should.

This fixes the has_more logic to cover this case, with new tests.

## Additional fixed issues

The tests picked up another couple of consistency issues with batch metadata. None of these caused any visible symptoms on the client.
1. MongoDB storage: `has_more` could be true in some cases where it shouldn't be. This resulted in some duplicate work at worst, but no consistency issues with the data sent.
2. Postgres storage: Could return 1001 rows in a batch, where the limit should be 1000.
3. `after` for a batch could be `0`, or the `next_after` value of a different bucket. These are not used on any client, so did not cause actual issues despite the incorrect values.

## Naming things

This updates the variable naming inside the method to make a distinction between the entire "batch", and individual yielded "chunks" inside it. The distinction between length and byte size limits should also be more clear now.

### DIFF
--- a/modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
@@ -288,7 +288,7 @@ export class MongoSyncBucketStorage
     checkpoint: utils.InternalOpId,
     dataBuckets: Map<string, InternalOpId>,
     options?: storage.BucketDataBatchOptions
-  ): AsyncIterable<storage.SyncBucketDataBatch> {
+  ): AsyncIterable<storage.SyncBucketDataChunk> {
     if (dataBuckets.size == 0) {
       return;
     }
@@ -369,16 +369,17 @@ export class MongoSyncBucketStorage
         if (currentChunk != null) {
           // There is an existing chunk we need to yield
           if (currentChunk.bucket == bucket) {
-            // Current and new chunk have the same bucket, so need has_more on the current one
-            // if currentChunk.bucket != bucket, the we reached the end of the previous bucket.
+            // Current and new chunk have the same bucket, so need has_more on the current one.
+            // If currentChunk.bucket != bucket, then we reached the end of the previous bucket,
+            // and has_more = false in that case.
             currentChunk.has_more = true;
             start = currentChunk.next_after;
           }
 
-          const yieldBatch = currentChunk;
+          const yieldChunk = currentChunk;
           currentChunk = null;
           chunkSizeBytes = 0;
-          yield { batch: yieldBatch, targetOp: targetOp };
+          yield { chunkData: yieldChunk, targetOp: targetOp };
           targetOp = null;
         }
 
@@ -419,10 +420,8 @@ export class MongoSyncBucketStorage
       currentChunk = null;
       // This is the final chunk in the batch.
       // There may be more data if and only if the batch we retrieved isn't complete.
-      if (batchHasMore) {
-        yieldChunk.has_more = true;
-      }
-      yield { batch: yieldChunk, targetOp: targetOp };
+      yieldChunk.has_more = batchHasMore;
+      yield { chunkData: yieldChunk, targetOp: targetOp };
       targetOp = null;
     }
   }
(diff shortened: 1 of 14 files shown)


modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
```diff
@@ -343,32 +348,36 @@ export class MongoSyncBucketStorage
...
+        // We need to start a new chunk
         let start: ProtocolOpId | undefined = undefined;
-        if (currentBatch != null) {
-          if (currentBatch.bucket == bucket) {
-            currentBatch.has_more = true;
+        if (currentChunk != null) {
+          // There is an existing chunk we need to yield
+          if (currentChunk.bucket == bucket) {
+            // Current and new chunk have the same bucket, so need has_more on the current one
+            // if currentChunk.bucket != bucket, the we reached the end of the previous bucket.
```
simolus3, line 373: typo, `then`?

modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
```diff
@@ -343,32 +348,36 @@ export class MongoSyncBucketStorage
...
+            currentChunk.has_more = true;
+            start = currentChunk.next_after;
           }
 
-          const yieldBatch = currentBatch;
-          start = currentBatch.after;
-          currentBatch = null;
-          batchSize = 0;
+          const yieldBatch = currentChunk;
```
simolus3, line 378: Should this be called `yieldChunk` now?

modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
```diff
@@ -315,8 +315,13 @@ export class MongoSyncBucketStorage
       });
     }
 
-    const limit = options?.limit ?? storage.DEFAULT_DOCUMENT_BATCH_LIMIT;
-    const sizeLimit = options?.chunkLimitBytes ?? storage.DEFAULT_DOCUMENT_CHUNK_LIMIT_BYTES;
+    // Internal naming:
```
simolus3, line 318: I wonder if it's worth renaming `storage.SyncBucketDataBatch` to `storage.SyncBucketDataChunk` to make this naming consistent everywhere. At least in my understanding, these chunks make up the batch, so yielding multiple things named "batch" from this method sounds like a deviation from this naming scheme.

modules/module-postgres-storage/src/storage/PostgresSyncRulesStorage.ts
```diff
@@ -527,20 +535,25 @@ export class PostgresSyncRulesStorage
...
 
-    if (currentBatch != null) {
-      const yieldBatch = currentBatch;
-      currentBatch = null;
-      yield { batch: yieldBatch, targetOp: targetOp };
+    if (currentChunk != null) {
+      const yieldChunk = currentChunk;
+      currentChunk = null;
+      // This is the final chunk in the batch.
+      // There may be more data if and only if the batch we retrieved isn't complete.
```
simolus3, line 552: nit: Since it's an iff, can we set `yieldChunk.has_more = batchRowCount >= batchRowLimit` directly? I know it's equivalent here, but at least in my mind that makes it clearer that comment and code describe the same thing :)

simolus3 (commented):
The changes look good to me. I also like the clear distinction between batches and chunks now. I'm wondering if we can be more consistent with that naming scheme though.

Also, I think the docs on `SyncBucketData.has_more` could be clearer now. At the moment, it says "True if the response does not contain all the data for this bucket, and another request must be made". But it actually sounds like `has_more` is true if there _could_ be more data, meaning that we can't guarantee that the batch is complete.

simolus3 (approved):
Looks good to me 👍

PR #256, opened 2025-05-06
Allow JSON request payload for RSocket

simolus3:
With the ongoing work to move more parts of the sync logic into the core extension, client SDKs no longer need to implement the BSON deserialization logic themselves (instead relying on our Rust decoder to take care of that). But while that covers response handing, clients are still responsible for crafting RSocket _requests_ (e.g. to transmit tokens or a user agent).

At the moment, our code to handle RSocket requests always assumes that the connection data and metadata will be encoded as BSON. This is unfortunate for clients, because they then need to ship a BSON encoder for the sole purpose of creating these tiny request payloads (and especially on JavaScript, our dependency on the BSON library requires special compilation support for `@powersync/common`). So while we continue to serve BSON only for RSocket requests, we should look for ways to get rid of BSON support on the client SDKs.

The RSocket protocol has space for two mime types for metadata and data fields as part of the `SETUP` frame, and the JavaScript SDK correctly sets those to `application/bson`. To support other clients without a BSON library in the future, this changes the sync service to also accept requests with an `application/json` content type. This allows e.g. the Kotlin SDK to adopt the RSocket protocol.

### DIFF
--- a/.changeset/wild-maps-brake.md
+++ b/.changeset/wild-maps-brake.md
@@ -1,0 +1,7 @@
+---
+'@powersync/service-rsocket-router': patch
+'@powersync/service-errors': patch
+'@powersync/service-core': patch
+---
+
+Allow RSocket request payload to be encoded as JSON
--- a/.changeset/wild-maps-brake.md
+++ b/.changeset/wild-maps-brake.md
@@ -1,7 +1,7 @@
 ---
 '@powersync/service-rsocket-router': patch
 '@powersync/service-errors': patch
-'@powersync/service-core': patch
+'@powersync/service-core': minor
 ---
 
 Allow RSocket request payload to be encoded as JSON
--- a/packages/rsocket-router/src/router/ReactiveSocketRouter.ts
+++ b/packages/rsocket-router/src/router/ReactiveSocketRouter.ts
@@ -75,7 +75,13 @@ export class ReactiveSocketRouter<C> {
             throw new errors.AuthorizationError(ErrorCode.PSYNC_S2101, 'No context meta data provided');
           }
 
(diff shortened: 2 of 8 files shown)


packages/rsocket-router/src/router/ReactiveSocketRouter.ts
```diff
@@ -113,13 +123,17 @@ export class ReactiveSocketRouter<C> {
   }
 }
 
+export interface ReactiveStreamRequest {
```
stevensJourney, line 126: A small nit would be to move this interface to the top of the file.

.changeset/wild-maps-brake.md
```diff
@@ -0,0 +1,7 @@
+---
+'@powersync/service-rsocket-router': patch
+'@powersync/service-errors': patch
+'@powersync/service-core': patch
```
stevensJourney, line 4: I think this should be a `minor` bump for `@powersync/service-core`?

FYI the `@powersync/service-core` and `@powersync/service-image` "package" (used to publish the DockerHub Docker image) are configured for fixed (linked) versioning. I think it would be nice to add an explicit changeset for `@powersync/service-image` mentioning that the new release adds support for JSON websocket request payloads (or something along those lines)

stevensJourney (commented):
The changes here make sense and look good to me.

.changeset/wild-maps-brake.md
simolus3: I've made it a `minor` bump for `service-core` and I've also added a changelog entry for the service image.

stevensJourney (approved):
Looks good to me!

PR #261, opened 2025-05-12
Additional log metadata on sync requests

rkistner:
Two general logging changes:

## 1. Child loggers

Pass a child logger instance through via the Context, so that we can configure metadata on the logger that applies to the entire request. This means that request logs will include common context, making it easier to debug:
1. user_agent
2. rid (unique request id, to correlate with other logs)
3. user_id
4. client_id

Passing the logger through the entire stack could be a pain in some cases. Right now, this is mostly limited to sync requests, so not too bad. If we start using it in many more places, [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage) would be a good way to handle this automatically.

## 2. Granular sync operation tracking

For each completed stream, instead of just logging the total number of synced operations, this logs more details. Example:

```
Sync stream complete {"client_id":"2ea02120-e7d1-4adb-8e9d-b3ddaffd494d","close_reason":client closing stream ","data_synced_bytes":23475179,"large_buckets":{"global[]":51000},"operation_counts":{"clear":0,"move":0,"put":51000,"remove":0},"operations_synced":51000,"rid":"s/0196c4f6-8a5c-7253-a451-ba517f9f0ec7","stream_ms":1331,"user_agent":"powersync-js/1.30.0 powersync-web Chrome/136 linux","user_id":"112c667e-4973-45cb-ade8-40df2ad45a2a"}
```

---

This updates the `uuid` dependency to have access to `uuid.v7()`, for a timestamp-based unique request id. With this version we don't need `@types/uuid` anymore.

* close_reason: Gives an indication (best effort guess) of whether the client or the server closed the stream. Typically the server closes the stream when the auth token expired, or when switching to a new sync rules version. The client closes the stream when the application is closed / tab reloaded, or when there were checksum failures.
* operation_counts: Same as operations_synced, but split into the 4 different types of operations.
* large_buckets: operations_synced for individual buckets. Only counts buckets that synced more than 100 operations in a batch, to avoid too much noise here.
* stream_ms: Similar to duration_ms logged in the generic request handler, but added here to have the details in one place.

### DIFF
--- a/packages/service-core/src/routes/route-register.ts
+++ b/packages/service-core/src/routes/route-register.ts
@@ -1,4 +1,5 @@
 import type fastify from 'fastify';
+import * as uuid from 'uuid';
 
 import { errors, HTTPMethod, logger, router } from '@powersync/lib-services-framework';
 import { Context, ContextProvider, RequestEndpoint, RequestEndpointHandlerPayload } from './router.js';
@@ -25,9 +26,12 @@ export function registerFastifyRoutes(
         handler: async (request, reply) => {
           const startTime = new Date();
           let response: router.RouterResponse;
+          const requestLogger = logger.child({
+            route: e.path,
+            rid: `h/${uuid.v7()}`
+          });
           try {
-            const context = await contextProvider(request);
-
+            const context = await contextProvider(request, { logger: requestLogger });
             let combined = {
               ...(request.params as any),
               ...(request.query as any)
@@ -63,7 +67,7 @@ export function registerFastifyRoutes(
             }
           } catch (ex) {
             const serviceError = errors.asServiceError(ex);
-            logger.error(`Request failed`, serviceError);
+            requestLogger.error(`Request failed`, serviceError);
 
             response = new router.RouterResponse({
               status: serviceError.errorData.status || 500,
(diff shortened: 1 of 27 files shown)


packages/service-core/src/routes/route-register.ts
```diff
@@ -83,13 +87,12 @@ export function registerFastifyRoutes(
           try {
             await reply.send(response.data);
           } finally {
-            await response.afterSend?.();
-            logger.info(`${e.method} ${request.url}`, {
+            await response.afterSend?.({ clientClosed: request.socket.closed });
+            requestLogger.info(`${e.method} ${request.url}`, {
```
stevensJourney, line 91: Should we be concerned with disposing these child loggers which are created per request? It seems like the Logger has a `close` method which we might be able to call in the `afterSend` hook

stevensJourney (approved):
This is an awesome feature which will greatly help with debugging 🙏

packages/service-core/src/routes/route-register.ts
rkistner: I couldn't find anything indicating that the child loggers need to be disposed explicitly. They re-use the same transports etc from the parent logger, only keeping some additional log metadata. They will be automatically garbage collected when not used anymore.

PR #267, opened 2025-05-21
Cleanup sync rules implementation

rkistner:
This contains no functional changes, but does contain some breaking "api" changes for the `@powersync/service-sync-rules` package. This is primarily to clean up the code to make it easier to make larger changes that I'll be working on soon.

Main changes:
1. Make individual query classes "immutable" for the most part. This removes the need for nullable fields while building the instances, making it clear which fields may actually be undefined.
2. Change field names to camelCase for consistency.
3. Remove some unused fields, and document the others.

### DIFF
--- a/packages/sync-rules/src/SqlDataQuery.ts
+++ b/packages/sync-rules/src/SqlDataQuery.ts
@@ -18,13 +18,7 @@ export interface SqlDataQueryOptions extends BaseSqlDataQueryOptions {
 }
 
 export class SqlDataQuery extends BaseSqlDataQuery {
-  static fromSql(
-    descriptorName: string,
-    bucketParameters: string[],
-    sql: string,
-    options: SyncRulesOptions,
-    ruleId?: string
-  ) {
+  static fromSql(descriptorName: string, bucketParameters: string[], sql: string, options: SyncRulesOptions) {
     const parsed = parse(sql, { locationTracking: true });
     const schema = options.schema;
 
@@ -172,7 +166,6 @@ export class SqlDataQuery extends BaseSqlDataQuery {
       descriptorName,
       bucketParameters,
       tools,
-      ruleId: ruleId ?? '',
       errors,
       extractors
     });
@@ -218,8 +211,7 @@ export class SqlDataQuery extends BaseSqlDataQuery {
           bucket: bucketId,
           table: outputTable,
           id: id,
-          data,
-          ruleId: this.ruleId
+          data
         } as EvaluationResult;
       });
     } catch (e) {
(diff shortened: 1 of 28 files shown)


packages/sync-rules/src/TableValuedFunctionSqlParameterQuery.ts
```diff
@@ -25,58 +42,48 @@ import { BucketDescription, BucketPriority, defaultBucketPriority } from './Buck
...
-    const query = new TableValuedFunctionSqlParameterQuery();
+    let errors: SqlRuleError[] = [];
 
-    query.errors.push(...checkUnsupportedFeatures(sql, q));
+    errors.push(...checkUnsupportedFeatures(sql, q));
 
     if (!(call.function.name in TABLE_VALUED_FUNCTIONS)) {
-      query.errors.push(new SqlRuleError(`Table-valued function ${call.function.name} is not defined.`, sql, call));
-      return query;
+      throw new SqlRuleError(`Table-valued function ${call.function.name} is not defined.`, sql, call);
```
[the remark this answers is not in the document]
rkistner: This is indeed a minor change in behavior, but it's fine in this case - the error is properly wrapped further up the stack. The previous behavior had too much potential for issues due to undefined fields.

packages/sync-rules/src/SqlDataQuery.ts
```diff
@@ -13,14 +13,23 @@ import { TableQuerySchema } from './TableQuerySchema.js';
...
+}
 
-  static fromSql(descriptor_name: string, bucket_parameters: string[], sql: string, options: SyncRulesOptions) {
+export class SqlDataQuery extends BaseSqlDataQuery {
+  static fromSql(
+    descriptorName: string,
+    bucketParameters: string[],
+    sql: string,
+    options: SyncRulesOptions,
+    ruleId?: string
```
simolus3, line 26: I'm not sure I fully understand the purpose of `ruleId` from a quick look, but it looks like this value is always set (and only ignored for tests). So I wonder if we should make it required and then set it to a bogus value in tests directly calling `SqlDataQuery.fromSql`.

simolus3 (commented):
I agree that making the queries immutable is a nicer setup, so I'm happy with these changes (apart from a minor nit) 👍

packages/sync-rules/src/SqlDataQuery.ts
rkistner: Good point. I looked at this again and realized for data queries it was only used in 2 tests, and nowhere outside tests. I removed it rather.

Issue #238, opened 2025-04-02
MySQL Replication Job Not Using Correct MySQL Port


joshuagauci:
There is a bug in BinLogReplicationJob.ts where the poolOptions are not using the same port as the original MySQL connection. This causes connection failures when trying to replicate MySQL data.

Steps to Reproduce:

Configure PowerSync to connect to a MySQL database on port 25060.

Run the replication job.

Observe that while the main MySQL connection is correctly using port 25060, the pool connections are defaulting to 3306.

The connection fails due to the incorrect port.

**Expected Behavior:**
The connection pool should inherit the port from the original MySQL connection configuration, ensuring that all connections use 25060 instead of the default 3306.

**Actual Behavior:**

The main connection is correctly using port 25060.

However, poolOptions default to 3306, causing connection timeouts and replication failures.

**Tested Workaround:**
I manually set the port in the connection factory configuration as follows:

const connectionManager = this.connectionFactory.create({
  idleTimeout: 30_000,
  port: 25060 // Manually setting the correct port
});

With this change, PowerSync was able to connect, but replication did not complete due to another connection timeout, which also defaulted to port 3306.

Logs:

`Connection Manager {"binlogListeners":[],"isClosed":false,"name":"ReplicationJob: mysql-demo-1","options":{"database":"demo","hostname":"test.url.com","id":"default","password":"testpassword","port":25060,"server_id":1,"tag":"default","type":"mysql","uri":"mysql://test:testpassword@test.url.com:25060/demo","username":"test"},"pool":{"config":{"connectionConfig":{"port":3306}}}}
`
PowerSync Connection Configuration:

```
replication:
  connections:
    - type: mysql
      uri: mysql://test:testpassword@test.url.com:25060/demo
```

or

```
replication:
  connections:
    - type: mysql
      uri: mysql://test:testpassword@test.url.com/demo:25060
      port: 25060
```

**Temporary Workaround:**

Setup a stream relay using socat to proxy port 3306 to 25060.

michaelbarnes:
Hi @joshuagauci thanks for reporting this and the detail. We'll work on reproducing this on our end and let the team know,

PR #269, opened 2025-05-21
[MySQL] BinLog Processing Queue backpressure handling

Rentacookie:
Sometimes when a large number of MySQL binlog events were received in a short timeframe, the bucket storage could not persist the events fast enough, causing a backlog in the processing queue. This could lead to out of memory errors because the queue size itself was unbounded. 

We now monitor the queue memory footprint, and if it is found to have reached the configured maximum, we pause the binlog listener until all the events currently in the queue have been processed, or after a certain amount of time has passed. Streaming is then immediately resumed.
This effectively limits the number of in flight events that are in memory at any given point when replicating.

To accomplish the above a big chunk of the binlog listening logic was moved to its own class. This cleans up the BinLogStream a bit and made for easier testing.

Additionally, the interim types declared for the Zongji listener were moved to the Zongji repository in this pull request:
https://github.com/powersync-ja/powersync-mysql-zongji/pull/3

The binlog listener config has been updated to include the port, which solves the issue reported here:
https://github.com/powersync-ja/powersync-service/issues/238


modules/module-mysql/src/replication/zongji/BinLogListener.ts
```diff
@@ -0,0 +1,206 @@
...
+        }
+      });
+
+      this.zongji.on('stopped', () => {
+        resolve();
+        logger.info('BinLog listener stopped. Replication ended.');
+      });
+
+      // Handle the edge case where the listener has already been stopped before completing startup
+      if (this.isStopped) {
```
stevensJourney, line 91: Just some thoughts/questions from my side on the logic here.

Should we remove the listeners which have been added here with `this.zongji.on` when we complete this promise? Or are they configured to only run once?

I also assume the logic in this promise creator callback is synchronous, which makes me wonder if we should be checking `this.isStopped` before registering the event handlers? Is it possible that we could trigger a stop event while creating this promise?

rkistner (dismissed):

PR #270, opened 2025-05-22
Clarify oplog entry format

simolus3:
The `data` field in `OplogEntry` will always be a nullable string, we don't store contents as an object directly.

This was a bit confusing to me when writing the new sync client for the core extension, I thought the `Record` representation was something we had to deal with. This fixes the type annotations to make it clear that that's not true.


rkistner (approved):
Thanks - this does help clarify it

PR #272, opened 2025-05-28
Add powersync_replication_lag_seconds metric

rkistner:
Record replication lag in the logs and as a new metric, to help diagnose and alert on replication delays.

New metric:

```
# HELP powersync_replication_lag_seconds Replication lag between the source database and PowerSync instance
# TYPE powersync_replication_lag_seconds gauge
powersync_replication_lag_seconds 15
```

New logs on commit:

```
info: powersync_15 Flushed 2 + 0 + 2 updates, 1kb in 5ms. Last op_id: 99. Replication lag: 2s {"flushed":{"bucket_data_count":2,"current_data_count":2,"duration":5,"parameter_data_count":0,"replication_lag_seconds":2,"size":1002}}
```

Note that there is some difference between the logs versus the metric:
1. The logs contain the delay for the transaction(s) being committed - the time between commit in the source db, and creating a new checkpoint in the powersync instance.
2. The metric is a sample of the current delay - the time elapsed since the oldest pending transaction has been committed to the source db. The metric is always 0 when there are no pending changes.

We generally use the difference between the source database timestamps and the powersync instance time to calculate the lag. If the time for one of them is out, that will cause a constant offset to the replication lag calculation. This also means the replication lag could be reported as negative.

If the active sync rules is in an error state, we report the time since it could last persist a change.
If there are no active sync rules (e.g. sync rules was deployed for the first time), the metric is not reported.

---

The initial implementation here used a normal gauge instead of an observable one (i.e. the value was set at specific points, rather than polled). I extended the metrics implementation to add a plain `Gauge` interface. Since then I've found that the observable gauge approach is just better, and switched to that, removing the Gauge interface again. Replication lag is better calculated when you measure it, rather than at specific events, and I think the same applies to most gauges.


modules/module-mongodb-storage/src/storage/implementation/PersistedBatch.ts
```diff
@@ -295,19 +299,51 @@ export class PersistedBatch {
...
-    const duration = performance.now() - startAt;
-    logger.info(
-      `powersync_${this.group_id} Flushed ${this.bucketData.length} + ${this.bucketParameters.length} + ${
-        this.currentData.length
-      } updates, ${Math.round(this.currentSize / 1024)}kb in ${duration.toFixed(0)}ms. Last op_id: ${this.debugLastOpId}`
-    );
+    if (flushedSomething) {
+      const duration = Math.round(performance.now() - startAt);
+      if (options?.oldestUncommittedChange != null) {
+        const replicationLag = Math.round((Date.now() - options.oldestUncommittedChange.getTime()) / 1000);
```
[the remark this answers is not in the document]
rkistner: Ignoring this nitpick.

modules/module-mysql/src/replication/BinLogStream.ts
```diff
@@ -560,6 +583,19 @@ AND table_type = 'BASE TABLE';`,
         return null;
     }
   }
+
+  async getReplicationLagMillis(): Promise<number | undefined> {
+    if (this.oldestUncommittedChange == null) {
+      if (this.isStartingReplication) {
```
Rentacookie, line 589: More of a nitpick, but looking at how this flag is used, a clearer name than `isStartingReplication` might be `isReplicating` with the boolean logic flipped. ie
```
if (this.oldestUncommittedChange == null) {
      if (this.isReplicating) {
        // We don't have any uncommitted changes, so replication is up-to-date.
        return 0;
      } else {
        // We don't have anything to compute replication lag with yet.
        return undefined;
      }
    }
    return Date.now() - this.oldestUncommittedChange.getTime();
}
```

rkistner: I went with `isStartingReplication` to highlight that it's specifically to keep track of the "startup" period, and not tracking stopping replication for example.

PR #287, opened 2025-06-25
[MySQL] Automatic schema change handling

Rentacookie:
Currently, when syncing from a MySQL database database schema changes are ignored. This is problematic when those changes affect tables that are in the defined sync rules. The work around for that has been to redeploy the sync rules, thereby triggering a full re-sync.

This change set adds functionality to automatically handle schema changes affecting tables in the sync rules by listening for and parsing DDL query binlog events.

The basic mechanism for this works as follows:
1.  Listen for query binlog events.
2. Parse these events to see if they affect the schema of tables in the sync rules.
3. Pause the binlog listener and update the stored bucket data based on the schema change if required
4. Restart the binlog listener.

Parsing of the query SQL is accomplished by using the [node-sql-parser](https://github.com/taozhi8833998/node-sql-parser) package. This does introduce a limitation in that if the library can't parse the query, we cannot easily interpret the intended schema change either. If this is the case, some best effort pattern matching is done to check if the query is a schema change that affects one of the replicated tables and a warning is logged.

The following schema changes are detectable:
- Create, rename, drop Table
- Truncate Table
- Create / Drop unique indexes and primary keys
- Add, modify, drop, rename columns

With the exception of Create Table, all of the schema events are detected by parsing the DDL statements received in the binlog query events.  For Create table, these changes are picked up when row events are received for the new table.


packages/service-core/src/storage/SourceTable.ts
```diff
@@ -1,14 +1,24 @@
 import { DEFAULT_TAG } from '@powersync/service-sync-rules';
 import * as util from '../util/util-index.js';
-import { ColumnDescriptor } from './SourceEntity.js';
+import { ColumnDescriptor, SourceEntityDescriptor } from './SourceEntity.js';
+
+export interface SourceTableOptions {
```
Rentacookie, line 5: This is more of code style change than anything else, we tend to rather create these encapsulating interfaces instead of having functions/constructors with a lot of parameters

packages/service-core/src/storage/SourceEntity.ts
```diff
@@ -10,17 +10,17 @@ export interface ColumnDescriptor {
...
    * If specified, this is specifically used to detect renames.
    */
   objectId: number | string | undefined;
   schema: string;
   name: string;
-  replicationColumns: ColumnDescriptor[];
+  /**
+   *  The columns that are used to uniquely identify a record in the source entity.
+   */
+  replicaIdColumns: ColumnDescriptor[];
```
Rentacookie, line 25: The original name here was not quite right and could be confusing.  Reverted to the name primarily used throughout the code base

packages/service-core/src/storage/SourceTable.ts
```diff
@@ -1,14 +1,24 @@
...
+}
 
 export interface TableSnapshotStatus {
   totalEstimatedCount: number;
   replicatedCount: number;
   lastKey: Uint8Array | null;
 }
 
-export class SourceTable {
+export class SourceTable implements SourceEntityDescriptor {
```
Rentacookie, line 21: This paves the way for some potential future cleanup, but also helps consolidating the naming of fields.

packages/service-core/src/metrics/open-telemetry/util.ts
```diff
@@ -61,7 +63,8 @@ export function createOpenTelemetryMetricsFactory(context: ServiceContext): Metr
   const meterProvider = new MeterProvider({
     resource: new Resource(
       {
-        ['service']: 'PowerSync'
+        ['service']: 'PowerSync',
+        ['service.version']: pkg.version
```
Rentacookie, line 67: This is a rather useful tag to include with our metrics.

modules/module-mysql/src/common/mysql-to-sqlite.ts
```diff
@@ -183,6 +183,9 @@ export function toSQLiteRow(row: Record<string, any>, columns: Map<string, Colum
           result[key] = row[key];
           break;
       }
+    } else {
+      // If the value is null, we just set it to null
+      result[key] = null;
```
Rentacookie, line 188: @rkistner I am not 100% sure if this is the correct place to fix it, but I found that without this, columns with null values would not be added to sqliterow and added to the storage.
There is a schema change test that adds a column that exposed that this might be a problem.

modules/module-mysql/src/replication/zongji/BinLogListener.ts
```diff
@@ -1,29 +1,66 @@
...
+const MAX_PAUSE_TIME_MS = 45_000;
 
 export type Row = Record<string, any>;
 
+export enum SchemaChangeType {
+  CREATE_TABLE = 'create_table',
+  RENAME_TABLE = 'rename_table',
+  DROP_TABLE = 'drop_table',
+  TRUNCATE_TABLE = 'truncate_table',
+  MODIFY_COLUMN = 'modify_column',
```
Rentacookie: Technically we can combine most of these column changes into one type, but keeping them seperate opens the door to some potential optimisations for handling the schema change.

modules/module-mysql/src/replication/zongji/BinLogListener.ts
```diff
@@ -67,126 +107,164 @@ export class BinLogListener {
...
+          zongji.resume();
+        }
       }
     });
 
-    zongji.on('ready', async () => {
-      // Set a heartbeat interval for the Zongji replication connection
-      // Zongji does not explicitly handle the heartbeat events - they are categorized as event:unknown
-      // The heartbeat events are enough to keep the connection alive for setTimeout to work on the socket.
-      await new Promise((resolve, reject) => {
```
Rentacookie, line 160: Whilst testing I discovered that trying to set up the heartbeat once the listener is already active does not work. The listener itself already has a lock on the connection then.

modules/module-mysql/src/replication/zongji/BinLogListener.ts
```diff
@@ -67,126 +107,164 @@ export class BinLogListener {
...
+        });
+        this.zongji.stop();
+        this.processingQueue.kill();
+      });
     }
   }
 
-  private get isStopped(): boolean {
-    return this.zongji.stopped;
+  public async replicateUntilStopped(): Promise<void> {
```
Rentacookie, line 210: This was required after decoupling the zongji stop/start from this binlog listener wrapper stop start. 

Internally the binlog listener can stop and start Zongji when required, not affecting the externally viewed overal stop/start state of this wrapper.

modules/module-mysql/src/replication/zongji/BinLogListener.ts
```diff
@@ -201,43 +279,192 @@ export class BinLogListener {
...
+  }
+
+  /**
+   *  Function that interprets a DDL query for any applicable schema changes.
+   *  If the query does not contain any relevant schema changes, an empty array is returned.
+   *
+   *  @param query
+   */
+  private toSchemaChanges(query: string): SchemaChange[] {
+    const ast = this.sqlParser.astify(query, { database: 'MySQL' });
```
rkistner: How closely does this parser follow the MySQL syntax? And what happens if it runs into syntax it doesn't support?

One example of a past issue:
https://github.com/taozhi8833998/node-sql-parser/issues/2232

To what extent would it be feasible to use the tablemap events instead of or in addition to this to detect changes?

modules/module-mysql/src/replication/zongji/BinLogListener.ts
```diff
@@ -67,126 +107,164 @@ export class BinLogListener {
...
-        );
-        zongji.pause();
-        const resumeTimeoutPromise = new Promise((resolve) => {
-          setTimeout(() => resolve('timeout'), MAX_QUEUE_PAUSE_TIME_MS);
-        });
+      this.logger.debug(`Received BinLog event:${evt.getEventName()}`);
+      // We have to handle schema change events before handling more binlog events,
+      // This avoids a bunch of possible race conditions
+      if (zongji_utils.eventIsQuery(evt)) {
+        await this.processQueryEvent(evt);
```
rkistner: What specific race conditions are covered here?

It feels like this has the possibility to introduce more race conditions, since the order is not preserved between schema changes (applied immediately here) and data modifications (queued).

Furthermore, zongi doesn't wait for these promises to complete AFAIK, so processing here can cause concurrent processing of events if multiple events come in shortly after each other, and uncaught rejections if the processing fails.

rkistner: Ok, reading more code I see it works like this:
1. Schema change detected (synchronously up to that point).
2. Stop the listener.
3. Wait until queue is empty.
4. Process the schema change.

So that would avoid the race conditions. I'm just wondering if it could be simpler to do this as part of the normal queue mechanism?

modules/module-mysql/src/replication/zongji/BinLogListener.ts
rkistner: Here's an example that crashes the process with a nasty error (thanks to LLMs for generating some nasty test cases):

```sql
CREATE TABLE sales (
    sale_date DATE,
    amount DECIMAL(10,2)
) PARTITION BY RANGE (YEAR(sale_date)) (
    PARTITION p2023 VALUES LESS THAN (2024),
    PARTITION p2024 VALUES LESS THAN (2025),
    PARTITION pmax VALUES LESS THAN MAXVALUE
);
```

The crash happens regardless of whether the table is referenced in sync rules or not, and you cannot recover by restarting - it will immediately crash at the same point again.

rkistner: Another example:

```sql
CREATE TABLE invisible_column_example (
  id INT,
  created_at DATE INVISIBLE
)
```

The point is not these specific examples though, just that it would be extremely difficult to cover all cases.

modules/module-mysql/src/replication/zongji/BinLogListener.ts
Rentacookie: I did try the normal queuing mechanism first, but the problem I ran into was that after handling the schema change we update the current binlogposition so that when the listener is restarted we process event from that point onwards. But there might already be a bunch of events in the queue, so if the listener is restarted with the binlog position post schema change. A bunch of duplicate events end up on the processing queue.

So that could be handled by processing all the events in the queue after the schema change, causing the binglog position to update and only then restarting the listener. But since the processing queue only has a single worker you run into a deadlock. As in you can't process any new tasks until the current schema change one has been completely finished and removed from the queue. 

The schema change task can't technically be truly completed until the listener is restarted, but maybe I can work around that with a callback 🤔 I'll check it out since I much preferred the processing queue only approach.

modules/module-mysql/src/replication/zongji/BinLogListener.ts
Rentacookie: I see what you mean. 😧 

Unfortunately the tablemap events are really only useable for detecting some column modifications. Dropping, truncating and renaming of tables do not result in any tablemap events. I did initially try a joint approach with tablemap + query parsing. But in the end since you have to parse all DDL queries anyway it felt simpler to settle on a single source for schema changes.

So that left the query parsing as the only real source of those schema changes. But as you say, if we can't parse the DDL queries our hands are rather tied. In those cases we probably have to end up skipping that specific schema change 😬 Maybe we list  that as a known issue and add a warning to the logs when it happens? 
That hopefully doesn't happen too often though, considering that node-sql-parser is a rather prolificly used and maintained. 🤞 

I am open to any other approaches though.

modules/module-mysql/src/replication/zongji/BinLogListener.ts
Rentacookie: Good news I was able to slot the schema handling back into the processing queue, by using a callback to restart the listener once the queue had emptied.

modules/module-mysql/src/replication/zongji/BinLogListener.ts
```diff
@@ -1,29 +1,75 @@
...
   onCommit: (lsn: string) => Promise<void>;
+  onSchemaChange: (change: SchemaChange) => Promise<void>;
 }
 
 export interface BinLogListenerOptions {
   connectionManager: MySQLConnectionManager;
   eventHandler: BinLogEventHandler;
-  includedTables: string[];
+  // Filter for tables to include in the replication
+  tableFilter: (tableName: string) => boolean;
```
Rentacookie: Adding this functionality fixed the issue where wild card tables could not be replicated

modules/module-mysql/src/replication/zongji/BinLogListener.ts
```diff
@@ -67,126 +118,156 @@ export class BinLogListener {
...
         zongji.pause();
-        const resumeTimeoutPromise = new Promise((resolve) => {
-          setTimeout(() => resolve('timeout'), MAX_QUEUE_PAUSE_TIME_MS);
-        });
-
-        await Promise.race([this.processingQueue.empty(), resumeTimeoutPromise]);
-
-        this.logger.info(`Binlog processing queue backlog cleared. Resuming Binlog listener.`);
+        const resumeTimeoutPromise = timers.setTimeout(MAX_PAUSE_TIME_MS);
+        await Promise.race([this.processingQueue.drain(), resumeTimeoutPromise]);
```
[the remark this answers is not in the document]
Rentacookie: Falso positive. Calling `queue.drain()` with no arguments indeed returns a promise.

rkistner (approved):

rkistner (approved):

PR #296, opened 2025-07-09
[Hotfix] 1.12.x [DO NOT MERGE]

rkistner:
Backports two logging fixes on the 1.12.x release:
1. #288
2. #295

### DIFF
--- a/modules/module-core/package.json
+++ b/modules/module-core/package.json
@@ -2,7 +2,7 @@
   "name": "@powersync/service-module-core",
   "repository": "https://github.com/powersync-ja/powersync-service",
   "types": "dist/index.d.ts",
-  "version": "0.1.1",
+  "version": "0.1.2-hotfix.0",
   "main": "dist/index.js",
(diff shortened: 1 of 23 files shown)


modules/module-core/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@powersync/service-module-core",
   "repository": "https://github.com/powersync-ja/powersync-service",
   "types": "dist/index.d.ts",
-  "version": "0.1.1",
+  "version": "0.1.2",
```
stevensJourney: This version already exists and is published to [NPMJS](https://www.npmjs.com/package/@powersync/service-module-core/v/0.1.2). ~Deploying these packages would most likely result in an error~ I believe we skip already published packages.

stevensJourney: But since there are no changes in this module, I guess we will just skip publishing and use the published version

stevensJourney (dismissed):

PR #300, opened 2025-07-16
Specify maxTimeMs for MongoDB storage queries

rkistner:
Specify a limit of 40s for each query, instead of relying on the connection socket timeout. The most common cause is a timeout when computing checksums on buckets with millions of operations. We can often recover by retrying the query - the second attempt is faster since the data would be cached in database memory.

Before:
```
[PSYNC_2001] Something went wrong
  connection 10 to a.b.c.d:27017 timed out
```

Now:
```
[PSYNC_S2403] Query timed out while reading checksums
```


stevensJourney (approved):
Looks good to me :)

PR #307, opened 2025-07-21
Sync streams

simolus3:
This is  a work-in-progress PR for sync streams.

Sync streams essentially consist of two features:

1. A new way to define sync rules in a way that only requires a single definition instead of parameter and data queries (the internal "Subqueries / New Sync Streams Syntax" document).
2. The ability for clients to explicitly subscribe to sync streams (the internal "Sync Streams" document).

This PR implements a small subset of feature 1 by desugaring the unified query into the parameter / data queries we have today. That seemed like the easiest way to get started on this, but I think it's not a good approach to handle subqueries (which are not implemented in this PR). I will open a follow-up PR to implement the new syntax in a clean structure able to handle that better.

Instead, most of this PR is related to protocol changes enabling the second feature. While most of the subscription management happens on the client, there are two fundamental changes:

1. Clients need to know why they're receiving a bucket: We want clients to be able to track progress (and also provide the usual `hasSynced` / `lastSyncedAt` fields) for individual stream subscriptions. Since progress is ultimately attached to buckets, clients need enough information to associate buckets with stream subscriptions.
2. Bucket queriers can yield the same bucket multiple times: Since clients can subscribe to the same stream with different parameters, it's possible to subscribe in a way that causes streams to have overlapping results. Consider for instance a stream defined as `SELECT * FROM assets WHERE id IN (request.args() -> 'asset_ids')` and a client subscribing once with `{'asset_id': [1]}` and another time with `{'asset_id': [1, 2]}`. Here, both subscriptions yield the bucket `assets[id=1]`. We obviously don't want the bucket to be included in the checkpoint multiple times, but clients need to know that progress on bucket `assets[id=1]` contributes to both stream subscriptions.

Most of the logic related to the first requirement is implemented in the `sync-rules` package:

1. In `SqlSyncRules.getBucketParameterQuerier`, we take an additional array of explicit stream subscriptions. When we find a matching stream, we evaluate parameter queries for each of those subscriptions. Otherwise, we evaluate the default subscription if the stream has one.
2. Each stream subscription has an opaque id, which is internal to a sync stream (so, as far as the service is concerned, stateless), and generated by the client. When a bucket is found as part of a querier, we either:
 - remember that the bucket was generated by a legacy sync rule or a default stream
 - or otherwise, include the opaque id of the stream that resulted in that bucket.

In `service-core`, we:

1. Forward stream subscriptions from the client to `sync-rules`.
2. Merge multiple buckets created from different stream subscriptions into one if they're duplicate.
  - TODO: Check if it makes sense to do this in `sync-rules` directly - it's a bit easier to do it when building checksums because we need access to all buckets (static and dynamic) to do this.
3. Apply subscription options, like a custom bucket priority for buckets within a subscription.

This is still missing tests, but I wanted to get it out for an initial discussion.

### DIFF
--- a/packages/service-core/src/util/protocol-types.ts
+++ b/packages/service-core/src/util/protocol-types.ts
@@ -16,11 +16,17 @@ export type BucketRequest = t.Decoded<typeof BucketRequest>;
 /**
  * A sync steam that a client has expressed interest in by explicitly opening it on the client side.
  */
-export const OpenedStream = t.object({
+export const RequestedStreamSubscription = t.object({
   /**
    * The defined name of the stream as it appears in sync stream definitions.
    */
   stream: t.string,
+  /**
+   * An opaque textual identifier assigned to this request by the client.
+   *
+   * Wh
+   */
+  client_id: t.string,
   /**
    * An optional dictionary of parameters to pass to this specific stream.
    */
@@ -34,26 +40,26 @@ export const OpenedStream = t.object({
   override_priority: t.number.optional()
 });
 
-export type OpenedStream = t.Decoded<typeof OpenedStream>;
+export type RequestedStreamSubscription = t.Decoded<typeof RequestedStreamSubscription>;
 
 /**
- * An overview of all opened streams as part of a streaming sync request.
+ * An overview of all subscribed streams as part of a streaming sync request.
  */
-export const OpenedStreams = t.object({
+export const StreamSubscriptionRequest = t.object({
   /**
    * Whether to sync default streams.
    *
-   * When disabled,only
+   * When disabled, only explicitly-opened subscriptions are included.
    */
   include_defaults: t.boolean.optional(),
 
   /**
    * An array of sync streams the client has opened explicitly.
    */
-  opened: t.array(OpenedStream)
+  opened: t.array(RequestedStreamSubscription)
 });
 
-export type StreamSubscriptions = t.Decoded<typeof OpenedStreams>;
+export type StreamSubscriptionRequest = t.Decoded<typeof StreamSubscriptionRequest>;
 
 export const StreamingSyncRequest = t.object({
   /**
@@ -94,7 +100,7 @@ export const StreamingSyncRequest = t.object({
   /**
    * If the client is aware of streams, an array of streams the client has opened.
    */
-  subscriptions: OpenedStreams.optional()
+  subscriptions: StreamSubscriptionRequest.optional()
 });
 
 export type StreamingSyncRequest = t.Decoded<typeof StreamingSyncRequest>;
@@ -107,7 +113,7 @@ export interface StreamingSyncCheckpointDiff {
   checkpoint_diff: {
     last_op_id: ProtocolOpId;
     write_checkpoint?: ProtocolOpId;
-    updated_buckets: BucketChecksumWithDescription[];
+    updated_buckets: CheckpointBucket[];
     removed_buckets: string[];
   };
 }
@@ -154,7 +160,7 @@ export interface StreamDescription {
 export interface Checkpoint {
   last_op_id: ProtocolOpId;
   write_checkpoint?: ProtocolOpId;
-  buckets: BucketChecksumWithDescription[];
+  buckets: CheckpointBucket[];
   streams: StreamDescription[];
 }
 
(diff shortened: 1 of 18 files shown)


packages/service-core/src/util/protocol-types.ts
```diff
@@ -13,9 +13,55 @@ export const BucketRequest = t.object({
...
+ */
+export const RequestedStreamSubscription = t.object({
+  /**
+   * The defined name of the stream as it appears in sync stream definitions.
+   */
+  stream: t.string,
+  /**
+   * An opaque textual identifier assigned to this request by the client.
+   */
+  client_id: t.string,
```
rkistner: Is there perhaps a different name we can use here? `client_id` can be confused with the `StreamingSyncRequest.client_id` (already have that issue with write checkpoints).

packages/service-core/src/util/protocol-types.ts
```diff
@@ -99,10 +150,16 @@ export type StreamingSyncLine =
  */
 export type ProtocolOpId = string;
 
+export interface StreamDescription {
+  name: string;
```
rkistner, line 153: How is this name derived? For example the request stream or client_id?

packages/service-core/src/util/protocol-types.ts
```diff
@@ -158,4 +215,43 @@ export interface BucketChecksum {
...
+export type BucketSubscriptionReason = BucketDerivedFromDefaultStream | BucketDerivedFromExplicitSubscription;
+
+/**
+ * A bucket has been included in a checkpoint because it's part of a default stream.
+ */
+export type BucketDerivedFromDefaultStream = {
+  /**
+   * The index (into {@link Checkpoint.streams}) of the stream defining the bucket.
+   */
+  default: number;
```
rkistner, line 267: It feels a little strange that we use a numeric index here for default streams, but the stream "client_id" for requested streams. Could we use the same thing for both for consistency - either the numeric index or the name? (In the case of name we'd probably have to reserve "default" or something for default streams).

packages/service-core/src/util/protocol-types.ts
simolus3: It's the name of the stream as it appears in the sync rules - I'll add a comment mentioning this.

Adding this information has two purposes:

1. It allows clients to list all subscriptions (normally they could just remeber what they've subscribed to, but there are also default streams to keep in mind).
2. It allows clients to determine which subscriptions were successful (i.e., actually exist on the server). This backs the `SyncSubscriptionDefinition.active` field in the proposed API.

packages/service-core/src/util/protocol-types.ts
simolus3: I've replaced client ids with their index in the request.

packages/service-core/src/util/protocol-types.ts
simolus3: I've removed that field - we can use the index of the request in this array as an implicit id.

packages/service-core/src/util/protocol-types.ts
rkistner: What happens here if the client had multiple stream subscriptions with the name stream name but different parameters?

simolus3: There would only be a single entry here in that case. That is enough because clients only need to know that the stream exists, so if they see a `StreamDescription` with the same name, they know all their subscriptions on that stream must be active (as it's impossible for a stream definition to "reject" certain parameters).

There is a separate, more granular mapping that describes why a bucket is part of a checkpoint, that indeed references a (stream, parameter) key. That's necessary because clients need the bucket -> stream subscription association to track progress for a particular stream + parameter pair. But `checkpoint.streams` just says "yup, your two subscriptions to stream `a` exist".

We need this separate mapping because it's possible to request a stream subscription that yields no bucket (e.g. when there's an additional JWT-filter in the select statement). This is not the same as the stream not existing, we would report `hasSynced` at the first completed checkpoint without downloading any data in that case.

rkistner: If the primary use case is knowing if the stream doesn't exist, I wonder if we should rather implement a more general error detection system for streams. "Stream doesn't exist" may be the most common "error", but there are other possibilities. While the current syntax doesn't have explicit parameter validation support, you could still get errors. Taking this example stream query:

```sql
select * from issues where id in stream.parameters -> 'issue_ids'.
```

If the client passes in anything other than an array into issue_ids, that query will error.

Right now, that would block the entire sync request. But I wonder if we should do those type of errors on individual stream subscriptions instead? The client can then decide whether errors on specific streams should should block the checkpoint or not, and handle missing streams and invalid parameters in the same way.

Then we'd still need something for default streams, but perhaps we can just have that as a separate `default_streams` list on the Checkpoint? The client would need separate handling for default streams vs other streams either way.

simolus3: I love that idea! I've added an `errors` array to this structure to allow identifying errors for each stream. Each error contains the subscription index that caused it, or the string `"default"`. E.g. it looks [like this](https://github.com/powersync-ja/powersync-service/blob/8a0273ac99fdacd90a9e64c48206f90c5cf72bdd/packages/service-core/test/src/sync/BucketChecksumState.test.ts#L838-L881).

> The client can then decide whether errors on specific streams should should block the checkpoint or not, and handle missing streams and invalid parameters in the same way

I think that's a reasonable approach for streams. For sync rules, I think we should keep the existing behavior. It's hard to argue that we need to keep fatal errors for backwards compatibility, but it would still be odd to go from global error to partial/potentially inconsistent data.

rkistner: Agreed on keeping existing behavior in sync-rules for backwards-compatibility.
For streams, this just moves the responsibility to the client - the client can choose to treat it as a fatal error (to make sure the client only has consistent data), or ignore it, potentially configurable per stream subscription. We don't actually need to implement all those options on the client in the first version, but it would be nice to have that option in the future.

PR #314, opened 2025-07-30
[MySQL]  Extend replication keep alive mechanism.

Rentacookie:
When the source MySQL database is idle for long periods of time, write checkpoints in powersync can be sparse.

We use [hearbeat](https://dev.mysql.com/doc/dev/mysql-server/9.3.0/page_protocol_replication_binlog_event.html#sect_protocol_replication_event_heartbeat) events on the BinLog to keep the replication connection alive. But now we also latch onto these events to update the keep alive timestamp on the sync rules.

### DIFF
--- a/.changeset/silent-suns-watch.md
+++ b/.changeset/silent-suns-watch.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-image': patch
+---
+
+Add attestations to Docker image.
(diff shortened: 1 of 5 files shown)


packages/service-core/src/api/diagnostics.ts
```diff
@@ -134,7 +134,7 @@ export async function getSyncRulesStatus(
     })
   );
 
-  if (live_status && status?.active && sourceConfig.type != 'mysql') {
+  if (live_status && status?.active) {
```
Rentacookie, line 137: @rkistner I think this was the only thing to re-enable the replication lag check for mysql?

rkistner (approved):
Looks great!

PR #315, opened 2025-07-31
[MongoDB Storage] Compact parameter lookups

rkistner:
The compact action now also compacts bucket parameter storage.

I many use cases, parameter lookups are not modified often, and take up fairly little storage. However, there are some cases where it _is_ modified often, and this causes two issues:
1. Storage increases indefinitely.
2. Parameter queries gets slower.

This compacts the parameter storage when using MongoDB storage.

See the new `docs/parameters-lookups.md` for details on the implementation.

One notable change is that we now do "snapshot" queries for parameter data, matching the snapshot state of the checkpoint query. This is required for consistency of parameter lookups during a compact operation. This required some refactoring to the storage APIs.

TODO:
 * [ ] Add tests specific to removing deleted parameter lookup entries.

### DIFF
--- a/.changeset/angry-mice-hide.md
+++ b/.changeset/angry-mice-hide.md
@@ -1,0 +1,9 @@
+---
+'@powersync/service-module-postgres-storage': minor
+'@powersync/service-module-mongodb-storage': minor
+'@powersync/service-core-tests': minor
+'@powersync/service-core': minor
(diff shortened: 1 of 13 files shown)


stevensJourney (approved):
This is a very cool use-case for MongoDB snapshot queries. The changes here look good to me, I could not spot any potential issues :D

PR #316, opened 2025-08-04
Feature/sdk reporting

EspressoTrip-v2:
Connection SDK reporting.
Added changes work in conjunction with the Report Module in powersync management. The purpose of the changes is to store sdk information that can be used in a dashboard or potentially for us to get detailed information on a instance level ( not implemented yet ). It is Mongo and Postgres compatible. 

Additions:

1. Event emitter engine, can define events that can be used anywhere. 
2. Posgres report storage factory
3. Mongo report storage factory

Basic flow:

- User connects, user information, jwt and meta are stored
- User connection refreshes or reconnects, if the connection event is within the same day the stored initial connection is updated. Else new connection is stored
- User disconnects connection is updated as disconnected, using the previously connected time for long running connections
Scraping:

- powersync-usage will scrape projection based on specified range ( day, week, month )
- powersync usage can also specify a ttl range for old data to be removed on the scrape, default is a month


modules/module-mongodb-storage/src/migrations/db/migrations/1752661449910-sdk-reporting.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const down: migrations.PowerSyncMigrationFunction = async (context) => {
+  const {
+    service_context: { configuration }
+  } = context;
+
+  const db = storage.createPowerSyncMongo(configuration.storage as MongoStorageConfig);
+
+  try {
+    if (await db.write_checkpoints.indexExists('connect_at')) {
```
Rentacookie: These still need to be changed to the `sdk_report_events` collection.

packages/service-core/src/storage/ReportStorageFactory.ts
```diff
@@ -0,0 +1,9 @@
+import { event_types } from '@powersync/service-types';
+
+export interface ReportStorageFactory extends AsyncDisposable {
```
Rentacookie: Doesn't look like this is a factory at all. We can probably just call it `ReportStorage`

modules/module-mongodb-storage/src/storage/MongoReportStorage.ts
```diff
@@ -0,0 +1,193 @@
...
+
+export class MongoReportStorage implements storage.ReportStorageFactory {
+  private readonly client: mongo.MongoClient;
+  public readonly db: PowerSyncMongo;
+
+  constructor(db: PowerSyncMongo) {
+    this.client = db.client;
+    this.db = db;
+  }
+
```
Rentacookie, line 98: Small nitpick here, but especially for classes that implement interfaces or abstract classes it is more readable to have the implemented public/ abstract methods be right at the top.

modules/module-mongodb-storage/src/storage/implementation/db.ts
```diff
@@ -37,6 +38,7 @@ export class PowerSyncMongo {
   readonly locks: mongo.Collection<lib_mongo.locks.Lock>;
   readonly bucket_state: mongo.Collection<BucketStateDocument>;
   readonly checkpoint_events: mongo.Collection<CheckpointEventDocument>;
+  readonly sdk_report_events: mongo.Collection<SdkConnectDocument>;
```
Rentacookie: This doesn't quite match the naming convention of the other collections here. The interface should probably be called `SdkConnectEventDocument` and the class variable `sdk_connect_events`?

packages/types/src/events.ts
```diff
@@ -0,0 +1,76 @@
...
+  connect_at: Date;
+  sdk: string;
+} & SdkUserData;
+
+export type SdkDisconnectEventData = {
+  disconnect_at: Date;
+  connect_at: Date;
+} & SdkUserData;
+
+export type SdkConnectDocument = {
```
Rentacookie: This should move to `modules/module-mongodb-storage/src/storage/implementation/models.ts` along with the other mongo document definitions 🤔

packages/service-core/src/storage/ReportStorageFactory.ts
```diff
@@ -0,0 +1,9 @@
...
+  reportSdkConnect(data: event_types.SdkConnectBucketData): Promise<void>;
```
Rentacookie: I would recommend moving the types that are being used as parameters here into this file. 
The only times we really split the types from the API is if those types are being used in external routes and in those cases the types are defined with ts-codec anyway

packages/service-core/src/emitters/EmitterEngine.ts
```diff
@@ -0,0 +1,39 @@
+import EventEmitter from 'node:events';
+import { logger } from '@powersync/lib-services-framework';
+import { event_types } from '@powersync/service-types';
+import { BaseEmitterEngine } from './emitter-interfaces.js';
+
+export class EmitterEngine implements BaseEmitterEngine {
```
Rentacookie: To bring this more in line with the naming of our other engines - this should probably be called `EventsEngine`. And the package would just be called `events`

packages/service-core/src/emitters/emitter-interfaces.ts
```diff
@@ -0,0 +1,12 @@
+import { event_types } from '@powersync/service-types';
+
+export interface BaseEmitterEngine {
```
Rentacookie: Does not look like we need this interface? Generally all the Engines are just classes.

packages/types/src/events.ts
```diff
@@ -0,0 +1,76 @@
...
+  sdk: string;
+  user_agent: string;
+  client_id: string;
+  user_id: string;
+  jwt_exp?: Date;
+  connect_at: Date;
+  disconnect_at?: Date;
+};
+
+export type ListCurrentConnections = {
```
Rentacookie: This seems to be the return type for the listCurrentConnections API call and the 
scrapeSdkData API call. If we want to do that we should probably give this a more general name like: `Connections`, `ConnectionReport`, `ConnectedClients`

modules/module-postgres-storage/src/types/models/SdkReporting.ts
```diff
@@ -0,0 +1,23 @@
+import * as t from 'ts-codec';
+import { bigint, jsonb } from '../codecs.js';
+
+export const Sdks = t.object({
```
Rentacookie, line 4: Are these types being sent over the wire or used on endpoints? If not then there is no need to use `ts-codec` to define these.

packages/service-core/src/storage/StorageProvider.ts
Rentacookie, line 1: We should probably rename `BucketStorageProvider` => `StorageProvider` since it no longer only provides bucket storage

packages/types/src/events.ts
```diff
@@ -0,0 +1,76 @@
...
+  user_agent?: string;
+  jwt_exp?: Date;
+  connect_at: Date;
+};
+
+export type DeleteOldSdkData = {
+  date: Date;
+};
+
+export type SdkConnectEventData = {
```
Rentacookie: Seems this event is exactly the same as SdkUserData so can be removed

packages/types/src/events.ts
```diff
@@ -0,0 +1,76 @@
...
+  connect_at: Date;
+} & SdkUserData;
+
+export type SdkConnectBucketData = {
+  connect_at: Date;
+  sdk: string;
+} & SdkUserData;
+
+export type SdkDisconnectEventData = {
+  disconnect_at: Date;
```
Rentacookie: I'm wondering since`connect_at` and `disconnect_at` refer to past events if they should rather be called:
`connected_at` and `disconnected_at`

packages/service-core/src/system/ServiceContext.ts
```diff
@@ -66,6 +69,8 @@ export class ServiceContextContainer implements ServiceContext {
       }
     });
 
+    this.emitterEngine = new EmitterEngine();
```
Rentacookie: We should probably hook this up to the lifecycle like the other engines, even if nothing happens on start/stop for now.

packages/service-core/src/routes/endpoints/sync-stream.ts
```diff
@@ -88,6 +94,7 @@ export const syncStreamed = routeDefinition({
         objectMode: false,
         highWaterMark: 16 * 1024
       });
+      service_context.emitterEngine.emit(event_types.EmitterEngineEvents.SDK_CONNECT_EVENT, sdkData);
```
Rentacookie: Nitpick, maybe move this up to line 79 to keep it at the same place we capture the metric?

packages/service-core/src/routes/endpoints/socket-route.ts
```diff
@@ -6,22 +6,30 @@ import * as util from '../../util/util-index.js';
...
         user_agent: context.user_agent
       };
-      const streamStart = Date.now();
+
+      const sdkData: event_types.SdkUserData = {
+        client_id: params.client_id,
+        user_id: context.user_id!,
+        user_agent: context.user_agent,
+        jwt_exp: context.token_payload?.exp ? new Date(context.token_payload.exp * 1000) : undefined,
+        connect_at: new Date(streamStart)
```
Rentacookie: Do we get any benefit from wrapping the already captured date in a new Date?

packages/service-core/src/emitters/EmitterEngine.ts
```diff
@@ -0,0 +1,39 @@
...
+  private emitter: EventEmitter;
+  events: Set<event_types.EmitterEngineEvents> = new Set();
```
Rentacookie: Since we expose these events via the public getter `listEvents`, does this class variable also still need to be exposed?

modules/module-mongodb-storage/src/migrations/db/migrations/1752661449910-sdk-reporting.ts
EspressoTrip-v2: good catch

modules/module-mongodb-storage/src/storage/MongoReportStorage.ts
EspressoTrip-v2: I did it this way for consistency in the repo. [MongoBucketStorage Class](https://github.com/powersync-ja/powersync-service/blob/d49bebe28d44471eed766d229bc51be0e4bcd65c/modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts#L17 is implemented the same way.)

Rentacookie: Unfortunately some bad naming is hiding the fact that MongoBucketStorage is actually a factory of `MongoSyncBucketStorage` instances. The `getInstance` method creates them.

From what I can see MongoReport storage does not create anything similar.

modules/module-postgres-storage/src/types/models/SdkReporting.ts
EspressoTrip-v2: They are used for type validation inside the db.sql.decoded return function for the postgres sdk scrape implementation

modules/module-mongodb-storage/src/storage/MongoReportStorage.ts
Rentacookie: Hmmm but looking at that class `MongoBucketStorage ` all the implemented methods are at the top?

EspressoTrip-v2: sorry thought you meant they way the interface is implemented on the class. Did not read that correctly

packages/service-core/src/routes/endpoints/socket-route.ts
EspressoTrip-v2: streamStart is ms I am using that to create a Date object required in the  events.

packages/service-core/src/routes/endpoints/sync-stream.ts
EspressoTrip-v2: Thsi was on line 80 till the merge with Simons commit. I put it back.

packages/service-core/src/storage/ReportStorageFactory.ts
EspressoTrip-v2: almost all are used in the ps-management service module and endpoints. I can move the SdkConnectBucketData locally if you prefer.

packages/types/src/events.ts
EspressoTrip-v2: Yeah this was an old type that i only used in one place till we changed the whole payload responses to the view it is. I will update this

packages/types/src/events.ts
EspressoTrip-v2: I have renamed the type, its being used in postgres module as well.

modules/module-mongodb-storage/src/migrations/db/migrations/1752661449910-connection-reporting.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+    if (await db.connection_report_events.indexExists('connection_user_id_index')) {
+      await db.connection_report_events.dropIndex('connection_user_id_index');
+    }
+    if (await db.connection_report_events.indexExists('connection_client_id_index')) {
+      await db.connection_report_events.dropIndex('connection_client_id_index');
+    }
+    if (await db.connection_report_events.indexExists('connection_index')) {
+      await db.connection_report_events.dropIndex('connection_index');
+    }
```
rkistner: No need for dropping the individual indexes when the entire collection is dropped - these will be dropped automatically with the collection.

modules/module-mongodb-storage/src/storage/implementation/MongoTestReportStorageFactoryGenerator.ts
```diff
@@ -0,0 +1,22 @@
+import { TestStorageOptions } from '@powersync/service-core';
+import { connectMongoForTests } from './util.js';
+import { MongoReportStorage } from '../MongoReportStorage.js';
+
+export type MongoTestStorageOptions = {
+  url: string;
+  isCI: boolean;
+};
+
+export const MongoTestReportStorageFactoryGenerator = (factoryOptions: MongoTestStorageOptions) => {
```
rkistner: Can this be moved to the tests? Alternatively, add a comment here on why this needs to be in the main src dir.

packages/types/src/reports.ts
```diff
@@ -0,0 +1,75 @@
+export enum EventsEngineEventType {
+  SDK_CONNECT_EVENT = 'sdk-connect-event',
+  SDK_DISCONNECT_EVENT = 'sdk-disconnect-event',
+  SDK_DELETE_OLD = 'sdk-delete-old'
+}
+
+export type TimeFrames = 'hour' | 'day' | 'week' | 'month';
```
rkistner: Where is this used?

packages/types/src/reports.ts
```diff
@@ -0,0 +1,75 @@
...
+export type SubscribeEvents = {
+  [EventsEngineEventType.SDK_CONNECT_EVENT]: ClientConnectionEventData;
+  [EventsEngineEventType.SDK_DISCONNECT_EVENT]: ClientDisconnectionEventData;
+  [EventsEngineEventType.SDK_DELETE_OLD]: DeleteOldConnectionData;
```
rkistner, line 14: What emits this event?

packages/types/src/reports.ts
```diff
@@ -0,0 +1,75 @@
...
+export interface EmitterEvent<K extends EventsEngineEventType> {
+  event: K;
+  handler: EventHandlerFunc<K>;
+}
+
+export type ConnectedUserData = {
+  client_id?: string;
+  user_id: string;
+  user_agent?: string;
+  jwt_exp?: Date;
```
rkistner: We can make this required - added comments elsewhere.

packages/service-core/src/routes/endpoints/socket-route.ts
```diff
@@ -1,27 +1,34 @@
...
         client_id: params.client_id,
         user_agent: context.user_agent
       };
-      const streamStart = Date.now();
+
+      const sdkData: event_types.ConnectedUserData & event_types.ClientConnectionEventData = {
+        client_id: params.client_id,
+        user_id: context.user_id!,
+        user_agent: context.user_agent,
+        jwt_exp: context.token_payload?.exp ? new Date(context.token_payload.exp * 1000) : undefined,
```
rkistner: Same as for sync-stream, the `exp` value must always be present at this point.

modules/module-postgres-storage/src/types/models/SdkReporting.ts
```diff
@@ -0,0 +1,23 @@
...
+  sdk: t.string,
+  clients: t.number,
+  users: t.number
+});
+
+export type Sdks = t.Encoded<typeof Sdks>;
+
+export const SdkReporting = t.object({
```
rkistner, line 12: ~It would help to have comments explain what these fields mean (same on `Sdks` above).~

I now see this is only used internally to define the query results - can just add a comment indicating that here.

packages/service-core/src/events/EventsEngine.ts
```diff
@@ -0,0 +1,38 @@
...
+    this.emitter.on(event.event, event.handler.bind(event));
+  }
+
+  get listEvents(): event_types.EventsEngineEventType[] {
+    return Array.from(this.events.values());
+  }
+
+  countListeners(eventName: event_types.EventsEngineEventType): number {
+    return this.emitter.listenerCount(eventName);
+  }
```
rkistner: Is this used anywhere?

modules/module-postgres-storage/src/storage/PostgresReportStorageFactory.ts
```diff
@@ -0,0 +1,308 @@
...
+    return {
+      gte: new Date(year, month, today).toISOString(),
+      lt: new Date(year, month, nextDay).toISOString()
+    };
+  }
+
+  async reportClientConnection(data: event_types.ClientConnectionBucketData): Promise<void> {
+    const { sdk, connected_at, user_id, user_agent, jwt_exp, client_id } = data;
+    const connectIsoString = connected_at.toISOString();
+    const jwtExpIsoString = jwt_exp!.toISOString();
```
rkistner: Here you do expect that `jwt_exp` is always set - rather have it as required on the type.

packages/service-core/src/storage/ReportStorage.ts
```diff
@@ -0,0 +1,11 @@
+import { event_types } from '@powersync/service-types';
+
+export interface ReportStorage extends AsyncDisposable {
+  reportClientConnection(data: event_types.ClientConnectionBucketData): Promise<void>;
+  reportClientDisconnection(data: event_types.ClientDisconnectionEventData): Promise<void>;
+  getConnectedClients(data: event_types.ClientConnectionsRequest): Promise<event_types.ClientConnectionReport>;
+  getClientConnectionReports(
+    data: event_types.ClientConnectionReportRequest
+  ): Promise<event_types.ClientConnectionReport>;
+  deleteOldConnectionData(data: event_types.DeleteOldConnectionData): Promise<void>;
```
rkistner, line 38: Each of these need a doc-comment explaining what it does, and what the expected input and outputs are. (Inputs and output descriptions can alternatively be specified in the types)

modules/module-postgres-storage/src/storage/PostgresReportStorageFactory.ts
```diff
@@ -0,0 +1,308 @@
...
+      };
+    }
+    return {
+      users: Number(result.users),
+      sdks: result.sdks?.data || []
+    };
+  }
+  private async listConnectionsQuery(data: event_types.ClientConnectionsRequest) {
+    const { range } = data;
+    if (!range) {
```
rkistner: This query appears the same as the one below, just without the `connected_at` filters. I'd recommend only using the query below, and just set the start date far enough back to include all results, for example `new Date(0)`.

modules/module-postgres-storage/src/storage/PostgresReportStorageFactory.ts
```diff
@@ -0,0 +1,308 @@
...
+        filtered AS (
+          SELECT
+            *
+          FROM
+            connection_report_events
+          WHERE
+            disconnected_at IS NULL
+            AND jwt_exp > NOW()
+            AND connected_at >= ${{ type: 1184, value: gt }}
+            AND connected_at <= ${{ type: 1184, value: lt }}
```
rkistner: The filters here appear somewhat contradicting - the `connected_at` filters imply this can be used to fetch connections at a time in the past, but the `disconnected_at` and `jwt_exp` filter implies this excludes clients that have disconnected already.

(As mentioned elsewhere, it would help to document the expected behavior - that might clear this up).

modules/module-mongodb-storage/src/storage/MongoReportStorage.ts
```diff
@@ -0,0 +1,195 @@
...
+  }
+  async getConnectedClients(data: event_types.ClientConnectionsRequest): Promise<event_types.ClientConnectionReport> {
+    const timeframeFilter = this.listConnectionsDateRange(data);
+    const result = await this.db.connection_report_events
+      .aggregate<event_types.ClientConnectionReport>([
+        {
+          $match: {
+            disconnected_at: { $exists: false },
+            jwt_exp: { $gt: new Date() },
+            ...timeframeFilter
```
rkistner: Same as commented on Postgres storage - we should clarify whether this is expected to only be for current connections, or whether the timeframe can be used to fetch connections in the past.

modules/module-postgres-storage/src/migrations/scripts/1684951997326-init.ts
```diff
@@ -128,6 +128,28 @@ export const up: migrations.PowerSyncMigrationFunction = async (context) => {
         CONSTRAINT unique_user_sync PRIMARY KEY (user_id, sync_rules_id)
       );
     `.execute();
+    await db.sql`
+      CREATE TABLE connection_report_events (
+        id TEXT PRIMARY KEY,
+        user_agent TEXT NOT NULL,
+        client_id TEXT NOT NULL,
```
rkistner: `client_id` is specified as `NOT NULL` here, but the code implies it can be null (that is typical for old clients that didn't have `client_id` yet).

modules/module-postgres-storage/src/storage/PostgresReportStorageFactory.ts
```diff
@@ -0,0 +1,308 @@
...
+    const connectIsoString = connected_at.toISOString();
+    await this.db.sql`
+      UPDATE connection_report_events
+      SET
+        disconnected_at = ${{ type: 1184, value: disconnectIsoString }},
+        jwt_exp = NULL
+      WHERE
+        user_id = ${{ type: 'varchar', value: user_id }}
+        AND client_id = ${{ type: 'varchar', value: client_id }}
+        AND connected_at = ${{ type: 1184, value: connectIsoString }}
```
rkistner, line 171: While the probability is low, there could be multiple connections with the same `(user_id, client_id, connected_at)` combination. I'd recommend using a unique id per connection instead. Two options:

1. Use the existing `rid` (request id) we generate for the logger - that could perhaps form the primary key here. This would be nice in that we could correlate these connections with the logs, but it would require some refactoring to make that id available in the request context.
2. Return the id from the `reportClientConnection` event, then use that id when disconnecting.

modules/module-mongodb-storage/src/storage/MongoReportStorage.ts
```diff
@@ -0,0 +1,195 @@
...
+      }
+    );
+  }
+  async reportClientDisconnection(data: event_types.ClientDisconnectionEventData): Promise<void> {
+    const { connected_at, user_id, client_id } = data;
+    await this.db.connection_report_events.findOneAndUpdate(
+      {
+        client_id,
+        user_id,
+        connected_at
```
rkistner, line 67: Same as for postgres - we should use connection ids here, rather than assume this combination is unique.

modules/module-mongodb-storage/test/src/connection-report-storage.test.ts
```diff
@@ -0,0 +1,256 @@
...
+): Partial<event_types.ClientConnection & { _id: string }>[] {
+  return sdks.map((sdk: Partial<event_types.ClientConnection & { _id: string }>) => {
+    const { _id, disconnected_at, connected_at, jwt_exp, ...rest } = sdk;
+    return {
+      ...rest
+    };
+  });
+}
+
+describe('SDK reporting storage', async () => {
```
rkistner: Tests should also specifically include a case where `client_id` is undefined.

modules/module-postgres-storage/test/src/connection-report-storage.test.ts
```diff
@@ -0,0 +1,339 @@
...
+function removeVolatileFields(sdks: event_types.ClientConnection[]): Partial<event_types.ClientConnection>[] {
+  return sdks.map((sdk) => {
+    const { id, disconnected_at, connected_at, jwt_exp, ...rest } = sdk;
+    return {
+      ...rest
+    };
+  });
+}
+
+describe('Connection report storage', async () => {
```
rkistner: Same as for MongoDB, tests should also specifically include a case where `client_id` is undefined.

rkistner: Additionally, is it feasible for these tests to primarily use the public storage APIs, rather than raw SQL? That way the tests could be shared between the Postgres and MongoDB implementations, potentially only having small number of db-specific tests if you do still need the lower-level access in some places.

modules/module-mongodb-storage/src/storage/MongoReportStorage.ts
```diff
@@ -0,0 +1,195 @@
...
+    const { year, month, today } = this.parseJsDate(new Date());
+    const nextDay = today + 1;
+    return {
+      user_id: userId,
+      client_id: clientId,
+      connected_at: {
+        // Need to create a new date here to sett the time to 00:00:00
+        $gte: new Date(year, month, today),
+        $lt: new Date(year, month, nextDay)
+      }
```
rkistner: This depends on the timezone of the server - not sure if that is what we want?

But I'm also wondering - do we specifically want this behavior ("if the connection event is within the same day the stored initial connection is updated")? Why not store these as separate connection events, and then handle any other logic in the aggregations?

packages/types/src/reports.ts
EspressoTrip-v2: This was moved to another repo so needs to be removed.

modules/module-mongodb-storage/src/storage/implementation/MongoTestReportStorageFactoryGenerator.ts
EspressoTrip-v2: It can be, I have kept it here because the MongoTestStorageFactoryGenerator.ts lives here as well.

packages/service-core/src/routes/endpoints/socket-route.ts
EspressoTrip-v2: Okay the actual Context type has this optional same as user_agent. I was not too sure why the previous dev did this, but I can either enforce it or see if I can change the Context type?

packages/types/src/reports.ts
EspressoTrip-v2: This is a side effect of the Context object in most part

modules/module-mongodb-storage/src/storage/MongoReportStorage.ts
EspressoTrip-v2: Dylan was concerned about storing the connections as logs in the ps service db, due to volume. Considering we are using per day, per week and per month, we only need day granularity. When a user refreshes the web sdk it actually causes a disconnect and a reconnect which would make multiple copies withing very a short period if they refresh a few times. So to reduce the redundant connections by the same user I did it this way. Which still gives us the minimal per day granularity. 

The date objects get converted to UTC time by Mongo.

modules/module-postgres-storage/src/storage/PostgresReportStorageFactory.ts
EspressoTrip-v2: I originally used that, but when we moved to updating an existing document. It was no longer viable option.

packages/types/src/reports.ts
EspressoTrip-v2: In this instance its the report module.

modules/module-postgres-storage/src/storage/PostgresReportStorageFactory.ts
EspressoTrip-v2: > While the probability is low, there could be multiple connections with the same `(user_id, client_id, connected_at)` combination. I'd recommend using a unique id per connection instead. Two options:
> 
> 1. Use the existing `rid` (request id) we generate for the logger - that could perhaps form the primary key here. This would be nice in that we could correlate these connections with the logs, but it would require some refactoring to make that id available in the request context.
> 2. Return the id from the `reportClientConnection` event, then use that id when disconnecting.

Could you give me an example when this (user_id, client_id, connected_at) might occur?
My understanding is the client_id is basically linked to the client db and would only change if there was a reinstall of the application... So would that mean that there is an instance where an application could be installed by the same user on a different device connected to the same instance and have the same client_id generated but with a different sdk?

modules/module-mongodb-storage/src/storage/MongoReportStorage.ts
rkistner: Ok, we can keep it as daily connections for now - it looks like we can change that later if needed without too much effort.

> The date objects get converted to UTC time by Mongo.

If we do keep daily values, we need to be very explicit about time zones. As is, if I run the server in UTC+2, the filter here would use `2025-08-21T22:00:00.000Z`. If the reporting job happens to use a different timezone, it would not match up with the way it is persisted here.

Now in our hosted environment at least, it is likely that the service runs in the UTC timezone already, and this won't make a difference. But if we're relying on that, it is better to be explicit - use UTC timezones here (`new Date(Date.UTC(year, month, today))`), and document that UTC-based timestamps should be used when querying the data.

modules/module-postgres-storage/src/storage/PostgresReportStorageFactory.ts
rkistner: In most cases the client_id _should_ be unique, but we don't have any guarantees on that. For example, the user can copy the database to another device (maybe not on Android/iOS, but on desktop it's easy), in which case the same client_id would be present on both. Older SDKs do not send client_id at all. And in some cases, the user_id is not unique either.

When combining with the connect_at timestamp it should be very rare that you get duplicates, but it is possible.

That said, since we're only "pre-aggregating" these connections per (user_id, client_id, day) anyway, I guess that doesn't really make a difference, and we can keep this as-is.

packages/service-core/src/storage/ReportStorage.ts
```diff
@@ -0,0 +1,41 @@
...
+   */
+  reportClientDisconnection(data: event_types.ClientDisconnectionEventData): Promise<void>;
+  /**
+   * Get currently connected clients.
+   * This will return any short or long term connected clients.
+   * Clients that have no disconnected_at timestamp and that have a valid jwt_exp timestamp are considered connected.
+   * @param data.range.start Needs to be UTC time string
+   * @param data.range.end Optional needs to be UTC time string
+   */
+  getConnectedClients(data: event_types.ClientConnectionsRequest): Promise<event_types.ClientConnectionReportResponse>;
```
rkistner: This helps, but still need more detail here.

The description says "currently connected clients", but the parameters allow passing in a timestamp range. So for example if pass in `{start: '2025-08-01T00:00Z', end: '2025-08-01T00:00Z'}`, would I get a list of clients who were connected at that time? Or in which cases should the time range be used?

modules/module-mongodb-storage/src/storage/MongoReportStorage.ts
EspressoTrip-v2: I understand, I made a note on the report storage interface that date strings are to be UTC. This code block is actually an internal update from within the server so no query strings are passed from outside. The logic of create and update are both the same, so the dates will always be the same timezone. I will document the UTC timezones in the relevant sdk where date ranges are passed as queries.

rkistner: You still need to use the UTC methods for this specific case, so that the recording doesn't depend on the service timezone. That includes how the Date is constructed here (see my example above), as well as the methods used in `parseJsDate()`.

modules/module-postgres-storage/test/src/connection-report-storage.test.ts
EspressoTrip-v2: > Same as for MongoDB, tests should also specifically include a case where `client_id` is undefined.

If the sdk has no client_id it then would also not have an sdk in the user_agent. Can I duplicate the user_id in the client_id entry, the sdk would in any case be stored as an unknown during the user_agent parse.

rkistner: There should still be a user-agent in those cases, although it would likely be the generic platform-specific one.

I'd recommend using a blank string for the client_id if you want to avoid using `undefined`.

modules/module-mongodb-storage/test/src/connection-report-storage.test.ts
EspressoTrip-v2: If client_id is undefined, I have made it an empty string. It is added as a user connection in the tests.

modules/module-postgres-storage/src/migrations/scripts/1684951997326-init.ts
```diff
@@ -128,6 +128,28 @@ export const up: migrations.PowerSyncMigrationFunction = async (context) => {
...
+
+    await db.sql`
+      CREATE INDEX sdk_list_index ON connection_report_events (connected_at, jwt_exp, disconnected_at)
+    `.execute();
+
+    await db.sql`CREATE INDEX sdk_user_id_index ON connection_report_events (user_id)`.execute();
+
+    await db.sql`CREATE INDEX sdk_client_id_index ON connection_report_events (client_id)`.execute();
+
+    await db.sql`CREATE INDEX sdk_index ON connection_report_events (sdk)`.execute();
```
rkistner: These must be in a new migration script, not an update to the existing one

modules/module-postgres-storage/src/storage/PostgresReportStorageFactory.ts
```diff
@@ -0,0 +1,258 @@
...
+    return await this.db.sql`
+      WITH
+        filtered AS (
+          SELECT
+            *
+          FROM
+            connection_report_events
+          WHERE
+            disconnected_at IS NULL
+            AND jwt_exp > NOW() AT TIME ZONE 'UTC'
```
rkistner: Both `jwt_exp` and `NOW()` have type `timestamp with time zone` (timestamptz), meaning the timestamp is effectively persisted and compared in UTC timezone.

`AT TIME ZONE 'UTC'` would convert it to a `timestamp without time zone` (timestamp) at UTC, which would at best have no effect here, and at worst break the comparison when the database or service runs in a different timezone (since you're comparing `timestamptz` with a `timestamp`, which does timezone conversions automatically using an unknown timezone).

EspressoTrip-v2: Sorry, was testing stuff after the Mongo, never removed it...

modules/module-postgres-storage/src/migrations/scripts/1684951997326-init.ts
EspressoTrip-v2: I have moved it out, I saw in Mongo there are separate migrations however Postgres has combined, took it that there was a reason for that.

rkistner: It must still be removed here.

FWIW, the Postgres module only had two migration files since it was created later than the MongoDB one, and hasn't has only had one change needing a migration since then.

modules/module-mongodb-storage/src/storage/MongoReportStorage.ts
```diff
@@ -0,0 +1,177 @@
+import { mongo } from '@powersync/lib-service-mongodb';
+import { storage } from '@powersync/service-core';
+import { event_types } from '@powersync/service-types';
+import { PowerSyncMongo } from './implementation/db.js';
+import { logger } from '@powersync/lib-services-framework';
+
+export class MongoReportStorage implements storage.ReportStorage {
+  private readonly client: mongo.MongoClient;
```
Rentacookie: This private readonly field looks to be unused.

modules/module-mongodb-storage/test/src/connection-report-storage.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      sdk: userData.user_week.sdk,
+      connected_at: newConnectAt,
+      jwt_exp: jwtExp,
+      client_id: userData.user_week.client_id,
+      user_id: userData.user_week.user_id,
+      user_agent: userData.user_week.user_agent
+    });
+
+    const sdk = await factory.db.connection_report_events.find({ user_id: userData.user_week.user_id }).toArray();
+    expect(sdk).toHaveLength(2);
```
Rentacookie: I'm probably just missing something, but why are we expecting connections to be 2, after only inserting 1?
Nit: Looks like the old `sdk` names for the variables are still being used here.

Rentacookie: Ah I see now - there is some initial test data being inserted before every test.

modules/module-postgres-storage/src/migrations/scripts/1684951997326-init.ts
```diff
@@ -128,6 +128,28 @@ export const up: migrations.PowerSyncMigrationFunction = async (context) => {
         CONSTRAINT unique_user_sync PRIMARY KEY (user_id, sync_rules_id)
       );
     `.execute();
+    await db.sql`
```
Rentacookie: - [ ] Remove the new table and indexes from old migration script

modules/module-postgres-storage/src/storage/PostgresReportStorageFactory.ts
```diff
@@ -0,0 +1,258 @@
...
+import { toInteger } from 'ix/util/tointeger.js';
+import { logger } from '@powersync/lib-services-framework';
+import { getStorageApplicationName } from '../utils/application-name.js';
+import { STORAGE_SCHEMA_NAME } from '../utils/db.js';
+
+export type PostgresReportStorageOptions = {
+  config: NormalizedPostgresStorageConfig;
+};
+
+export class PostgresReportStorageFactory implements storage.ReportStorage {
```
Rentacookie: This class is and file is still called `PostgresReportStorageFactory` instead of `PostgresReportStorage` to match the renamed interface.
The implemented interface methods should also ideally be moved to the top to match the pattern in the MongoReportStorage

rkistner (approved):

PR #320, opened 2025-08-07
Update dependencies

rkistner:
Notably:
 * Node v22.17.1 (minor security update only affecting Windows)
 * Fastify 5.4.0
 * @sentry/node 10.2.1
 * opentelemetry 2.0.1 (also used by sentry) (has some breaking API changes)
 
Also includes a missing license field, and moves a test dependency to devDependencies.

### DIFF
--- a/.changeset/olive-games-destroy.md
+++ b/.changeset/olive-games-destroy.md
@@ -1,0 +1,6 @@
+---
+'@powersync/service-sync-rules': minor
+'@powersync/service-image': minor
+---
+
+Add the `fixed_json_extract` compatibility option. When enabled, JSON-extracting operators are updated to match SQLite more closely.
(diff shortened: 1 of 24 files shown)


packages/service-core/src/metrics/open-telemetry/util.ts
```diff
@@ -43,31 +39,34 @@ export function createOpenTelemetryMetricsFactory(context: ServiceContext): Metr
...
     }
   });
 
+  const resource = resourceFromAttributes({
+    ['service']: 'PowerSync',
+    ['service.version']: pkg.version,
+    ['instance_id']: instanceIdPromise
+  });
+
+  resource.waitForAsyncAttributes?.();
```
stevensJourney, line 67: Does triggering this method kick-off something in the background? Just wondering why we call this when we don't (and can't since this is a sync function) await it here.

rkistner: Yes - added comment now.

PR #329, opened 2025-08-18
Compression for sync streams

rkistner:
Background on compression options we're investigating in PowerSync: https://github.com/powersync-ja/powersync-service/discussions/330

This enables gzip and zstd as transport compression options for sync streams. Since data sync is often performed on metered mobile connections or slow WiFi connections, this could reduce data costs and increase performance for those cases.

This uses the standard `Accept-Encoding` request header, and `Content-Encoding` response header. As an example, Chrome would send `accept-encoding: gzip, deflate, br, zstd`, then we'd response with `content-encoding: zstd`.

`zstd` is the best option, and is supported by default in Chrome 123+ and Firefox 126+. `gzip` is provided as a fallback option, and is supported in Safari and many native clients.

Since browsers send the `accept-encoding` header by default, this will have an immediate effect for the Web SDK. For other SDKs we'd have to test individually, and see whether we need more client-side changes to enable compression.

## Performance

These are not scientific benchmarks, but just gives an initial idea of expected performance. Since decompression is performed by platform-provided native code in each SDK, this adds minimal overhead and minimal client-side changes (we're not manually decompressing using JavaScript, for example).

Synthetic test dataset: 1.6 million rows of data, 384MiB uncompressed download size, 117MiB data size (the rest is metadata).

zstd compression gives 72.4MiB download size (19% of the original size); gzip gives 76.2MiB (20% of the original size).

In all cases (uncompressed, zstd, gzip), it took around 13s to download the data with curl, and 52s with the diagnostics app. So on a desktop machine, the primary difference is in bandwidth usage, rather than performance.

When downloading the data with curl, the nodejs process averaged around 90% CPU usage for uncompressed data, and 120% when using zstd or gzip. Note that it uses more than 1 core since compression uses a separate thread.

## Testing clients

OkHttp supports gzip by default - likely affects both React-Native and Kotlin.

Chrome:
 * Works with zstd and gzip out-of-the-box: `accept-encoding: gzip, deflate, br, zstd`

NodeJS SDK:
 * Uses gzip out of the box: `accept-encoding: gzip, deflate`
 * We can make minor SDK modifications to use zstd on Node versions that support it

Dart/Flutter:
 * Uses gzip out of the box: `accept-encoding: gzip`

React-Native on Android:
 * Uses gzip out of the box: `accept-encoding: gzip`

Kotlin on Android:
 * Uses gzip out of the box: `accept-encoding: gzip`

## TODO

 * [x] Test for issues and performance on each SDK
 * [x] ~Proper benchmarks~ According to my initial tests, performance is on par to uncompressed speeds when bandwidth is not a limitation. Will test constrained bandwidth environments later.
 * [x] ~Test `permessage-deflate` for websockets~ #337

### DIFF
--- a/packages/service-core/src/routes/compression.ts
+++ b/packages/service-core/src/routes/compression.ts
@@ -1,5 +1,5 @@
-import { PassThrough, pipeline, Readable, Transform } from 'node:stream';
 import type Negotiator from 'negotiator';
+import { PassThrough, pipeline, Readable, Transform } from 'node:stream';
 import * as zlib from 'node:zlib';
 import { RequestTracker } from '../sync/RequestTracker.js';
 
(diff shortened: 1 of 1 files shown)


packages/service-core/src/routes/endpoints/sync-stream.ts
```diff
@@ -80,10 +80,11 @@ export const syncStreamed = routeDefinition({
       });
 
       const byteContents = useBson ? sync.bsonLines(syncLines) : sync.ndjson(syncLines);
-      const stream = Readable.from(sync.transformToBytesTracked(byteContents, tracker), {
+      const plainStream = Readable.from(sync.transformToBytesTracked(byteContents, tracker), {
```
simolus3, line 85: I'm not sure if we need both or if we want to update the metric for billing, but would it make sense to count bytes after applying compression here?

rkistner: For PowerSync Cloud, billing is based on uncompressed size. But either way it could make sense to track the compressed size as well as a separate metric

rkistner: Added now as `powersync_data_sent_bytes_total`.

packages/service-core/src/routes/compression.ts
```diff
@@ -0,0 +1,78 @@
...
+        // This does the actual compression in a background thread pool.
+        zlib.createZstdCompress({
+          // We need to flush the frame after every new input chunk, to avoid delaying data
+          // in the output stream.
+          flush: zlib.constants.ZSTD_e_flush,
+          params: {
+            // Default compression level is 3. We reduce this slightly to limit CPU overhead
+            [zlib.constants.ZSTD_c_compressionLevel]: 2
+          }
+        }),
```
simolus3: minor nit, feel free to ignore: Since this is the only part that is meaningfully different between `zstd` and `gzip`, can we move it into a separate function? Then `maybeCompressResponseStream` could take the `stream.Transform` instance for the compressor and the common `tracker.setCompressed()`, `transform` and `encodingHeaders` lines can be shared.

simolus3 (dismissed):
Looks good to me 👍 

I also wonder if we want to consider supporting brotli to give Safari something better than gzip, but that may be to small to matter.

simolus3 (approved):

Issue #286, opened 2025-06-24
pg timestamp/timestamptz should be mapped to ISO-8601 compatible string


ImSingee:
Currently, the timestamptz is mapped to `YYYY-MM-DD hh:mm:ss.sssZ` format and timestamp is mapped to `YYYY-MM-DD hh:mm:ss.sss`.

The document says "This is compatible with ISO8601 and SQLite’s functions.".

However, this format (use a space instead of `T` to separate date and time) isn't compatible with ISO-8601. (From [Wikipedia](https://en.wikipedia.org/wiki/ISO_8601#:~:text=Separating%20date%20and%20time%20parts%20with%20other%20characters%20such%20as%20space%20is%20not%20allowed%20in%20ISO%C2%A08601%2C%20but%20allowed%20in%20its%20profile%20RFC%203339.), `Separating date and time parts with other characters such as space is not allowed in ISO 8601, but allowed in its profile RFC 3339.`)

Additionally, SQLite [supports using `T` as separator](https://www.sqlite.org/lang_datefunc.html#:~:text=A%20time%2Dvalue%20can%20be%20in%20any%20of%20the%20following%20formats%20shown%20below.):

```
YYYY-MM-DD
YYYY-MM-DD HH:MM
YYYY-MM-DD HH:MM:SS
YYYY-MM-DD HH:MM:SS.SSS
YYYY-MM-DDTHH:MM
YYYY-MM-DDTHH:MM:SS
YYYY-MM-DDTHH:MM:SS.SSS
HH:MM
HH:MM:SS
HH:MM:SS.SSS
now
DDDDDDDDDD
```

So, I think, powersync should switch to use `T` as separator instead of the space.

Issue #299, opened 2025-07-12
Schema generations doesn't take well custom postgres types


Fabricio-191:
It's a small issue, in my postgres database (through supabase), i have a custom type:

`CREATE DOMAIN api.rating_value AS FLOAT CHECK (VALUE BETWEEN 0 AND 5);`

And the generated schema column for columns with that data type turns out to be `Column.text` instead of `Column.real` (i'm using dart/flutter)

It would be great if it could detect the subtype of the custom type, if not it would be nice a warning/comment while generating the schema like: `x column datatype not recognized, assigned text type`

stevensJourney:
Hi @Fabricio-191, Thanks for the feedback. 

We currently default the column type to `TEXT` if the detected column does not fall in to a [defined](https://github.com/powersync-ja/powersync-service/blob/1aafdaf0710d7ddda36f54b70cd1f62856b73d63/packages/sync-rules/src/ExpressionType.ts#L84) set of types.

I agree that your suggestions are could be valuable. We'll use this feedback to schedule investigating these improvements.

rkistner:
FYI, I ran into the same issue when I tried to define a custom domain to allow uploading base64-encoded data to a BYTEA column via the supabase APIs. It would be good to solve this, but could also be considered a breaking change, so we may need to make this opt-in if we add support.

rkistner:
Another report of custom nested types not working as expected - [Discord thread](https://discord.com/channels/1138230179878154300/1407695064846831646/1407695064846831646).

> An example is a column I have for files metadata, which have this definition
> 
> CREATE TYPE public.file_metadata AS
(
    etag text,
    size bigint,
    mimetype text,
    last_modified double precision,
    status file_status,
    visible boolean,
    color text,
    duration double precision
);
> And what I get in the powersync sqlite is `("",664295,image/png,1749628841028,upload_error,t,,)`

PR #331, opened 2025-08-19
Fix timestamp representation

simolus3:
The mongo and postgres replicators used to represent timestamp values with a space between the date and time parts, whereas ISO-6801 would require the letter `T`. While we would like to align the format with that standard, doing so is a backwards-incompatible change since existing rows would sync differently.

This PR introduces two new concepts to fix this issue:

1. `SqliteInputValue`: This expands the `SqliteValue` type used by sync rules to represent SQLite rows to also support a `CustomSqliteValue`. Values of that type can decay dynamically into `SqliteValue` by giving them additional context. We can thus represent date values as `CustomSqliteValue`s, that, depending on context, use the legacy or the ISO-6801 format.
2. As part of the context as pass to `CustomSqliteValue`s, this adds the `Quirk` class representing a historical issue of the sync service we can't fix without a backwards-incompatible change. Users can opt in to fixes with the `fixed_quirks` option in the sync-rules config, and there's also a `CompatibilityLevel` to fix some quirks by default. The idea is that streams would implicitly raise that level internally.

When evaluating parameter and data rows, `CustomSqliteValue`s are mapped into the actual values using the context specified in sync rules.
The Mongo and Postgres adapters have been fixed to return these custom values instead of applying a fixed format for date values.

Users can opt into this fix by adding this section to their sync rules:

```yaml
fixed_quirks:
  - non_iso8601_timestamps
```

There are a number of other issues also worth fixing (in particular, #299, handling postgis types properly, fixing the `->>` operator, encoding sync rule id in bucket identifiers). Since most of this PR is just laying the groundwork for those changes, I'll open additional PRs for those.

Closes https://github.com/powersync-ja/powersync-service/issues/286.


packages/jpgwire/src/util.ts
```diff
@@ -256,9 +257,11 @@ export function timestamptzToSqlite(source?: string) {
   if (isNaN(parsed.getTime())) {
     return null;
   }
-  const text = parsed.toISOString().replace('T', ' ').replace('.000', '').replace('Z', '');
 
-  return `${text}${precision ?? ''}Z`;
+  const baseValue = parsed.toISOString().replace('.000', '').replace('Z', '');
+  const baseText = `${baseValue}${precision ?? ''}Z`;
+
+  return new TimeValue(baseText.replace('T', ' '), baseText);
```
rkistner: I'd recommend also removing the `replace('.000', '')` transformation from the baseValue - it can break sorting.

packages/sync-rules/src/types/time.ts
```diff
@@ -0,0 +1,31 @@
...
+ * This is not ISO 6801 compatible, but changing it would be breaking existing users. So, this option is opt-in and
+ * disabled by default until a major upgrade.
+ */
+export class TimeValue extends CustomSqliteValue {
+  // YYYY-MM-DD hh:mm:ss.sss / YYYY-MM-DD hh:mm:ss.sssZ
+  readonly legacyRepresentation: string;
+  // YYYY-MM-DDThh:mm:ss.sss / YYYY-MM-DDThh:mm:ss.sssZ
+  readonly iso8601Representation: string;
+
+  constructor(legacyRepresentation: string, iso8601Representation: string) {
```
rkistner: Would it be feasible to just pass one parameter in? For example only pass in the iso8601 value, then do the legacy conversion only if needed? Or is the legacy conversion too different between the source dbs?

packages/jpgwire/src/util.ts
simolus3: Just to be clear, in that case we also wouldn't join the `precision` back in afterwards? So we just drop the fraction part here? I think that makes sense given that SQLite doesn't support it, but wouldn't that be another breaking change?

rkistner: Oh, I misread the code here - I thought we're dropping `.000` from the source timestamp, but we're dropping it from the parsed timestamp.

The core issue I was thinking of is that if you store have these two timestamps in the same Postgres column:
```
2025-08-19T12:27:58.000Z
2025-08-19T12:27:58.138Z
```

Then we'll sync it as:
```
2025-08-19T12:27:58Z
2025-08-19T12:27:58.138Z
```

If you then sort these values, they'll appear in the wrong order.

But the base issue here is that Postgres sends the timestamp(tz) as a string, dropping any trailing zeros. So just from that string alone, we can't know how many to add back. Some workarounds I can think of:
1. Use the precision of the underlying column, e.g. `timestamptz(3)` - not sure how feasible it is to get that info.
2. Use the max supported precision (6).
3. Make this user-configurable.

simolus3: Alright, so the old behavior is to keep the precision from Postgres (which would be missing trailing zeroes or entirely for whole-second values).

The new format (also applied to timestamps without timezones and time values without dates) is number 2 you've suggested, always padding the precision to 6 digits.

rkistner: Ok, happy with that. I think we just need some tests for the new format for MySQL and MongoDB as well, so that we have explicit test cases for the old and new format for each source db. (For example, for MongoDB we always have millisecond precision)

simolus3: I have added explicit date format tests for MongoDB and PostgreSQL. I don't think we need another test for MySQL because that format never used the old representation, which is implicitly asserted by this test (because it matches on the `SqliteInputRow`, it would fail if there was a custom type involved in there). https://github.com/powersync-ja/powersync-service/blob/c7ded6280911345b2043e492d29a0bc8466cc185/modules/module-mysql/test/src/mysql-to-sqlite.test.ts#L213-L233

simolus3:
I agree with giving the configuration keys a more positive name 👍 

How would we deal with a combination of `edition: 1` and sync streams being used, should that be an error? Or do we want the option of sync steams using the old behavior as an opt-out strategy?

rkistner:
I'm happy with that approach.

I'm starting to think it makes sense to just encourage specifying an explicit edition (i.e. have it in all our templates), rather than setting the default with sync streams.

> Much further into the future, make edition: 3 the default and disallow older editions.

Before we drop support for older editions we'll need to have a proper versioning and support policy. For example, we could potentially drop it in version 2.x or 3.x of the service, while still maintaining some level of support (bugfixes) for older major versions.

.changeset/olive-bags-wave.md
```diff
@@ -0,0 +1,6 @@
+---
+'@powersync/service-sync-rules': minor
+'@powersync/service-image': minor
+---
+
+Introduce the `config` option on sync rules which cna be used to opt-in to new features and backwards-incompatible fixes of historical issues with the PowerSync service.
```
rkistner:
```suggestion
Introduce the `config` option on sync rules which can be used to opt-in to new features and backwards-incompatible fixes of historical issues with the PowerSync service.
```

.changeset/popular-zoos-hang.md
```diff
@@ -0,0 +1,6 @@
+---
+'@powersync/service-sync-rules': minor
+'@powersync/service-image': minor
+---
+
+Add the `timestamps_iso8601` option in the `config:` block for sync rules. When enabled, timestmaps are consistently formatted using ISO 8601 format.
```
rkistner:
```suggestion
Add the `timestamps_iso8601` option in the `config:` block for sync rules. When enabled, timestamps are consistently formatted using ISO 8601 format.
```

PR #332, opened 2025-08-20
Remove binary_data field from protocol request

simolus3:
The `binary_data` field on sync requests is supposed to be true if clients will receive data as BSON documents instead of as JSON.

Since the sync service knows which format it'll use, this field doesn't actually have to be set by clients at all. Indeed it's not supposed to, but the Rust client sends it anyway. This could lead to an inconsistency between `binary_data` being true and binary data actually being sent:

1. Requesting a sync stream with `Accept: application/vnd.powersync.bson-stream` errors if `binary_data` is not explicitly set to true in the request, because we're encoding values as BSON but `bucketDataBatch` things it's JSON.
2. Requesting a sync stream via HTTP and JSON, but also setting `binary_data: true` is a near-miss. This is what happens today with the Rust sync client, this only works because the unencoded events emitted by `bucketDataBatch` are serialized in the `ndjson` transformer.
3. For web socket streams, nothing has changed, except that `binary_data` is now passed as a separate parameter to the sync stream.

### DIFF
--- a/modules/module-mongodb-storage/test/src/__snapshots__/storage_sync.test.ts.snap
+++ b/modules/module-mongodb-storage/test/src/__snapshots__/storage_sync.test.ts.snap
@@ -39,7 +39,7 @@ exports[`sync - mongodb > compacting data - invalidate checkpoint 2`] = `
       "bucket": "mybucket[]",
       "data": [
         {
-          "checksum": -93886621n,
+          "checksum": -93886621,
           "op": "CLEAR",
           "op_id": "2",
         },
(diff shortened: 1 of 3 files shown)


rkistner (dismissed):

modules/module-mongodb-storage/test/src/__snapshots__/storage_sync.test.ts.snap
```diff
@@ -39,7 +39,7 @@ exports[`sync - mongodb > compacting data - invalidate checkpoint 2`] = `
       "bucket": "mybucket[]",
       "data": [
         {
-          "checksum": -93886621n,
+          "checksum": -93886621,
```
simolus3, line 42: Note: These are no longer bigints because `consumeIterator` uses `JSONBig` to parse string sync lines, but now we've changed `syncResponse` to emit direct objects.

PR #333, opened 2025-08-21
Encode sync rule id in bucket id

simolus3:
When deploying new sync rules, the service will re-index everything from a consistent snapshot. Since it re-uses existing bucket ids for this, this means that oplog history checksums within a bucket change too. This is bad for clients, which may download new data ontop of their old oplog without deleting it right away. That increases the sync time (because clients have to wait for a consistent snapshot only to realize that their checksums are off and restart) and can be confusing (because sync progress appears to reach 1.0 before restarting).

A backwards-incompatible change is to encode the id of persisted sync rules in the created bucket ids. This means that every redeploy leads to different bucket ids, causing clients to drop their old buckets instead of trying to consider that data. Since this format would break existing deployments, it's only enabled for sync streams or with an opt-in.


rkistner:
We could also potentially enable this for any new sync rules deploy - in that case we'd just need to persist the state in the sync_rules storage collection/table. But not sure if that's actually worth the extra implementation effort - the approach here would eventually cover the majority of cases as we encourage users to switch to sync streams or the new defaults.

.changeset/rich-shrimps-push.md
```diff
@@ -0,0 +1,6 @@
+---
+'@powersync/service-sync-rules': minor
+'@powersync/service-image': minor
+---
+
+Add the `versioned_bucket_ids` quirk which, when opting in to the fix, generates bucket ids in a way that is more efficient for clients to adopt around sync rule redeploys.
```
rkistner: This should be updated to reflect the new config format.

rkistner (dismissed):

rkistner (approved):

PR #339, opened 2025-08-26
WIP: Better decoding of custom postgres types

simolus3:
This adds a compatibility option that, when enabled, will map custom postgres types into appropriate structures instead of syncing the raw representation.
Today, we support a static list of builtin Postgres types. Any other type will be synced as the underlying text representation as receive from the replication stream.

This PR improves that to add support for custom types. These types are resolved by querying their description from `pg_type`. Specifically, this supports:

- Domain types, which wrap an underlying type and add constraints. There isn't much to parse here, Postgres already encodes them as the underlying type. We just have to recognize this alias to apply the decoder for the underlying type oid.
- Composite types, which group multiple named values. We receive them as a tuple of raw values that we have to associate with field names when decoding to get an object back.

We also need to support nested types composed from these (e.g. a composite type could reference an array of a domain type that is itself another composite type). So we perform a BFS traversal through types when fetching them, and the decoding logic is recursive to handle these structures.

To decode the tuples that composite values are encoded as, this also refactors how arrays are parsed (since the structure is quite similar). The new decoder also supports arrays with a non-standard delimiter (which is also fetched from `pg_type`).

Closes https://github.com/powersync-ja/powersync-service/issues/299.

TODOS:

- [x] Domain types
- [x] Composite types
- [ ] Enums
- [ ] Ranges
- [ ] Test with actual sync
- [ ] Use this information in `getConnectionSchema()`.


modules/module-postgres/src/api/PostgresRouteAPIAdapter.ts
```diff
@@ -332,9 +338,11 @@ GROUP BY schemaname, tablename, quoted_name`
         if (pg_type.startsWith('_')) {
           pg_type = `${pg_type.substring(1)}[]`;
         }
+
+        const knownType = this.typeCache.registry.lookupType(Number(column.atttypid));
         table.columns.push({
           name: column.attname,
-          sqlite_type: sync_rules.expressionTypeFromPostgresType(pg_type).typeFlags,
+          sqlite_type: sync_rules.ExpressionType.fromTypeText(knownType.sqliteType()).typeFlags,
```
simolus3, line 345: It is worth noting that this doesn't take the current sync rules / compatibility options into account (I'm not sure if they're available here). So a `DOMAIN` type wrapping `FLOAT` would get represented as `TYPE_REAL` instead of `TYPE_TEXT`, even though we might replicate values as strings.

I think this may be fine because types are implemented as casts on the client-side, so the actual value returned by the view will be correct.

rkistner (approved):

packages/jpgwire/package.json
```diff
@@ -15,12 +15,15 @@
...
+    "build:tests": "tsc -b test/tsconfig.json",
+    "test": "vitest"
   },
   "dependencies": {
     "@powersync/service-jsonbig": "workspace:^",
     "@powersync/service-sync-rules": "workspace:^",
     "date-fns": "^4.1.0",
-    "pgwire": "github:kagis/pgwire#f1cb95f9a0f42a612bb5a6b67bb2eb793fc5fc87"
+    "pgwire": "github:kagis/pgwire#f1cb95f9a0f42a612bb5a6b67bb2eb793fc5fc87",
+    "vitest": "^3.0.5"
```
rkistner, line 27: For future reference, we need to be careful about adding dependencies like this in `dependencies` - should be in `devDependencies` instead. This can add some unnecessary dependencies to the Docker image - see #361.

PR #346, opened 2025-09-01
[MongoDB Storage] Checksum calculation improvements

rkistner:
Follow-up to #343. That PR implemented a "fallback" query when the checksum query timed out, which would then query each bucket individually, and limit the number of operations we aggregate on for each bucket. That approach is less likely to time out, but slower when there are many buckets.

This now merges the two approaches: We always limit the number of operations we're aggregating on, but still do the query across multiple buckets at a time. This reduces the worst-case runtime, since we don't have to wait for the first query to timeout, and can handle multiple buckets efficiently.

This fixes a specific bug in the old approach (regression in 1.15.1): The fallback query did not handle the case correctly where we already have a partial checksum for a bucket, which could result in an incorrect checksum being returned and cached. It should have been rare since it would only happen when _both_:
1. A partial checksum is already cached.
2. The query still times out.

Since the query is much less likely to time out when we have partial checksums available, it should not be common.

This also adds a couple additional tests (which would have failed with the bug above), and splits some of the larger test files into smaller ones.


modules/module-mongodb-storage/src/storage/implementation/MongoChecksums.ts
```diff
@@ -44,7 +68,7 @@ export class MongoChecksums {
   /**
    * Calculate (partial) checksums from bucket_state and the data collection.
    *
-   * Results are not cached.
+   * Results are not cached here.
```
simolus3:
```suggestion
   * This method is only called by {@link ChecksumCache.getChecksumMap}, which is responsible for caching its result.
```

modules/module-mongodb-storage/src/storage/implementation/MongoChecksums.ts
```diff
@@ -11,12 +12,34 @@ import {
...
+const DEFAULT_OPERATION_BATCH_LIMIT = 50_000;
+
 /**
  * Checksum query implementation.
+ *
+ * General implementation flow is:
+ * 1. getChecksums() -> check cache for (partial) matches. If not found or partial match, query the remainder using getChecksumsInternal().
+ * 2. getChecksumsInternal() -> query bucket_state for partial matches. Query the remainder using queryPartialChecksums().
+ * 3. queryPartialChecksums() -> split into batches of 200 buckets at a time -> queryPartialChecksumsInternal()
+ * 4. queryPartialChecksumsInternal() -> aggregate over 50_000 operations in bucket_data at a time
```
simolus3: Nit: To me at least, `get` and `query` sound like relatively cheap operations. And it sounds like at least the last three steps can be fairly expensive. Perhaps `compute` is a more appropriate verb here? What do you think about

- `getChecksumsInternal()`: Perhaps `computeChecksumsFromLastCompact` (but I know that's not exactly accurate)
- `queryPartialChecksums` and `queryPartialChecksumInternal`: Perhaps also with `compute`?

rkistner: Updated the names and descriptions - could you check if it makes more sense now?

It's tricky because they all _kind of_ do the same thing, but with varying levels of batching and/or cache utilization. Luckily only 2 of these are public.

PR #358, opened 2025-09-05
Sync rules: Export stream SQL functions

simolus3:
This updates the public `sync-rules` API to make it easier to write sync stream aware IDEs for sync rules:

- Export `STREAM_FUNCTIONS` exporting all functions available to streams (used to be internal before).
- Give `auth.user_id()` a correct `debugName` (it was `request.user_id()` before).
- Improve the types of named functions returned by `generateSqlFunctions` to make documentation and other fields available.

### DIFF
--- a/packages/sync-rules/src/request_functions.ts
+++ b/packages/sync-rules/src/request_functions.ts
@@ -107,7 +107,7 @@ export const request_jwt: SqlParameterFunction = {
   usesUnauthenticatedRequestParameters: false
 };
 
-export function generateUserIdFunction(debugName: string): SqlParameterFunction {
+export function generateUserIdFunction(debugName: string, sameAsDesc: string): SqlParameterFunction {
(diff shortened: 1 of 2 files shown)


packages/sync-rules/src/request_functions.ts
```diff
@@ -107,25 +107,27 @@ export const request_jwt: SqlParameterFunction = {
...
+    debugName,
+    parameterCount: 0,
+    call(parameters: ParameterValueSet) {
+      return parameters.userId;
+    },
+    getReturnType() {
+      return ExpressionType.TEXT;
+    },
+    detail: 'Authenticated user id',
+    documentation: "The id of the authenticated user.\nSame as `request.jwt() ->> 'sub'`.",
```
rkistner: Probably need to adjust this reference as well

PR #363, opened 2025-09-10
Move input conversion to source stream

simolus3:
This reverts the bucket batch API to only accept regular input rows (instead of `SqliteInputRow`s that need additional context to be evaluated). As a consequence, the source streams (`BinLogStream`, `ChangeStream`, `WalStream`) are now responsible for applying the context (which is not a problem because they keep a reference to the sync rules).

Applying the conversion as early as possible fixes issues when a `CustomSqliteValue` is used in a primary key, because we'd serialize it to BSON in that case. It also simplifies the bucket implementation just a tiny bit.

This also applies a small optimization to `applyRowContext` to avoid copying the row if it contains no custom values.

### DIFF
--- a/.changeset/dry-pots-leave.md
+++ b/.changeset/dry-pots-leave.md
@@ -1,0 +1,9 @@
+---
+'@powersync/service-module-postgres': patch
+'@powersync/service-module-mongodb': patch
+'@powersync/service-core': patch
+'@powersync/service-module-mysql': patch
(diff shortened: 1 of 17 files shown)


rkistner (approved):
Thanks, confirmed this fixes the issue.

I added a test that failed on versions 1.15.0 -> 1.15.5, confirming that this has the same replica identity behavior as 1.14.0.

PR #377, opened 2025-10-15
Sync streams: Support aliases

simolus3:
As reported [on Discord](https://discord.com/channels/1138230179878154300/1422138173907144724/1427962895425208382), using sync streams with a subquery having a table alias results in now rows being synced.

Sync rules avoid this issue by making aliases in parameter queries illegal and checking for that. We could do the same for sync streams, but that feels like an unecessary restriction. The actual bug here is that it's not consistently clear whether a `tableName: string` refers to the name in the source schema or the aliased name it has when parsing expressions. Because sync streams reference the aliased name when checking whether a given row can be a parameter input, this caused input tables to not be recognized.

This refactors `SqlTools` and related structures in `sync-rules` to always be aware of whether a name refers to an alias or the source table. When looking up values to create parameter lookups or bucket ids, we should always use the schema name. For output names, we use aliased names. By introducing the `AvailableTable` class for tables that have been added to a result set via `FROM`, we're guaranteed to always have both names available.

This could be used to lift the restriction of aliases for sync rules, but that feels unecessary given the eventual migration to sync streams.

### DIFF
--- a/packages/sync-rules/src/SqlParameterQuery.ts
+++ b/packages/sync-rules/src/SqlParameterQuery.ts
@@ -212,7 +212,8 @@ export class SqlParameterQuery {
    * The table name or alias, as referred to in the SQL query.
    * Not used directly outside the query.
    *
-   * Currently, this always matches sourceTable.name.
+   * Since aliases aren't allowed in parameter queries, this always matches sourceTable.name (checked by
+   * {@link fromSql}).
    */
   readonly table: AvailableTable;
 
@@ -308,7 +309,7 @@ export class SqlParameterQuery {
    */
   evaluateParameterRow(row: SqliteRow): EvaluatedParametersResult[] {
     const tables = {
-      [this.table.schemaName]: row
+      [this.table.nameInSchema]: row
     };
     try {
       const filterParameters = this.filter.filterRow(tables);
--- a/packages/sync-rules/src/sql_filters.ts
+++ b/packages/sync-rules/src/sql_filters.ts
@@ -46,6 +46,7 @@ import {
 } from './types.js';
 import { isJsonValue } from './utils.js';
 import { CompatibilityContext } from './compatibility.js';
+import { TablePattern } from './TablePattern.js';
 
 export const MATCH_CONST_FALSE: TrueIfParametersMatch = [];
 export const MATCH_CONST_TRUE: TrueIfParametersMatch = [{}];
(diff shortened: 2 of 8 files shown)


packages/sync-rules/src/sql_filters.ts
```diff
@@ -53,13 +53,64 @@ export const MATCH_CONST_TRUE: TrueIfParametersMatch = [{}];
...
 Object.freeze(MATCH_CONST_FALSE);
 
+/**
+ * A table that has been made available to a result set by being included in a `FROM`.
+ */
+export class AvailableTable {
+  /**
+   * The name of the table in the schema.
+   */
+  schemaName: string;
```
rkistner: The name here makes it appear like it's the name of the schema itself, e.g. `public`. Is there a different name we could use here? Maybe something like `originalName` or even just `name`?

packages/sync-rules/src/sql_filters.ts
```diff
@@ -53,13 +53,64 @@ export const MATCH_CONST_TRUE: TrueIfParametersMatch = [{}];
 Object.freeze(MATCH_CONST_TRUE);
 Object.freeze(MATCH_CONST_FALSE);
 
+/**
+ * A table that has been made available to a result set by being included in a `FROM`.
+ */
+export class AvailableTable {
```
rkistner, line 63: One thing that confused me here, is why does this not have/need the name of the schema?

It appears that it's because this is used to lookup references inside queries, and not to reference full source tables on its own - TablePattern is used for that. It would be good to document that explicitly.

packages/sync-rules/src/SqlParameterQuery.ts
```diff
@@ -214,7 +214,7 @@ export class SqlParameterQuery {
    *
    * Currently, this always matches sourceTable.name.
    */
-  readonly table: string;
+  readonly table: AvailableTable;
```
rkistner, line 218: Is the comment above still true?

simolus3: Yes, it's checked by `SqlParameterQuery.fromSql` just like it was before. I've expanded the comment to point that out.

packages/sync-rules/src/sql_filters.ts
simolus3: Good point, I've expanded the comment.

packages/sync-rules/src/sql_filters.ts
simolus3: I'd like to avoid the ambiguity of `name` given that that's what caused this issue for sync streams in the first place. How about `nameInSchema`?

PR #380, opened 2025-10-17
Reimplement dangerous request parameter check with visitor

simolus3:
The `usesAuthenticatedRequestParameters` and `usesUnauthenticatedRequestParameters` properties are currently computed whenever we're constructing or composing a clause in the `sync-rules` compiler.

While that's a fairly simple implementation, it doesn't scale well if we're trying to infer more information about used clauses in the future. In particular, I'm looking at a lint that would require knowing whether a query uses stream parameters. Adding that information (or more features in the future) to all clauses would complicate the implementation.

So, this refactors the check to use the visitor pattern:

- When a clause has child clauses, it exposes a function for traversing them.
- The request function and parameter table ref clauses have special identifiers to recognize them.
- A visitor traverses all clauses to check whether authenticated or unauthenticated request parameters are used.
- The `OR`combinator for parameter match clauses also has a special key so that we can apply the special behavior (it only counts if both sides are using authenticated parameters).

### DIFF
--- a/.changeset/chilled-lamps-hammer.md
+++ b/.changeset/chilled-lamps-hammer.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Remove `usesAuthenticatedRequestParameters` and `usesUnauthenticatedRequestParameters` in favor of `DetectRequestParameters` computing them on-demand.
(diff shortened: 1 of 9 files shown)


rkistner (approved):
Looks good - does seem like this will make it easier to add other checks in the future

PR #386, opened 2025-10-29
Test Postgres 18

rkistner:
Postgres 18 was [released](https://www.postgresql.org/about/news/postgresql-18-released-3142/) on September 25. It should work without any changes in powersync-service - this merely confirms it by adding it to our test matrix.

I modified one test slightly, since Postgres 18 now has more robust handling for dropped + re-created publications.


modules/module-postgres/test/src/wal_stream.test.ts
```diff
@@ -315,13 +315,74 @@ bucket_definitions:
...
+      const serverVersion = await context.connectionManager.getServerVersion();
+
       await context.loadActiveSyncRules();
-      await expect(async () => {
+
+      if (serverVersion!.compareMain('18.0.0') >= 0) {
         await context.replicateSnapshot();
-      }).rejects.toThrowError(MissingReplicationSlotError);
+        // No error expected in Postres 18
+        // TODO: introduce new test scenario for Postgres 18 that _does_ invalidate the replication slot.
```
stevensJourney, line 325: Just for my understanding. In Postgres 18, would the current replication slot automatically continue working after recreating the publication?

stevensJourney (approved):

modules/module-postgres/test/src/wal_stream.test.ts
rkistner: Yes, in Postgres 18 it continues working. You can even wait and re-create the publication at a later point, but in that case the stream may miss some data.

PR #387, opened 2025-10-30
[Postgres] Improve handling of "lost" replication slots

rkistner:
This adds a check for replication slots that are "lost" due to `max_slot_wal_keep_size` exceeded, to automatically re-create the slot if needed. There is an explicit `wal_status` field since Postgres 13+ that we now check, as well as some additional error message checks.

The slot health check is now also modified to only wait 2 minutes for the slot to become healthy, rather than 120 tries. This is relevant because each individual try can take 2 minutes in some scenarios, which can cause the overall check to only fail after 4 hours.

This does not yet solve the issue of this health check potentially causing high load on the source database. For that we should probably use an exponential back-off mechanism for the overall replication retries (separate PR).


stevensJourney (approved):
Looks good to me :)

PR #389, opened 2025-11-07
Add option to restrict datetime precision

simolus3:
This adds the `timestamp_max_precision` option for the `config` block when defining sync rules. Valid options are `seconds`, `milliseconds, `microseconds` and `nanoseconds`. Restricting precision is convenient in some cases, e.g. when other parts of an app can't handle sub-millisecond precision (something we noticed for some users migrating to `edition: 2` which pads postgres timestamps to six digits).

When a source database provides a time value, it now also provides:

- The inner precision of that time value (milliseconds for mongo, micros for mysql and postgres, nanoseconds for mssql). That doesn't mean that every value has that many fractional digits (postgres could e.g. emit `11:03:00.000000` as `11:03:00`), but it reflects the maximum precision supported.
- A default precision to use when no max precision is configured on sync rules. This matches the inner precision for mongo and postgres. For mysql, we use milliseconds as the default precision for backwards compatibility, while supporting microseconds as an opt-in.

Depending on compatibility options, a time value is then formatted with a precision of `min(timestamp_max_precision ?? source.default_precision, source.inner_precision)`. We also pad values to ensure we consistently use the same amount of digits.

### DIFF
--- a/modules/module-mssql/src/common/mssqls-to-sqlite.ts
+++ b/modules/module-mssql/src/common/mssqls-to-sqlite.ts
@@ -45,6 +45,7 @@ export function toSqliteInputRow(row: any, columns: sql.IColumnMetadata): Sqlite
           const date = row[key] as DateWithNanosecondsDelta;
           if (isNaN(date.getTime())) {
             result[key] = null;
+            break;
           }
 
           const originalFormat = date.toISOString();
--- a/modules/module-mssql/test/src/mssql-to-sqlite.test.ts
+++ b/modules/module-mssql/test/src/mssql-to-sqlite.test.ts
@@ -221,9 +221,11 @@ describe('MSSQL Data Types Tests', () => {
   test('Date types mappings', async () => {
     const beforeLSN = await getLatestLSN(connectionManager);
     const testDate = new Date('2023-03-06T15:47:00.123Z');
+    // This adds 0.4567 milliseconds to the JS date, see https://github.com/tediousjs/tedious/blob/0c256f186600d7230aec05553ebad209bed81acc/src/data-types/datetime2.ts#L74.
+    // Note that there's a typo in tedious there. When reading dates, the property is actually called nanosecondsDelta.
+    // This is only relevant when binding datetime values, so only in this test.
     Object.defineProperty(testDate, 'nanosecondDelta', {
       enumerable: false,
-      // Of course, nanosecondDelta is not measured in nanoseconds. It's actually in seconds.
       value: 0.0004567
     });
     await connectionManager.query(


modules/module-mssql/src/common/mssqls-to-sqlite.ts
```diff
@@ -32,14 +36,24 @@ export function toSqliteInputRow(row: any, columns: sql.IColumnMetadata): Sqlite
...
           break;
         case sql.TYPES.DateTime:
         case sql.TYPES.DateTime2:
         case sql.TYPES.SmallDateTime:
         case sql.TYPES.DateTimeOffset: // The offset is lost when the driver converts to Date. This needs to be handled in the sql query.
-          const date = row[key] as Date;
-          result[key] = isNaN(date.getTime()) ? null : date.toISOString();
+          const date = row[key] as DateWithNanosecondsDelta;
+          if (isNaN(date.getTime())) {
+            result[key] = null;
```
rkistner, line 47: The code still continues parsing below, which raise an error on `date.toISOString()`.

(Not sure in which cases this branch can actually happen, and whether we can add a unit test for this)

modules/module-mssql/test/src/mssql-to-sqlite.test.ts
```diff
@@ -214,7 +220,12 @@ describe('MSSQL Data Types Tests', () => {
 
   test('Date types mappings', async () => {
     const beforeLSN = await getLatestLSN(connectionManager);
-    const testDate = new Date('2023-03-06T15:47:00.000Z');
+    const testDate = new Date('2023-03-06T15:47:00.123Z');
+    Object.defineProperty(testDate, 'nanosecondDelta', {
+      enumerable: false,
+      // Of course, nanosecondDelta is not measured in nanoseconds. It's actually in seconds.
+      value: 0.0004567
+    });
```
rkistner, line 230: Where we parse dates, this is called `nanosecondsDelta` (plural), while this uses a singular `nanosecondDelta`. But the tests are passing... bug in tedious? https://github.com/tediousjs/tedious/pull/1084 (open since 2020...)

Does this mismatch only affect tests? Would be good to add a comment here.

simolus3: This only affects binding datetime values to parameters in prepared statements, so only this test. I've added a comment.

modules/module-mssql/src/common/mssqls-to-sqlite.ts
simolus3: cc @Rentacookie did you run into this with values emitted by MSSQL or is this a defensive check?

My understanding is that it would be impossible for tedious to emit invalid datetime values since they're assembled from the individual fields. The min/max values are reported correctly (and we have tests for those).

There's a `break` below this line now.

modules/module-mssql/src/common/mssqls-to-sqlite.ts
```diff
@@ -32,14 +36,24 @@ export function toSqliteInputRow(row: any, columns: sql.IColumnMetadata): Sqlite
...
-          result[key] = toISOTimeString(row[key] as Date);
+          result[key] = toISOTimeValue(row[key] as DateWithNanosecondsDelta);
           break;
         case sql.TYPES.DateTime:
         case sql.TYPES.DateTime2:
         case sql.TYPES.SmallDateTime:
         case sql.TYPES.DateTimeOffset: // The offset is lost when the driver converts to Date. This needs to be handled in the sql query.
-          const date = row[key] as Date;
-          result[key] = isNaN(date.getTime()) ? null : date.toISOString();
+          const date = row[key] as DateWithNanosecondsDelta;
```
Rentacookie, line 45: Wow this was a nice find. If only their docs mentioned that it was possible to retrieve the nanoseconds 🤯 
https://github.com/tediousjs/tedious/blob/master/src/value-parser.ts#L740

I'm glad we support this now though. It's currently listed as a limitation in our docs that we only support up to millisecond granularity.

modules/module-mssql/src/common/mssqls-to-sqlite.ts
Rentacookie: No I didn't, this was just a defensive check that I carried over from the MySQL type mappings.

modules/module-mssql/src/common/mssqls-to-sqlite.ts
simolus3: Thanks for mentioning the docs, I'll update them 👍

PR #394, opened 2025-11-11
Chore/general client analytics

EspressoTrip-v2:
Added general connection analytics query for Mongo and Postgres report storage.

Query will return a paginated list of connections which can be filtered by:
- user_id
- client_id
- date_range

Purpose: Ability to get more granular user/client specific connection data.

Bugfix:
report update UTC time for PostgresReportStorage

### DIFF
--- a/modules/module-mongodb-storage/src/utils/util.ts
+++ b/modules/module-mongodb-storage/src/utils/util.ts
@@ -148,7 +148,6 @@ export const createPaginatedConnectionQuery = async <T extends mongo.Document>(
    * */
   return {
     items,
-    total: 0,
     count,
     /** Setting the cursor to the connected at date of the last item in the list */
     cursor: count === limit ? items[items.length - 1].connected_at.toISOString() : undefined,
--- a/modules/module-postgres-storage/src/storage/PostgresReportStorage.ts
+++ b/modules/module-postgres-storage/src/storage/PostgresReportStorage.ts
@@ -309,7 +309,6 @@ export class PostgresReportStorage implements storage.ReportStorage {
      * */
     return {
       items,
-      total: 0,
       /** Setting the cursor to the connected at date of the last item in the list */
       cursor: count === limit ? items[items.length - 1].connected_at.toISOString() : undefined,
       count,
--- a/packages/types/src/reports.ts
+++ b/packages/types/src/reports.ts
@@ -104,7 +104,6 @@ export type ClientConnectionAnalyticsRequest = {
 };
 export type PaginatedResponse<T> = {
   items: T[];
-  total: number;
   count: number;
   cursor?: string;
   more: boolean;
--- a/modules/module-mongodb-storage/test/src/__snapshots__/connection-report-storage.test.ts.snap
+++ b/modules/module-mongodb-storage/test/src/__snapshots__/connection-report-storage.test.ts.snap
@@ -227,7 +227,6 @@ exports[`Report storage tests > Should show paginated response of all connection
     },
   ],
   "more": false,
-  "total": 0,
 }
 `;
 
@@ -262,7 +261,6 @@ exports[`Report storage tests > Should show paginated response of all connection
     },
   ],
   "more": true,
-  "total": 0,
 }
 `;
 
@@ -297,7 +295,6 @@ exports[`Report storage tests > Should show paginated response of all connection
     },
   ],
   "more": false,
-  "total": 0,
 }
 `;
 
@@ -332,7 +329,6 @@ exports[`Report storage tests > Should show paginated response of all connection
     },
   ],
   "more": true,
-  "total": 0,
 }
 `;
 
@@ -367,7 +363,6 @@ exports[`Report storage tests > Should show paginated response of all connection
     },
   ],
   "more": false,
-  "total": 0,
 }
 `;
 
@@ -390,7 +385,6 @@ exports[`Report storage tests > Should show paginated response of connections of
     },
   ],
   "more": false,
-  "total": 0,
 }
 `;
 
@@ -437,7 +431,6 @@ exports[`Report storage tests > Should show paginated response of connections ov
     },
   ],
   "more": false,
-  "total": 0,
 }
 `;
 
@@ -454,6 +447,5 @@ exports[`Report storage tests > Should show paginated response of connections ov
     },
   ],
   "more": false,
-  "total": 0,
 }
 `;
--- a/modules/module-postgres-storage/test/src/__snapshots__/connection-report-storage.test.ts.snap
+++ b/modules/module-postgres-storage/test/src/__snapshots__/connection-report-storage.test.ts.snap
@@ -228,7 +228,6 @@ exports[`Report storage tests > Should show paginated response of all connection
     },
   ],
   "more": false,
-  "total": 0,
 }
 `;
 
@@ -267,7 +266,6 @@ exports[`Report storage tests > Should show paginated response of all connection
     },
   ],
   "more": true,
-  "total": 0,
 }
 `;
 
@@ -306,7 +304,6 @@ exports[`Report storage tests > Should show paginated response of all connection
     },
   ],
   "more": false,
-  "total": 0,
 }
 `;
 
@@ -345,7 +342,6 @@ exports[`Report storage tests > Should show paginated response of all connection
     },
   ],
   "more": true,
-  "total": 0,
 }
 `;
 
@@ -384,7 +380,6 @@ exports[`Report storage tests > Should show paginated response of all connection
     },
   ],
   "more": false,
-  "total": 0,
 }
 `;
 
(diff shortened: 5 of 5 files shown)


.github/workflows/packages_release.yaml
```diff
@@ -22,6 +22,9 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v5
+        with:
+          # check out full history. development packages need this for changesets
+          fetch-depth: 0
```
rkistner, line 27: Is this a new issue? I haven't seen issues related to this in this repo yet.

modules/module-mongodb-storage/src/utils/util.ts
```diff
@@ -114,3 +114,50 @@ export function setSessionSnapshotTime(session: mongo.ClientSession, time: bson.
...
+  const createQuery = (cursor?: string) => {
+    if (!cursor) {
+      return query;
+    }
+    return {
+      $and: [
+        query,
+        {
+          /** We are using the connected at date as the cursor so that the functionality works the same on Postgres implementation
+           * The id field in postgres is an uuid, this will work similarly to the ObjectId in Mongodb
```
rkistner: Is the id comment relevant here? (the query doesn't appear to be related to the id)

modules/module-mongodb-storage/src/utils/util.ts
```diff
@@ -114,3 +114,50 @@ export function setSessionSnapshotTime(session: mongo.ClientSession, time: bson.
...
+          connected_at: {
+            $lt: new Date(cursor)
+          }
+        }
+      ]
+    } as mongo.Filter<T>;
+  };
+
+  /** cursor.count() deprecated */
+  const total = await collection.countDocuments(query);
```
rkistner: The count can be quite slow to calculate if there are a lot of results. Is the count required, or can we remove it?

modules/module-postgres-storage/src/storage/PostgresReportStorage.ts
```diff
@@ -106,8 +107,69 @@ export class PostgresReportStorage implements storage.ReportStorage {
...
+      }
+    }
+
+    /** Create a date range filter if it exists */
+    if (date_range) {
+      const { start, end } = date_range;
+      intermediateQuery +=
+        queryParams.length === 0
+          ? ` WHERE connected_at >= $1 AND connected_at <= $2`
+          : ` AND connected_at >= $${queryParams.length + 1} AND connected_at <= $${queryParams.length + 2}`;
```
rkistner, line 146: This appears to add multiple WHERE clauses if both date_range and client_id/user_id is specified - will this work? Same for `cursor` below.

.github/workflows/packages_release.yaml
EspressoTrip-v2: Yeah, you get `Error: Failed to find where HEAD diverged from origin/main. Does origin/main exist?` without the full history. This is for development package publishing...

modules/module-mongodb-storage/src/utils/util.ts
EspressoTrip-v2: No it isn't! Never removed the comment, done so.

modules/module-postgres-storage/src/storage/PostgresReportStorage.ts
EspressoTrip-v2: It builds up the query based on the query parameter legth and the adds a and or where, I have added some tests to validate functionality. The query looks like this:
`SELECT id, user_id, client_id, user_agent, sdk, jwt_exp::text AS jwt_exp, disconnected_at, connected_at::text AS connected_at, disconnected_at::text AS disconnected_at  FROM connection_report_events WHERE client_id = $1 AND user_id = $2 AND connected_at >= $3 AND connected_at <= $4 ORDER BY connected_at DESC`

packages/types/src/reports.ts
```diff
@@ -83,3 +94,18 @@ export type ClientConnectionReportRequest = {
...
+export type ClientConnectionAnalyticsRequest = {
+  client_id?: string;
+  user_id?: string;
+  cursor?: string;
+  limit?: number;
+  date_range?: ClientConnectionReportRequest;
+};
+export type PaginatedResponse<T> = {
+  items: T[];
+  total: number;
```
rkistner: Since the total is now just 0, can we remove the field rather?

modules/module-postgres-storage/src/storage/PostgresReportStorage.ts
```diff
@@ -225,6 +288,36 @@ export class PostgresReportStorage implements storage.ReportStorage {
...
+    const statement = this.clientsConnectionPagination(data);
+
+    const result = await this.db.queryRows<ClientConnectionResponse>(statement.mainQuery);
+    const items = result.map((item) => ({
+      ...item,
+      /** JS Date conversion to match document schema used for Mongo storage */
+      connected_at: new Date(item.connected_at),
+      disconnected_at: item.disconnected_at ? new Date(item.disconnected_at) : undefined,
+      jwt_exp: item.jwt_exp ? new Date(item.jwt_exp) : undefined
+      /** */
```
rkistner: Should this empty comment be here?

packages/service-core-tests/src/tests/register-report-tests.ts
```diff
@@ -133,4 +143,113 @@ export async function registerReportTests(factory: storage.ReportStorage) {
...
+      cursor: '<removed-for-snapshot>',
+      items: removeVolatileFields(initial.items)
+    };
+    expect(cleanedInitial).toMatchSnapshot();
+    const connections = await factory.getGeneralClientConnectionAnalytics({
+      cursor
+    });
+    const cleaned = {
+      ...connections,
+      items: removeVolatileFields(initial.items)
```
rkistner: This tests `initial` again - should probably test `connections.items` instead?

modules/module-postgres-storage/src/storage/PostgresReportStorage.ts
```diff
@@ -106,8 +107,70 @@ export class PostgresReportStorage implements storage.ReportStorage {
...
+  } {
+    const { cursor, limit, client_id, user_id, date_range } = params;
+    const queryLimit = limit || 100;
+    const queryParams: pg_wire.StatementParam[] = [];
+    let countQuery = `SELECT COUNT(*) AS total FROM connection_report_events`;
+    let query = `SELECT id, user_id, client_id, user_agent, sdk, jwt_exp::text AS jwt_exp, disconnected_at, connected_at::text AS connected_at, disconnected_at::text AS disconnected_at  FROM connection_report_events`;
+    let intermediateQuery = '';
+    /** Create a user_id/ client_id filter is they exist  */
+    if (client_id || user_id) {
+      if (client_id && !user_id) {
```
rkistner: For these filters, use `!= null` rather than just relying on the general truthiness, otherwise a blank string would be treated the same as null, which is probably not what we want here.

modules/module-mongodb-storage/src/storage/MongoReportStorage.ts
```diff
@@ -43,6 +44,27 @@ export class MongoReportStorage implements storage.ReportStorage {
...
 
+  async getGeneralClientConnectionAnalytics(
+    data: event_types.ClientConnectionAnalyticsRequest
+  ): Promise<event_types.PaginatedResponse<event_types.ClientConnection>> {
+    const { cursor, date_range } = data;
+    const limit = data?.limit || 100;
+
+    const connected_at = date_range ? { connected_at: { $lte: date_range.end, $gte: date_range.start } } : undefined;
+    const user_id = data.user_id ? { user_id: data.user_id } : undefined;
+    const client_id = data.client_id ? { client_id: data.client_id } : undefined;
```
rkistner: Same as on Postgres storage, use `!= null` for these.

modules/module-mongodb-storage/src/utils/util.ts
```diff
@@ -114,3 +114,44 @@ export function setSessionSnapshotTime(session: mongo.ClientSession, time: bson.
...
+  collection: mongo.Collection<T>,
+  limit: number,
+  cursor?: string
+) => {
+  const createQuery = (cursor?: string) => {
+    if (!cursor) {
+      return query;
+    }
+    const connected_at = query.connected_at
+      ? { $lt: new Date(cursor), $gte: query.connected_at.$gte }
```
rkistner, line 129: If you have many connected_at values with the same timestamp, this cursor implementation could skip some of them if they fall on page boundaries. I think this is okay for an initial implementation, but should look into something like using `(connected_at, _id)` as the cursor. Same applies for Postgres.

EspressoTrip-v2: Yeah I am aware of this, also thought initial implementation I would keep it this way. Assumed the chances of this would be on the lower side. But can use the id if it becomes an issue.

packages/service-core-tests/src/tests/register-report-tests.ts
EspressoTrip-v2: Thanks good catch 👍🏻

PR #396, opened 2025-11-12
Improve compacting & checksum pre-calculations

rkistner:
## Checksum pre-calculations

We occasionally see this error in the checksum pre-calculation step after the initial snapshot completed: `[Processing Sync Rules] [PSYNC_S2403] Query timed out while reading checksums`. The pre-calculation roughly worked as follows:
1. Fetch batch of 5000 buckets that need a pre-calculation.
2. Split into smaller batches of 200 at a time to do the checksum queries.
3. Calculate checksums over a max of 50 000 operations at a time.
4. Once all 5000 buckets have the checksums computed, persist it.

In certain scenarios under high load, especially if the documents themselves are large, a checksum query can timeout. And if _any_ of the smaller batches time out, the entire larger batch is restarted.

This improves the process in two ways:
1. Start by only fetching batches of 200 at a time.
2. Check the estimated size of the buckets. If the buckets are large, limit the calculation to a smaller batch to only contain around 50 000 operations at a time.

This is not a guaranteed fix, but should reduce the timeouts.

## Compacting buckets

#375 introduced a change to only compact buckets with changes. However, the change in logic had another side-effect: If buckets are modified while the compact process is running, they could be repeatedly re-compacted, resulting in the compact process never finishing.

This changes the logic to:
1. Only compact buckets with a minimum of 10 new operations since the last compact.
2. Avoid re-compacting the 100 most recently processed buckets.

This logic may also need further tweaks over time as we get more real-world usage data.

### DIFF
--- a/packages/service-core/src/storage/SyncRulesBucketStorage.ts
+++ b/packages/service-core/src/storage/SyncRulesBucketStorage.ts
@@ -218,7 +218,7 @@ export interface CompactOptions {
   moveBatchQueryLimit?: number;
 
   /**
-   * Minimof of 1, default of 10.
+   * Minimum of 1, default of 10.
    */
   minBucketChanges?: number;
 


packages/service-core/src/storage/SyncRulesBucketStorage.ts
```diff
@@ -217,6 +217,11 @@ export interface CompactOptions {
   /** Minimum of 1 */
   moveBatchQueryLimit?: number;
 
+  /**
+   * Minimof of 1, default of 10.
```
stevensJourney:
```suggestion
   * Minimum of 1, default of 10.
```

Seems like a typo :)

PR #413, opened 2025-11-24
feat: Sync Stream Metadata Logs

stevensJourney:
# Overview

This allows connecting clients to specify additional metadata to be included in the PowerSync service logs. Typical use cases might be to log the client's application version or other device metadata.

A new optional `app_metadata` field is now accepted for the HTTP and WebSocket `/sync/stream` endpoint. This can contain a `Record<string,string>` map of values.

The service will now log the metadata when a client's sync stream starts and stops. This metadata is not currently added to all logs, the request `rid` can be used to correlate metadata with other logs.

The service limits the amount of logged/stored data. We currently only log a maximum of 10 keys in the provided `app_metadata`. Each string is capped to 50 characters in length. These values might be configurable in future.

A JSON sample for a log

```json
{
  "container": "default",
  "level": "info",
  "message": "Sync stream started",
  "namespace": "anamespace",
  "node": "anode",
  "pod": "apod",
  "timestamp": "2025-11-24T12:13:45.647Z",
  "_entry": "{\"app_metadata\":{\"appVersion\":\"1.0.0\"},\"client_id\":\"2788c57b-1cf4-41a0-bc0f-a46c5\",\"client_params\":{},\"level\":\"info\",\"message\":\"Sync stream started\",\"rid\":\"s/019ac8-c6a0-7455-93c2-c41f83963d50\",\"timestamp\":\"2025-11-24T12:13:45.647Z\",\"user_agent\":\"powersync-js/1.43.0 powersync-web Chromium/142 macos\",\"user_id\":\"123\"}"
}

```

This also adds logging `client_params` to sync stream requests. This could be used to aid with debugging.

### DIFF
--- a/.changeset/old-buckets-lay.md
+++ b/.changeset/old-buckets-lay.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Fix compiling streams with multiple parameter matchers in a subquery.
--- a/packages/sync-rules/src/streams/from_sql.ts
+++ b/packages/sync-rules/src/streams/from_sql.ts
@@ -407,18 +407,63 @@ class SyncStreamCompiler {
       tools.error('This subquery must return exactly one column', query);
     }
 
-    const column = tools.compileRowValueExtractor(query.columns?.[0]?.expr);
-    if (isClauseError(column)) {
+    const columnOrError = tools.compileRowValueExtractor(query.columns?.[0]?.expr);
+    if (isClauseError(columnOrError)) {
       return;
     }
+    const column = columnOrError;
 
-    const where = tools.compileClause(query.where);
+    const where = this.whereClauseToFilters(tools, query.where);
+    const filter = where.toDisjunctiveNormalForm(tools);
 
+    function checkValidSubqueryFilter(
+      operator: FilterOperator
+    ): CompareRowValueWithStreamParameter | EvaluateSimpleCondition | null {
+      if (operator instanceof CompareRowValueWithStreamParameter || operator instanceof EvaluateSimpleCondition) {
+        return operator;
+      }
+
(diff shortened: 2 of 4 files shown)


packages/service-core/src/routes/endpoints/sync-stream.ts
```diff
@@ -75,6 +76,16 @@ export const syncStreamed = routeDefinition({
...
     const controller = new AbortController();
     const tracker = new sync.RequestTracker(metricsEngine);
+
+    const formattedAppMetadata = payload.params.app_metadata
+      ? limitParamsForLogging(payload.params.app_metadata)
+      : undefined;
+
+    logger.info('Sync stream started', {
+      app_metadata: formattedAppMetadata,
+      client_params: payload.params.parameters ? limitParamsForLogging(payload.params.parameters) : undefined
```
stevensJourney, line 86: The type of client parameters is currently `Record<string, any>` which can contain nested objects. The formatting here stringifies and truncates top-level items. Open to other suggestions here.

packages/service-core/src/util/param-logging.ts
```diff
@@ -0,0 +1,60 @@
...
+  maxKeyCount: number;
+  maxStringLength: number;
+};
+
+/**
+ * Default options for {@link limitParamsForLogging}.
+ */
+export const DEFAULT_PARAM_LOGGING_FORMAT_OPTIONS: ParamLoggingFormatOptions = {
+  maxKeyCount: 10,
+  maxStringLength: 50
```
rkistner: I'd recommend limits of 20 and 100 here, to give some breathing room, while still having a reasonable overall limit (around 2KB).

stevensJourney: 👍 Updated those values

PR #415, opened 2025-11-26
[MSSQL] Alpha: Support replication from SQL Server / Azure SQL Database

Rentacookie:
### Introduce SQL Server / Azure SQL Database as a replication datasource

This change set enables PowerSync to listen for changes on SQL Server Databases that have Change Data Capture (CDC) enabled.

This is achieved by polling the CDC tables for updates on a regular interval and writing the changes to the bucket storage.  For more details on how CDC works please see documentation:
[CDC for SQL Server](https://learn.microsoft.com/en-us/sql/relational-databases/track-changes/about-change-data-capture-sql-server?view=sql-server-ver17)
[CDC for Azure SQL Database](https://learn.microsoft.com/en-us/azure/azure-sql/database/change-data-capture-overview?view=azuresql)

#### Requirements: 

1.  CDC enabled SQL Server database
2. `_powersync_checkpoints` table to allow PowerSync to generate regular updates. Table should be CDC enabled.
3. CDC enabled on all the tables PowerSync needs to replicate.
4. Database user with the following:
    - Read / Write permissions on the PowerSync checkpoints table.
    - Read permissions on the replicated tables
    - `cdc_reader` role which grants access to CDC tables and functions
    - `VIEW SERVER PERFORMANCE STATE` (SQL Server only)
    - `VIEW DATABASE PERFORMANCE STATE` (SQL Server and Azure SQL)

See [init.sql](https://github.com/powersync-ja/powersync-service/pull/415/files#diff-6d8209253278a9a82de041dea5c47113b52834a2167165353f083bfd21035d6d) for examples.

#### Self Hosted Demo available here: 

- [ ] Update link when merged
https://github.com/powersync-ja/self-host-demo/pull/57

### DIFF
--- a/modules/module-mssql/src/replication/CDCStream.ts
+++ b/modules/module-mssql/src/replication/CDCStream.ts
@@ -13,13 +13,7 @@ import { getUuidReplicaIdentityBson, MetricsEngine, SourceEntityDescriptor, stor
 import { SqliteInputRow, SqliteRow, SqlSyncRules, TablePattern } from '@powersync/service-sync-rules';
 
 import { ReplicationMetric } from '@powersync/service-types';
-import {
-  BatchedSnapshotQuery,
-  IdSnapshotQuery,
-  MSSQLSnapshotQuery,
-  PrimaryKeyValue,
-  SimpleSnapshotQuery
-} from './MSSQLSnapshotQuery.js';
+import { BatchedSnapshotQuery, MSSQLSnapshotQuery, SimpleSnapshotQuery } from './MSSQLSnapshotQuery.js';
 import { MSSQLConnectionManager } from './MSSQLConnectionManager.js';
 import { getReplicationIdentityColumns, getTablesFromPattern, ResolvedTable } from '../utils/schema.js';
 import {
@@ -286,18 +280,14 @@ export class CDCStream {
     return resolvedTable;
   }
 
-  private async snapshotTableInTx(
-    batch: storage.BucketStorageBatch,
-    table: MSSQLSourceTable,
-    limited?: PrimaryKeyValue[]
-  ): Promise<void> {
+  private async snapshotTableInTx(batch: storage.BucketStorageBatch, table: MSSQLSourceTable): Promise<void> {
     // Note: We use the "Read Committed" isolation level here, not snapshot isolation.
     // The data may change during the transaction, but that is compensated for in the streaming
     // replication afterward.
     const transaction = await this.connections.createTransaction();
     await transaction.begin(sql.ISOLATION_LEVEL.READ_COMMITTED);
     try {
-      await this.snapshotTable(batch, transaction, table, limited);
+      await this.snapshotTable(batch, transaction, table);
 
       // Get the current LSN.
       // The data will only be consistent once incremental replication has passed that point.
@@ -325,8 +315,7 @@ export class CDCStream {
   private async snapshotTable(
     batch: storage.BucketStorageBatch,
     transaction: sql.Transaction,
-    table: MSSQLSourceTable,
-    limited?: PrimaryKeyValue[]
+    table: MSSQLSourceTable
   ) {
     let totalEstimatedCount = table.sourceTable.snapshotStatus?.totalEstimatedCount;
     let replicatedCount = table.sourceTable.snapshotStatus?.replicatedCount ?? 0;
@@ -335,9 +324,7 @@ export class CDCStream {
     // We do streaming on two levels:
     // 1. Coarse select from the entire table, stream rows 1 by one
     // 2. Fine level: Stream batches of rows with each fetch call
-    if (limited) {
-      query = new IdSnapshotQuery(transaction, table, limited);
-    } else if (BatchedSnapshotQuery.supports(table)) {
+    if (BatchedSnapshotQuery.supports(table)) {
       // Single primary key - we can use the primary key for chunking
       const orderByKey = table.sourceTable.replicaIdColumns[0];
       query = new BatchedSnapshotQuery(
@@ -404,30 +391,27 @@ export class CDCStream {
 
       // Important: flush before marking progress
       await batch.flush();
-      if (limited == null) {
-        let lastKey: Uint8Array | undefined;
-        if (query instanceof BatchedSnapshotQuery) {
-          lastKey = query.getLastKeySerialized();
-        }
-        if (lastCountTime < performance.now() - 10 * 60 * 1000) {
-          // Even though we're doing the snapshot inside a transaction, the transaction uses
-          // the default "Read Committed" isolation level. This means we can get new data
-          // within the transaction, so we re-estimate the count every 10 minutes when replicating
-          // large tables.
-          totalEstimatedCount = await this.estimatedCountNumber(table, transaction);
-          lastCountTime = performance.now();
-        }
-        const updatedSourceTable = await batch.updateTableProgress(table.sourceTable, {
-          lastKey: lastKey,
-          replicatedCount: replicatedCount,
-          totalEstimatedCount: totalEstimatedCount
-        });
-        this.tableCache.updateSourceTable(updatedSourceTable);
 
-        this.logger.info(`Replicating ${table.toQualifiedName()} ${table.sourceTable.formatSnapshotProgress()}`);
-      } else {
-        this.logger.info(`Replicating ${table.toQualifiedName()} ${replicatedCount}/${limited.length} for resnapshot`);
+      let lastKey: Uint8Array | undefined;
+      if (query instanceof BatchedSnapshotQuery) {
+        lastKey = query.getLastKeySerialized();
       }
+      if (lastCountTime < performance.now() - 10 * 60 * 1000) {
+        // Even though we're doing the snapshot inside a transaction, the transaction uses
+        // the default "Read Committed" isolation level. This means we can get new data
+        // within the transaction, so we re-estimate the count every 10 minutes when replicating
+        // large tables.
+        totalEstimatedCount = await this.estimatedCountNumber(table, transaction);
+        lastCountTime = performance.now();
+      }
+      const updatedSourceTable = await batch.updateTableProgress(table.sourceTable, {
+        lastKey: lastKey,
+        replicatedCount: replicatedCount,
+        totalEstimatedCount: totalEstimatedCount
+      });
+      this.tableCache.updateSourceTable(updatedSourceTable);
+
+      this.logger.info(`Replicating ${table.toQualifiedName()} ${table.sourceTable.formatSnapshotProgress()}`);
 
       if (this.abortSignal.aborted) {
         // We only abort after flushing
--- a/modules/module-mssql/src/replication/MSSQLSnapshotQuery.ts
+++ b/modules/module-mssql/src/replication/MSSQLSnapshotQuery.ts
@@ -1,5 +1,4 @@
 import { bson, ColumnDescriptor, SourceTable } from '@powersync/service-core';
-import { SqliteValue } from '@powersync/service-sync-rules';
 import { ServiceAssertionError } from '@powersync/lib-services-framework';
 import { MSSQLBaseType } from '../types/mssql-data-types.js';
 import sql from 'mssql';
@@ -14,8 +13,6 @@ export interface MSSQLSnapshotQuery {
    */
   next(): AsyncIterableIterator<sql.IColumnMetadata | sql.IRecordSet<any>>;
 }
-
-export type PrimaryKeyValue = Record<string, SqliteValue>;
 
 /**
  * Snapshot query using a plain SELECT * FROM table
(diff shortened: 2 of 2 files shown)


modules/module-mssql/.npmignore
```diff
@@ -0,0 +1 @@
+dev
```
Rentacookie: I've been wondering if we maybe need to use this more? I noticed files and folders appearing in our released npm packages which don't need to be there.

rkistner: Personally I like the inverse more: Specifying a whitelist of "files" in each package.json. We're not currently very consistent in this though - should perhaps add that to each package.

What are specific examples you've seen of files that are being included that shouldn't be?

packages/service-errors/src/codes.ts
```diff
@@ -297,6 +297,14 @@ export enum ErrorCode {
 
   // ## PSYNC_S2xxx: Service API
 
+  /**
+   *  Required updates in the Change Data Capture (CDC) are no longer available.
+   *
+   *  Possible causes:
+   *  * Older data has been cleaned up due to exceeding the retention period.
+   */
+  PSYNC_S1500 = 'PSYNC_S1500',
```
rkistner: Just a formatting issue, but this should be just above the `## PSYNC_S2xxx: Service API` comment.

modules/module-mssql/src/replication/MSSQLSnapshotQuery.ts
```diff
@@ -0,0 +1,230 @@
...
+    yield columnMetadata;
+
+    const keyDefinition = this.table.sourceTable.replicaIdColumns[0];
+    const ids = this.keys.map((record) => record[keyDefinition.name]);
+
+    const request = this.transaction.request();
+    const stream = request.toReadableStream();
+    request
+      .input('ids', ids)
+      .query(`SELECT * FROM ${this.table.toQualifiedName()} WHERE ${escapeIdentifier(keyDefinition.name)} = @ids`);
```
rkistner: This query is likely wrong (equality match on array of ids), but it actually seems this entire "resnapshot" code path is unused. Can remove the entire `IdSnapshotQuery` class, and the `limited` option on `snapshotTableInTx` and `snapshotTable`.

modules/module-mssql/src/api/MSSQLRouteAPIAdapter.ts
```diff
@@ -0,0 +1,283 @@
...
+          columns: Object.values(result.columns).map((column) => column.name),
+          rows: result.map((row) => {
+            return Object.values(row).map((value: any) => {
+              const sqlValue = sync_rules.applyValueContext(
+                sync_rules.toSyncRulesValue(row),
+                sync_rules.CompatibilityContext.FULL_BACKWARDS_COMPATIBILITY
+              );
+
+              if (typeof sqlValue == 'bigint') {
+                return Number(row);
```
rkistner: This seems wrong - should be `value` rather than `row`?

We could also consider removing support for this for SQL Server, same as MongoDB where we don't support it either. This was mainly intended for running debug queries on the source database directly from the PowerSync dashboard, but we have no plans to support that in the new dashboard yet - rather pushing users to use the db's own tooling for this (e.g. Supabase UI for Postgres).

modules/module-mssql/src/common/LSN.ts
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   *  Creates an LSN instance from the provided string representation. An error is thrown if the format is invalid.
+   *  @param stringLSN
+   */
+  static fromString(stringLSN: string): LSN {
+    if (!/^[0-9A-Fa-f]{8}:[0-9A-Fa-f]{8}:[0-9A-Fa-f]{4}$/.test(stringLSN)) {
+      throw new ReplicationAssertionError(
+        `Invalid LSN string. Expected format is [00000000:00000000:0000]. Got: ${stringLSN}`
+      );
+    }
```
rkistner: Should this check perhaps only allow lower-case forms? If mixed case is used, LSN comparison can break.

modules/module-mssql/src/common/mssqls-to-sqlite.ts
```diff
@@ -0,0 +1,151 @@
...
+      switch (columnMetadata.type) {
+        case sql.TYPES.BigInt:
+          // MSSQL returns BIGINT as a string to avoid precision loss
+          if (typeof row[key] === 'string') {
+            result[key] = BigInt(row[key]);
+          }
+          break;
+        case sql.TYPES.Bit:
+          // MSSQL returns BIT as boolean
+          result[key] = row[key] ? 1 : 0;
```
rkistner: Use 1n for true, and 0n for false - that is important to preserve these as integers rather than floating-point values throughout sync rules. Can use `SQLITE_TRUE` / `SQLITE_FALSE` constants for this.

modules/module-mssql/src/common/mssqls-to-sqlite.ts
```diff
@@ -0,0 +1,151 @@
...
+        case sql.TYPES.DateTime2:
+        case sql.TYPES.SmallDateTime:
+        case sql.TYPES.DateTimeOffset: // The offset is lost when the driver converts to Date. This needs to be handled in the sql query.
+          const date = row[key] as Date;
+          result[key] = isNaN(date.getTime()) ? null : date.toISOString();
+          break;
+        case sql.TYPES.Binary:
+        case sql.TYPES.VarBinary:
+        case sql.TYPES.Image:
+          result[key] = new Uint8Array(Object.values(row[key]));
```
rkistner: What is the source type here? This conversion appears potentially quite inefficient.

modules/module-mssql/src/replication/CDCPoller.ts
```diff
@@ -0,0 +1,247 @@
...
+    }
+  }
+
+  public async replicateUntilStopped(): Promise<void> {
+    this.logger.info(`CDC polling started with interval of ${this.pollingIntervalMs}ms...`);
+    this.logger.info(`Polling a maximum of ${this.pollingBatchSize} transactions per polling cycle.`);
+    while (!this.isStopped) {
+      // Don't poll if already polling (concurrency guard)
+      if (this.isPolling) {
+        await timers.setTimeout(this.pollingIntervalMs);
```
rkistner: It may be better to throw a hard error in this case - there should never be two concurrent polling loops running.

rkistner (commented):
Still need to review the replication implementation, but here are some initial comments.

modules/module-mssql/src/common/MSSQLSourceTable.ts
```diff
@@ -0,0 +1,54 @@
...
+  get cdcSchema() {
+    return this.options.captureInstance.schema;
+  }
+
+  get CTTable() {
+    return `${this.cdcSchema}.${this.captureInstance}_CT`;
+  }
+
+  get allChangesFunction() {
+    return `${this.cdcSchema}.fn_cdc_get_all_changes_${this.captureInstance}`;
```
rkistner: Is there any risk of "SQL injection" here and in the other similar functions? (I haven't checked where these values come from, and whether or not they can always be assumed to be safe)

modules/module-mssql/src/utils/mssql.ts
```diff
@@ -0,0 +1,420 @@
...
+  const rawLSN: Buffer = result[0].max_lsn;
+  return LSN.fromBinary(rawLSN);
+}
+
+/**
+ *  Escapes an identifier for use in MSSQL queries.
+ *  @param identifier
+ */
+export function escapeIdentifier(identifier: string): string {
+  return `[${identifier}]`;
```
rkistner: There's no actual escaping here - opens up the possibility for SQL injection.

modules/module-mssql/src/replication/CDCStream.ts
```diff
@@ -0,0 +1,691 @@
...
+   *  This requires that the MSSQL DB user has the VIEW DATABASE PERFORMANCE STATE permission.
+   * @param table
+   * @param transaction
+   */
+  async estimatedCountNumber(table: MSSQLSourceTable, transaction?: sql.Transaction): Promise<number> {
+    const request = transaction ? transaction.request() : await this.connections.createRequest();
+    const { recordset: result } = await request.query(
+      `SELECT SUM(row_count) AS total_rows
+       FROM sys.dm_db_partition_stats
+       WHERE object_id = OBJECT_ID('${table.toQualifiedName()}')
```
rkistner: Even if escaping is properly applied to the name for the purpose of querying, the same name may not be safe for use inside a string. Rather pass in as a query parameter.

modules/module-mssql/src/utils/mssql.ts
```diff
@@ -0,0 +1,420 @@
...
+
+  return errors;
+}
+
+export async function ensurePowerSyncCheckpointsTable(connectionManager: MSSQLConnectionManager): Promise<string[]> {
+  const errors: string[] = [];
+  try {
+    // check if the dbo_powersync_checkpoints table exists
+    const { recordset: checkpointsResult } = await connectionManager.query(`
+    SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = '${connectionManager.schema}' AND TABLE_NAME = '${POWERSYNC_CHECKPOINTS_TABLE}';
```
rkistner: Use query parameters for the schema and table names.

modules/module-mssql/src/utils/mssql.ts
```diff
@@ -0,0 +1,420 @@
...
+      return errors;
+    }
+  } catch (error) {
+    errors.push(`Failed ensure ${POWERSYNC_CHECKPOINTS_TABLE} table is correctly configured: ${error}`);
+  }
+
+  // Try to create the table
+  try {
+    await connectionManager.query(`
+  CREATE TABLE ${connectionManager.schema}.${POWERSYNC_CHECKPOINTS_TABLE} (
```
rkistner: Schema should be escaped. There appears to be many cases like this - please check throughout the codebase.

modules/module-mssql/src/utils/mssql.ts
```diff
@@ -0,0 +1,420 @@
...
+export async function isTableEnabledForCDC(options: IsTableEnabledForCDCOptions): Promise<boolean> {
+  const { connectionManager, table, schema } = options;
+
+  const { recordset: checkResult } = await connectionManager.query(
+    `
+      SELECT 1 FROM cdc.change_tables ct
+         JOIN sys.tables    AS tbl ON tbl.object_id = ct.source_object_id
+         JOIN sys.schemas   AS sch ON sch.schema_id = tbl.schema_id
+      WHERE sch.name = '${schema}'
+        AND tbl.name = '${table}'
```
rkistner: These should be query parameters.

modules/module-mssql/src/utils/mssql.ts
```diff
@@ -0,0 +1,420 @@
...
+      );
+      return false;
+    }
+  }
+  return true;
+}
+
+export async function getMinLSN(connectionManager: MSSQLConnectionManager, captureInstance: string): Promise<LSN> {
+  const { recordset: result } = await connectionManager.query(
+    `SELECT sys.fn_cdc_get_min_lsn('${captureInstance}') AS min_lsn`
```
rkistner: This should be a query parameter.

modules/module-mssql/src/utils/schema.ts
```diff
@@ -0,0 +1,172 @@
...
+        col.name AS [name],
+        typ.name AS [type],
+        typ.system_type_id AS type_id,
+        typ.user_type_id AS user_type_id
+      FROM sys.columns AS col
+        JOIN sys.tables AS tbl ON tbl.object_id = col.object_id
+        JOIN sys.schemas AS sch ON sch.schema_id = tbl.schema_id
+        JOIN sys.types AS typ ON typ.user_type_id = col.user_type_id
+      WHERE sch.name = '${schema}'
+        AND tbl.name = '${tableName}'
```
rkistner: These should be query parameters. Also check other queries here - there are a couple of cases like this.

rkistner (commented):
Overall architecture looks good. My main comments here are around SQL injection possibilities, primarily with the table and schema names.

modules/module-mssql/.npmignore
Rentacookie: I saw that the tests are being included, not sure if they need to be though?

rkistner: Tests should be excluded ideally.

modules/module-mssql/src/utils/schema.ts
Rentacookie: Thanks - I switched to using parameters later on, but missed updating the older queries 😓

modules/module-mssql/src/utils/mssql.ts
Rentacookie: In SQL Server square brackets are used for escaping identifiers, unless I missed something.

rkistner: This just wraps the name, doesn't escape characters inside it. For example, what happens when the table name is `mytable] -- some other sql here`?

modules/module-mssql/src/common/MSSQLSourceTable.ts
Rentacookie: I think the risk is pretty low in this case, but the schema should still be escaped. The capture instance is a name generated by the database so should be safe.

modules/module-mssql/src/common/mssqls-to-sqlite.ts
Rentacookie: Good point. Binary mssql types are received as a Buffer. Thinking about it, do we even need to do the conversion in that case?

modules/module-mssql/src/common/mssqls-to-sqlite.ts
Rentacookie: Oh nice - I wasn't aware of those constants :)

modules/module-mssql/src/common/LSN.ts
Rentacookie: I see what you mean I'll rather enforce the uppercase format in that case. In the few places that LSNs appear as hex strings in MSSQL the upper case form is used. Usually they are stored as binary and we do the conversion to an uppercase hex string anyway..

modules/module-mssql/src/api/MSSQLRouteAPIAdapter.ts
Rentacookie: Ah makes sense - I'll remove the functionality then.

packages/service-errors/src/codes.ts
Rentacookie: 👌

modules/module-mssql/src/utils/mssql.ts
Rentacookie: Ah I see now what you mean - I'll fix that 🤦

modules/module-mssql/src/api/MSSQLRouteAPIAdapter.ts
```diff
@@ -0,0 +1,244 @@
...
+          WHEN typ.name IN ('float', 'real')
+            AND col.precision > 0
+          THEN typ.name + '(' + CAST(col.precision AS VARCHAR) + ')'
+          ELSE typ.name
+        END AS formatted_type
+      FROM sys.tables AS tbl
+        JOIN sys.schemas AS sch ON sch.schema_id = tbl.schema_id
+        JOIN sys.columns AS col ON col.object_id = tbl.object_id
+        JOIN sys.types AS typ ON typ.user_type_id = col.user_type_id
+      WHERE sch.name = '${this.connectionManager.schema}'
```
rkistner: Another case that should use query parameters.

modules/module-mssql/src/common/mssqls-to-sqlite.ts
rkistner: In that case, `new Uint8Array(buffer)` is probably the best option.

I'd say we do still need the conversion. Technically Buffer inherits Uint8Array, but there are subtle differences, so I'd rather just convert to a normal Uint8Array so that we have consistent types everywhere.

.changeset/thin-snails-compete.md
```diff
@@ -0,0 +1,15 @@
+---
+'@powersync/service-core': minor
+'@powersync/service-module-mssql': minor
+'@powersync/service-module-postgres-storage': patch
+'@powersync/service-module-mongodb-storage': patch
+'@powersync/service-module-postgres': patch
+'@powersync/service-errors': patch
+'@powersync/service-module-mysql': patch
+'@powersync/service-image': patch
```
stevensJourney: Should `@powersync/service-image` also be a minor bump? I believe that should have the same version as `@powersync/service-core`

modules/module-mssql/src/replication/MSSQLSnapshotQuery.ts
Rentacookie: Makes sense thank you

Issue #410, opened 2025-11-20
Unused modules take up memory


rkistner:
From some basic testing, when running an instance using MongoDB for both the source database and bucket storage, disabling the Postgres and MySQL modules could reduce memory usage by around 30MB. It may also reduce startup time.

In my tests I specifically disabled the registration of these 3 modules, as well as their associated imports:
https://github.com/powersync-ja/powersync-service/blob/57f766092a00cfd25ecc9cbd9b91bd54e9bb773d/service/src/entry.ts#L24-L26

See these heap snapshots - note the differences in (compiled code) and (string).

And it's not a very scientific measurement, but heap usage for the replication process differed by around 30MB on my machine, which is a little more than the difference measured in (compiled code) and (string).

All modules enabled:

<img width="3836" height="1631" alt="Image" src="https://github.com/user-attachments/assets/e64c9e32-c063-4eb5-b28d-586cffd24780" />

Postgres and MySQL disabled:
<img width="3836" height="1631" alt="Image" src="https://github.com/user-attachments/assets/6d1bb992-8fa7-4c91-851a-a7a91102e7b2" />

Proposal: Make the modules load their dependencies dynamically when needed, instead of importing everything upfront.
Alternatively, instead of doing this for all dependencies, focus on specific bigger ones, such as `node-sql-parser`, and perhaps the database drivers. Either way, this may require refactoring some synchronous methods to async.

PR #418, opened 2025-11-27
Dynamically load connection modules for reduced memory usage

najamansari:
Fix for #410.

Module loading is moved to `service/src/util/module-loader.ts`. This change has
been applied to all three runners (unified, API, sync).

Tests for the module loader have been added in `service/test/src/util/module-loader.test.ts`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Login to Docker Hub
+        if: github.event_name != 'pull_request'
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
@@ -48,6 +49,7 @@ jobs:
       - uses: actions/checkout@v5
 
       - name: Login to Docker Hub
+        if: github.event_name != 'pull_request'
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
@@ -95,6 +97,7 @@ jobs:
       - uses: actions/checkout@v5
 
       - name: Login to Docker Hub
+        if: github.event_name != 'pull_request'
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
@@ -171,6 +174,7 @@ jobs:
       - uses: actions/checkout@v5
 
       - name: Login to Docker Hub
+        if: github.event_name != 'pull_request'
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
@@ -244,6 +248,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Login to Docker Hub
+        if: github.event_name != 'pull_request'
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}


service/src/util/module-loader.ts
```diff
@@ -0,0 +1,56 @@
...
+    ...new Set(config.connections?.map((connection) => connection.type) || []),
+    `${config.storage.type}-storage` // Using template literal is clear
+  ];
+
+  // 2. Map connection types to their module loading promises
+  const modulePromises = requiredConnections.map(async (connectionType) => {
+    // Exclude 'mongo' connections explicitly early
+    if (connectionType.startsWith('mongo')) {
+      return null; // Return null for filtering later
+    }
```
rkistner: Why are MongoDB modules excluded, and not included in the module map?

service/src/util/module-loader.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    try {
+      // Dynamically import and instantiate the class
+      const ModuleClass = await modulePromise();
+      return new ModuleClass();
+    } catch (error) {
+      // Log an error if the dynamic import fails (e.g., module not installed)
+      logger.error(`Failed to load module for ${connectionType}:`, error);
+      return null;
+    }
```
rkistner: Is there any scenario where the process could feasibly continue if module loading fails? If not, I'd recommend just raising the error (skip the try-catch), which should bubble up and terminate the process.

The only exception I can think of is if this could allow better error messages, e.g. to highlight additional config issues.

service/src/util/module-loader.ts
```diff
@@ -0,0 +1,56 @@
...
+const ModuleMap: DynamicModuleMap = {
+  mysql: () => import('@powersync/service-module-mysql').then((module) => module.MySQLModule),
+  postgresql: () => import('@powersync/service-module-postgres').then((module) => module.PostgresModule),
+  'postgresql-storage': () =>
+    import('@powersync/service-module-postgres-storage').then((module) => module.PostgresStorageModule)
+};
+
+/**
+ * Utility function to dynamically load and instantiate modules.
+ * This function can optionally be moved to its own file (e.g., module-loader.ts)
```
rkistner: This is already in module-loader.ts?

service/src/entry.ts
rkistner: Can these imports also be removed?

service/src/util/module-loader.ts
```diff
@@ -0,0 +1,56 @@
...
+interface DynamicModuleMap {
+  [key: string]: () => Promise<any>;
+}
+
+const ModuleMap: DynamicModuleMap = {
+  mysql: () => import('@powersync/service-module-mysql').then((module) => module.MySQLModule),
+  postgresql: () => import('@powersync/service-module-postgres').then((module) => module.PostgresModule),
+  'postgresql-storage': () =>
+    import('@powersync/service-module-postgres-storage').then((module) => module.PostgresStorageModule)
+};
```
rkistner: Could it help to split this into separate maps for replication modules and storage modules? That may allow better types here, and avoid the need to add `-storage` below.

It may help further to construct the module directly in the map functions.

najamansari:
@rkistner Thanks for taking a look! Yes there's definitely a little improvement when running the server against MongoDB in my fix vs the main branch baseline.

On my branch:

|         | 10 s  | 60 s  | 600 s |
| ------- | ----- | ----- | ----- |
| Total   | 83642 | 83980 | 84730 |
| Code    | 16246 | 16562 | 17219 |
| Strings | 16377 | 16377 | 16398 |

On the baseline (main branch):

|         | 10 s  | 60 s  | 600 s |
| ------- | ----- | ----- | ----- |
| Total   | 95833 | 96108 | 96923 |
| Code    | 18050 | 18309 | 19028 |
| Strings | 23952 | 23952 | 23973 |

During both sets of tests I spun up fresh DB/BE/UI containers via Docker compose, started powersync-server via entry.ts in unified runner mode (default) and took snapshots at the 10/60/600 second marks. During this time, I opened up the todo app UI in two different browser windows and made small operations in each (creating a list, adding tasks to a list).

service/src/util/module-loader.ts
najamansari: AIUI we always need MongoDB since powersync requires it for its own state. That is why I excluded it from the optional loader.

service/src/util/module-loader.ts
najamansari: ack, logging the error here and rethrowing

service/src/util/module-loader.ts
najamansari: ack, fixed

service/src/entry.ts
najamansari: ack

service/src/util/module-loader.ts
najamansari: Thanks, made some refactorings, could you perhaps take another look? The type remains the same but the code is cleaner.

service/src/util/module-loader.ts
rkistner: MongoDB is also optional - the service can use either postgres-storage and mongodb-storage for the internal state.

najamansari: Understood, fixed accordingly.

rkistner (dismissed):

Rentacookie (dismissed):

Rentacookie (approved):

Issue #419, opened 2025-12-02
[Postgres] Parsing large arrays fail with RangeError


rkistner:
When replicating a large Postgres array, replication can fail with `RangeError: Maximum call stack size exceeded`. In this case it's not an issue with recursion, with rather with `String.fromCharCode(...charCodes)` with too many characters. According to an old [stackoverflow answer](https://stackoverflow.com/a/22747272/214837), this can start happening around 65k characters.

This error is not related to the client or sync rules - if the source row contains a column like this at all, replication can fail.

Example replication error:

```
RangeError: Maximum call stack size exceeded
    at StructureParser.quotedString (file:///app/node_modules/.pnpm/@powersync+service-jpgwire@0.21.6/node_modules/@powersync/service-jpgwire/dist/structure_parser.js:91:23)
    at StructureParser.parseArrayInner (file:///app/node_modules/.pnpm/@powersync+service-jpgwire@0.21.6/node_modules/@powersync/service-jpgwire/dist/structure_parser.js:132:49)
    at StructureParser.parseArray (file:///app/node_modules/.pnpm/@powersync+service-jpgwire@0.21.6/node_modules/@powersync/service-jpgwire/dist/structure_parser.js:113:28)
    at PgType._decodeArray (file:///app/node_modules/.pnpm/@powersync+service-jpgwire@0.21.6/node_modules/@powersync/service-jpgwire/dist/pgwire_types.js:152:42)
    at PgType.decode (file:///app/node_modules/.pnpm/@powersync+service-jpgwire@0.21.6/node_modules/@powersync/service-jpgwire/dist/pgwire_types.js:142:25)
    at CustomTypeRegistry.decodeDatabaseValue (file:///app/node_modules/.pnpm/@powersync+service-module-postgres@0.16.12/node_modules/@powersync/service-module-postgres/dist/types/registry.js:189:34)
    at PostgresTypeResolver.decodeTuple (file:///app/node_modules/.pnpm/@powersync+service-module-postgres@0.16.12/node_modules/@powersync/service-module-postgres/dist/types/resolver.js:184:52)
    at PostgresTypeResolver.constructAfterRecord (file:///app/node_modules/.pnpm/@powersync+service-module-postgres@0.16.12/node_modules/@powersync/service-module-postgres/dist/types/resolver.js:155:29)
[...]

PR #420, opened 2025-12-02
Fix parsing large escaped strings from Postgres structures

simolus3:
When I wrote `StructureParser.quotedString`, my idea was to use an array of char codes followed by `String.fromCharCode(...charCodes)` as a replacement for a string buffer, assuming that that should be the most efficient way to compose strings char-by-char.

That assumption was completely wrong though:

- The array of char codes is not a typed array, which is inefficient.
- Calling `String.fromCharCode` with a large number of parameters can lead to stack overflow errors.

As it turns out, the best way to implement a string buffer in JavaScript is to [just use a string](https://github.com/dart-lang/sdk/blob/053f0ed0057cbb4e935d6532e63bfcf943ac8dc3/sdk/lib/_internal/js_runtime/lib/core_patch.dart#L537-L539). So, this refactors `quotedString` to accumulate characters into that string.

I've done some benchmarking with a single-element array containing a large escaped string. Parsing a 70MB string takes around 2.5 seconds which is certainly not great, but might be as good as it gets (profiling with `--profile` only gives me hot functions, I'm not sure if there's a way to diagnose any bad lines in particular).
Considering we have a 15MB row limit anyway, reasonable rows should decode reasonably quickly.

Closes https://github.com/powersync-ja/powersync-service/issues/419.

### DIFF
--- a/packages/jpgwire/src/structure_parser.ts
+++ b/packages/jpgwire/src/structure_parser.ts
@@ -74,18 +74,22 @@ export class StructureParser {
 
     const start = this.offset;
     let buffer = '';
-    let previousWasBackslash = false;
 
     while (true) {
-      const char = this.currentCharCode();
+      const startPosition = this.offset;
 
-      if (previousWasBackslash) {
-        if (char != CHAR_CODE_DOUBLE_QUOTE && char != CHAR_CODE_BACKSLASH) {
-          this.error('Expected escaped double quote or escaped backslash');
-        }
-        buffer += this.currentChar();
-        previousWasBackslash = false;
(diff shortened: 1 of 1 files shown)


rkistner (approved):
Happy with the fix, but I'm wondering if it could be made even more efficient by building the string in chunks at a time, instead of 1 character at a time? For example, (1) find the next backslash or quote, (2) append the entire chunk up to that point, (3) repeat.

I haven't done any performance testing though, so not sure how big a difference this would make.

rkistner (approved):

PR #425, opened 2025-12-04
[Incremental Reprocessing] Concurrent storage write batch support

rkistner:
This modifies the APIs and implementation of `BucketStorageBatch` to allow using multiple batches concurrently without introducing consistency issues. The first goal of this is to allow reading the replication stream while snapshotting - see #426 for a Postgres implementation.

The actual writes to the database still rely on exclusive database locks to a large extent, so there isn't true concurrency yet. But it does allow running a replication stream concurrently with replication snapshots without breaking.

Specifically, this changes:
1. Rely less on local variables in the batch, using the database state instead.
2. Remove the distinction between `commit()` and `keepalive()`. The only remaining difference is that keepalive is essentially `commit({allowEmptyCommit: true})`.
3. `snapshot_done` is now explicitly set after completing a snapshot, rather than setting it automatically on commit. No commit will go through until snapshot_done is set.
4. `current_data` deletes are now soft-deletes (storing an empty document/row instead of deleting the document/row), converted into hard deletes on the next commit. We need this to consistently handle snapshots while streaming. (see Storage Changes below).
5. Fix concurrency issues for flushing and committing batches, especially in Postgres storage (previously not an issue since we wouldn't ever do this concurrently).
6. Change `SourceTable.id` to be strongly typed, specifically checking for `string` in Postgres storage and `ObjectId` in MongoDB storage. This unfortunately had a big knock-on effect in tests, including now having different checksums in postgres storage vs mongodb storage.

## Storage Changes

The storage changes uses the storage version mechanism introduced in https://github.com/powersync-ja/powersync-service/pull/487. This is the first change that actually uses a difference in storage collections/tables, so it's a bit of an experiment in that regard.

1. This introduces an unstable storage version 3, which uses a new `v3_current_data` collection/table.
2. This storage version is not enabled yet - only used in tests. We can continue iterating on this storage version until we enable this by default.
3. The `v3_current_data` collection/table is initialized when storage version 3 is first used, instead of using the migration system. It is not removed when downgrading to older versions.
4. `v3_current_data` has a new `pending_delete` field, to support soft deletes.

This means the new soft delete functionality with the improved concurrency support is only available when using storage version 3, and existing instances would be unaffected by this change.

The MongoDB storage changes are fairly clean:
1. The same queries can be used on the two collections interchangeably for the most part.
2. A new `VersionedPowerSyncDatabase` class is added with additional checks, to ensure that we never use the wrong collection by accident.
3. In a small number of places, the behavior differs based on the storage config.

In Postgres storage this is more complicated, since the table name and field names are hardcoded in the SQL query, and it doesn't have the same central class through which all the database calls go like the MongoDB storage. I refactored this to move all `current_data` access through a central `PostgresCurrentDataStore` class, to keep this managable.

## Other changes

Other more minor changes:
1. Reduce migration logs in tests.
2. Rely less on a split between initial snapshot and streaming replication in Postgres tests.
3. Add a `cause` to `ReplicationAbortedError`.
4. Runs most storage tests on all storage versions.

### DIFF
--- a/packages/sync-rules/src/sync_plan/plan.ts
+++ b/packages/sync-rules/src/sync_plan/plan.ts
@@ -47,9 +47,9 @@ export interface TableProcessor {
   /**
    * All of these expressions exclusively depend on the {@link sourceTable}.
    *
-   * All of the filters must evaluate to a "true-ish" value for the row to be processed.
+   * All of the filters must evaluate to a truthy (according to SQL definitions) value for the row to be processed.
    */
-  filters: SqlExpression<ColumnSqlParameterValue>[];
+  filters: SqlExpression<TableProcessorData>[];
   /**
    * How to partition rows.
    *
@@ -75,10 +75,11 @@ export interface TableProcessor {
  *   - Two table-valued partition keys evaluating two multiple rows leads to a quadratic amount of buckets being
  *     generated.
  */
-export type TableProcessorTableValuedFunction = Omit<
-  EvaluateTableValuedFunction<ColumnSqlParameterValue>,
-  'type' | 'outputs'
->;
+export interface TableProcessorTableValuedFunction {
+  functionName: string;
+  functionInputs: SqlExpression<ColumnSqlParameterValue>[];
+  filters: SqlExpression<ColumnSqlParameterValue>[];
+}
 
 export interface TableProcessorTableValuedFunctionOutput {
   /**
@@ -92,12 +93,20 @@ export interface TableProcessorTableValuedFunctionOutput {
 }
 
 /**
+ * External data that can be used in expressions for {@link TableProcessor}s.
+ *
+ * This allows references to the row in the table being processed, and references to outputs of joined
+ * {@link TableProcessor#tableValuedFunctions}.
+ */
+export type TableProcessorData = ColumnSqlParameterValue | TableProcessorTableValuedFunctionOutput;
+
+/**
  * A scalar partition key evaluates to a single value in the source or parameter row. For instance, a stream definition
  * like `SELECT * FROM users WHERE id = auth.user_id()` would generate a scalar key to partition users by the `id`
  * column.
  */
 export interface PartitionKey {
-  expr: SqlExpression<ColumnSqlParameterValue | TableProcessorTableValuedFunctionOutput>;
+  expr: SqlExpression<TableProcessorData>;
 }
 
 /**
@@ -121,7 +130,7 @@ export interface StreamDataSource extends TableProcessor {
   columns: ColumnSource[];
 }
 
-export type ColumnSource = 'star' | { expr: SqlExpression<ColumnSqlParameterValue>; alias: string };
+export type ColumnSource = 'star' | { expr: SqlExpression<TableProcessorData>; alias: string };
 
 /**
  * A mapping describing how {@link StreamDataSource}s are combined into buckets.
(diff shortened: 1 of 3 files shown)


modules/module-mongodb-storage/src/storage/implementation/MongoBucketBatch.ts
```diff
@@ -552,7 +547,7 @@ export class MongoBucketBatch
     // 5. TOAST: Update current data and bucket list.
     if (afterId) {
       // Insert or update
-      const after_key: SourceKey = { g: this.group_id, t: sourceTable.id, k: afterId };
+      const after_key: SourceKey = { g: this.group_id, t: sourceTable.id as bson.ObjectId, k: afterId };
```
stevensJourney: Shouldn't this type cast be using `mongoTableId`?

modules/module-mongodb-storage/src/storage/implementation/MongoBucketBatch.ts
```diff
@@ -752,23 +693,165 @@ export class MongoBucketBatch
...
+          }
+        }
+      ]
+    };
+
+    const new_last_checkpoint = {
+      $cond: [
+        can_checkpoint,
+        {
+          $max: ['$last_checkpoint', { $literal: this.persisted_op }, { $toLong: '$keepalive_op' }]
```
stevensJourney: I'm not sure about the specifics here, but I noticed the Postgres implementation coerces the, potentially null `persisted_op` to zero. Should we do that here also?

rkistner: Since this takes the `$max` of these values, any null values would be smaller than others and be ignored. I added an explicit 0 as a fallback here now to match the Postgres behavior, and also simplified the Postgres query.

While it _should_ never reach the case where the value ends up as null, there are no explicit checks for that in the code, so it is safer to have the 0 fallback.

libs/lib-services/src/logger/logger-index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './Logger.js';
 export { Logger } from 'winston';
+export { createLogger, format, transports } from 'winston';
```
simolus3: nit:

```suggestion
export { createLogger, format, Logger, transports } from 'winston';
```

modules/module-postgres-storage/src/storage/PostgresBucketStorageFactory.ts
```diff
@@ -195,6 +214,34 @@ export class PostgresBucketStorageFactory extends storage.BucketStorageFactory {
     });
   }
 
+  /**
+   * Lazy-initializes storage-version-specific structures, if needed.
+   */
+  private async initializeStorageVersion(storageConfig: storage.StorageVersionConfig) {
+    if (!storageConfig.softDeleteCurrentData) {
```
simolus3, line 229: Even if we don't use those tables by default, does it really hurt to create them unconditionally in the migration?

A downside I see is that we'd then have to write subsequent migrations if we want to change them before making v3 the default? But since the feature is unstable, couldn't we also use destructive migrations to delete and re-create tables in that case?

rkistner: I want to move away from the old migration system, but don't have enough examples to implement a full new system yet. So for now I'm just using the `CREATE ... IF NOT EXISTS` approach.

Two reasons for this:
1. Long-term, I want better support for downgrading service versions. Migrations currently block that - you have no practical option to migrate to a version that does not include the migration, other than dropping and re-creating the database. Storage versions handle down migrations fine, as long as you're not actively using newer storage versions than what is supported in the service version (still need to implement an actual process to do that).
2. As you noted, we may want to do a _lot_ of changes still in storage v3 before making it the default. Current plan is to use v3 for incremental reprocessing, which would have much bigger changes.

I think we can better finalize the structure for this once we get more of the incremental reprocessing changes implemented - it would give a lot more actual examples to work from.

simolus3: Makes sense, thanks for the explanation 👍

PR #428, opened 2025-12-08
Postgres custom type edge cases

### LINKED ISSUE #405: [Postgres] box[] values are not parsed correctly on initial replication
#403 fixed parsing of some custom types during initial replication. However, the `box[]` type still replicates incorrectly.

For this input type:
```
ARRAY[box(point '(1,2)', point '(3,4)'), box(point '(5, 6)', point '(7,8)')]
```

Streaming replication works correctly:
```
["(3,4),(1,2)","(7,8),(5,6)"]
```

But initial replication gives this:

```
["(3","4)","(1","2);(7","8)","(5","6)"]
```

The issue is that by the time we get to the custom decoding here, the incorrect array decoding was already performed:

https://github.com/powersync-ja/powersync-service/blob/c656695f7e1031f723a2c460d0544c3208422fe0/modules/module-postgres/src/replication/WalStream.ts#L587

I'm not sure what the best place is to fix this, but since it's a fairly niche edge case, we can wait for @simolus3 to take a look.

rkistner:
During the initial replication (via `WalStream.snapshotTable`), this now takes the type registry into account when decoding values. This fixes parsing `box[]` values, which were already parsed correctly for streaming replication (using the type registry) but not the initial one. Closes #405.

This also makes decoding tests in `pg_test.test.ts` more realistic by fetching types for regular queries.

Additionally, this fixes decoding arrays of enums, e.g. `moods[]`. These used to be replicated as `{sad,happy}` for both the snapshot and streaming queries, which has been fixed to `["sad","happy"]` in this PR. 

Finally, one thing that is still failing is decoding custom domain values (see the `rating_value` test). Legacy format gives a number `1` for the snapshot query, but a string `"1"` for streaming replication. The reason is that postgres seems to return the inner type as a `typeOid` when selecting domain values?

### DIFF
--- a/modules/module-postgres/src/replication/WalStream.ts
+++ b/modules/module-postgres/src/replication/WalStream.ts
@@ -559,8 +559,6 @@ WHERE  oid = $1::regclass`,
     }
     await q.initialize();
 
-    let columns: { i: number; name: string; typeOid: number }[] = [];
-    let columnMap: Record<string, number> = {};
     let hasRemainingData = true;
     while (hasRemainingData) {
       // Fetch 10k at a time.
@@ -574,15 +572,6 @@ WHERE  oid = $1::regclass`,
       // There are typically 100-200 rows per chunk.
       for await (let chunk of cursor) {
         if (chunk.tag == 'RowDescription') {
-          // We get a RowDescription for each FETCH call, but they should
-          // all be the same.
-          let i = 0;
-          columns = chunk.payload.map((c) => {
-            return { i: i++, name: c.name, typeOid: c.typeOid };
-          });
-          for (let column of chunk.payload) {
-            columnMap[column.name] = column.typeOid;
-          }
           continue;
         }
 
--- a/.changeset/afraid-weeks-matter.md
+++ b/.changeset/afraid-weeks-matter.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-module-postgres': patch
+---
+
+Fix decoding arrays of enums, fix decoding `box[]` columns during initial replication.


rkistner (commented):
Fixes look good to me

modules/module-postgres/test/src/pg_test.test.ts
```diff
@@ -562,27 +551,53 @@ INSERT INTO test_data(id, time, timestamp, timestamptz) VALUES (1, '17:42:01.12'
...
+        composite: '("{2,3}",bar,sad)',
+        nested_composite: '(t,"(""{2,3}"",bar,sad)")',
+        boxes: '["(3","4)","(1","2);(7","8)","(5","6)"]',
+        mood: 'happy',
+        moods: '{sad,happy}'
+      });
+
+      const oldFormatQueried = applyRowContext(queried, CompatibilityContext.FULL_BACKWARDS_COMPATIBILITY);
+      expect(oldFormatQueried).toMatchObject({
+        rating: 1,
```
simolus3, line 569: This really should be `rating: '1'`, because that's the value the replication sync would give us.

Looking at what happens here, it seems like postgres returns the inner type id when selecting columns that are domain types (it doesn't do that for WAL messages). So we have no way to distinguish the domain value from its inner value.
This is still bad, but I'd say the fix is to enable `edition: 2` where this is no longer a problem.

modules/module-postgres/src/replication/WalStream.ts
```diff
@@ -541,8 +559,6 @@ WHERE  oid = $1::regclass`,
     }
     await q.initialize();
 
-    let columns: { i: number; name: string; typeOid: number }[] = [];
-    let columnMap: Record<string, number> = {};
```
simolus3, line 545: Unless I'm missing something, it looks like we can rely on `pgwire` to track the row description and push it into `PgRow.columns`. So we don't have to do this ourselves here.

rkistner: To be honest, I don't remember exactly why I had it like that. I may have missed that pgwire does it internally.

simolus3: It's also possible that pgwire didn't expose columns on rows in an earlier version. Either way, it seems to work now.

PR #432, opened 2025-12-09
Granular sync rule parsing

rkistner:
This is an internal factoring in the sync-rules package, and should have no visible impact on the service on its own. This makes no change to storage format or functionality, but does bring us closer to being able to implement [Incremental reprocessing](https://github.com/powersync-ja/powersync-service/discussions/349).

This changes the sync rule structure to split:
1. The old `SqlSyncRules` class is the sync rules "definition", as parsed from the sync rules yaml file.
2. A new `HydratedSyncRules` class is used to represent the combination of the parsed sync rules, with any relevant state in bucket storage. In a way, this is similar to how `VersionedSyncRules` was used before. This now uses HydrationState as a more flexible representation of the old BucketIdTransformer.

Right now, the only relevant state is the sync rules "version", used in bucket names if `versioned_bucket_ids` is enabled. This avoids the need to pass in the `bucketIdTransformer` in specific functions, instead keeping it as state in `HydratedSyncRules`.

The second big change is to split out some functionality from BucketSource:
1. `BucketDataSource`: Represents data queries, more specifically the transform of replicated data -> bucket data.
2. `ParameterIndexLookupCreator`: Represents table index lookups for parameter queries, more specifically the transform of replicated data -> parameter lookup data.

The `BucketSource` itself still exists, handling stream subscriptions and queries.

Now, along with the split, the `BucketDataSource`s and `ParameterIndexLookupCreator`s are also pulled up into a flat structure in the top-level. What this means is that when a new version of sync rules is deployed, each of the `BucketDataSource` and `ParameterIndexLookupCreator` instances can be compared with the ones in the old version of sync rules. If there is overlap, we can re-use the existing data, instead of re-replicating that from scratch. That is the core of the [Incremental reprocessing](https://github.com/powersync-ja/powersync-service/discussions/349) project.

Some specifics in how data is persisted and linked for each:
1. `BucketDataSource` represents the combination of data queries in a bucket definition. If any of those queries changes, we need a new set of buckets. Each unique `BucketDataSource` needs a unique bucket name prefix (the part before the bucket parameters).
2. `ParameterIndexLookupCreator` represents a single parameter query that references a source table.
    a. Each unique `ParameterIndexLookupCreator` needs an unique identifier in the `bucket_parameters` lookup table, but it does not have to correlate with the bucket name.
    b. `ParameterIndexLookupCreator` does not need any reference to the bucket name prefix. The same `ParameterIndexLookupCreator` can be re-used in different bucket definitions in different sync rules versions, and even in different bucket definitions in the same sync rules version.
    c. In theory, if the query is the same, the same `ParameterIndexLookupCreator` can be used across multiple bucket definitions.
3. `BucketParameterQuerierSource` (internal concept) does not have any persisted data.
    a. At _runtime_, the hydrated `BucketParameterQuerierSource` needs a reference to the `ParameterIndexLookupCreator`(s) to evaluate table lookups (if relevant). In theory, this can be different in different sync rules version.
    b. At _runtime_, the hydrated `BucketParameterQuerierSource` needs a reference to the bucket name prefixes for the related `BucketDataSource`. This can be different in different sync rules version that use the same `BucketParameterQuerierSource`.

Note that even though we can now split out a sync rules definition into these different parts, the plan is not to persist the definitions for these parts separately. We would however store mappings of definition -> bucket name prefix or parameter lookup source name, so that these could be re-used across sync rules versions.

Further work required in this PR:
* [x] Split out the implementations for sync streams.
* [x] Check that the new generated bucketPrefix is used everywhere
* [x] Use HydrationState for parameter lookup values (name and queryId).
* [x] ~Implement logic for comparing source entities, and for merging entities from multiple sync rules into one.~ - Later PR
* [x] Create a diagram to indicate the relationships between the different source entities.

### Bucket definitions

```mermaid
erDiagram
    BucketSource ||--|| BucketDefinitionDataSource : dataSource
    BucketSource ||--o{ SqlParameterQuery : parameterLookupSources
    BucketSource ||--o{ SqlParameterQuery : parameterQuerierSources
    BucketSource ||--o{ StaticSqlParameterQuery : parameterQuerierSources
    SqlParameterQuery }o--|| BucketDefinitionDataSource : querierDataSource
    StaticSqlParameterQuery }o--|| BucketDefinitionDataSource : querierDataSource

    BucketDefinitionDataSource ||--o{ SqlDataQuery : contains
    BucketSource["SqlBucketDescriptor (BucketSource)"] {
        string name
        SYNC_RULE type
        true subscribedToByDefault
    }
    BucketDefinitionDataSource["BucketDefinitionDataSource (BucketDefinitionDataSource)"] {
        string unqiueName "BucketSource.name"
        string[] bucketParameters "debug use only"
    }
    SqlDataQuery {
        TablePattern table
        string sql
    }
    SqlParameterQuery["SqlParameterQuery (ParameterIndexLookupCreator)"] {
        TablePattern sourceTable
        string sql
        ParameterLookupScope defaultLookupScope
    }
    StaticSqlParameterQuery["StaticSqlParameterQuery"] {
        string sql
    }
```

### Sync streams

Each sync stream can have multiple variants, where each variant has a specific bucket parameter "shape" (more details here that I'm ignoring for now).
Each variant can have multiple subqueries, which each references a single source table.
A variant can also use request parameters directly.
The ParameterIndexLookupCreator is tied to a specific subquery, so that it's one per source table. In code there may be multiple per subquery, but in reality it would only be one per subquery at the moment.
The BucketDataSource is tied to a variant - there is effectively one set of buckets for each variant.

The diagram here is slightly simplified from reality.

```mermaid
erDiagram
    BucketSource ||--o{ StreamVariant : variants
    StreamVariant ||--|| SyncStreamDataSource : dataSources
    StreamVariant ||--|| SyncStreamParameterQuerierSource : parameterQuerierSources
    
    SyncStreamParameterQuerierSource }o--|| SyncStreamDataSource : querierDataSource
    StreamVariant ||--o{ SubqueryEvaluator : subqueries
    SubqueryEvaluator ||--o{ SubqueryParameterLookupSource : lookupSources
    SyncStreamParameterQuerierSource ||--o{ SubqueryParameterLookupSource : "implicit lookup relationship via subqueries"

    BucketSource["SyncStream (BucketSource)"] {
        string name
        SYNC_STREAM type
        boolean subscribedToByDefault
    }
    SyncStreamDataSource["SyncStreamDataSource (BucketDataSource)"] {
        TablePattern table
        string uniqueName "BucketSource.name|variant.id"
    }
    StreamVariant {

    }
    SubqueryEvaluator {
        TablePattern table
    }
    SyncStreamParameterQuerierSource["createParameterQuerierSource (internal)"] {

    }
    SubqueryParameterLookupSource["SubqueryParameterLookupSource"] {
        TablePattern parameterTable
        ParameterLookupScope defaultLookupScope
    }
```


rkistner:
@simolus3 It would be good if you could specifically check the changes to sync streams

packages/sync-rules/src/BucketSource.ts
```diff
@@ -1,63 +1,77 @@
...
   /**
-   * Whether {@link pushBucketParameterQueriers} may include a querier where
-   * {@link BucketParameterQuerier.hasDynamicBuckets} is true.
+   * Bucket prefix if no transformations are defined.
    *
-   * This is mostly used for testing.
+   * Transformations may use this as a base, or may generate an entirely different prefix.
    */
-  hasDynamicBucketQueries(): boolean;
+  readonly defaultBucketPrefix: string;
```
simolus3: Perhaps mention that this must be unique per sync rules configuration?

packages/sync-rules/src/BucketSource.ts
```diff
@@ -69,13 +83,153 @@ export interface BucketSource {
...
   debugWriteOutputTables(result: Record<string, { query: string }[]>): void;
+}
 
-  debugRepresentation(): any;
+/**
+ * A parameter lookup source defines how to extract parameter lookup values from parameter queries.
+ *
+ * This is only relevant for parameter queries that query tables.
+ */
+export interface BucketParameterLookupSourceDefinition {
```
simolus3: nit: Maybe it's just me, but these names don't sound particularly descriptive or helpful to me (and I keep getting `BucketParameterLookupSourceDefinition` and `BucketParameterQuerierSourceDefinition` mixed up).

Since sync streams already make the definition of parameters a bit blurry, I wonder if other terms could be more intuitive here:

- For `BucketParameterQuerierSource`, maybe something like `ConnectionBucketsQuerier` or `ConnectionBucketsResolver` makes it clearer what this is actually doing.
- For `BucketParameterLookupSource`, maybe something like `IndexLookupCreator` since that's really what parameter lookups are? Then if we need different kind of indexes in the future (say a range index for IVM with order by), we could potentially add those more easily in the future by introducing super interfaces.

packages/sync-rules/src/BucketSource.ts
```diff
@@ -69,13 +83,153 @@ export interface BucketSource {
...
+ *
+ * There are two ways to define bucket sources: Via sync rules made up of parameter and data queries, and via stream
+ * definitions that only consist of a single query.
+ */
+export interface BucketDataSource {
+  /**
+   * Given a row as it appears in a table that affects sync data, return buckets, logical table names and transformed
+   * data for rows to add to buckets.
+   */
+  evaluateRow(options: EvaluateRowOptions): EvaluationResult[];
```
simolus3: All implementations are required to use the bucket prefix they get from the hydrated store in `BucketDataSourceDefinition.createDataSource`, right?

I wonder if, instead of adding the prefix manually (and an implementation forgetting to do that breaking the whole system), we should make it statically impossible to misuse this. A potential API:

```TypeScript
export interface EvaluatedRowConsumer {
  // This takes care of extracting the id and creates an EvaluatedRow
  consumeRow(parameters: SqliteJsonValue[], table: string, data: SqiteJsonRow): void
  consumeError(description: string): void
}

export interface BucketDataSourceDefinition {
  // May or may not be used to influence bucket names
  readonly uniqueName: string;

  // I don't think this method needs access to hydrated state outside of the prefix, which it shouldn't have to lookup itself.
  createDataSource(): BucketDataSource;
}

export interface BucketDataSource {
  evaluateRow(row: TableRow, consumer: EvaluatedRowConsumer) void;
}
```

A `HydratedSyncRules` instance can implement `evaluateRowWithErrors` by creating the result array, and implement `EvaluatedRowConsumer` by pushing into it while applying the bucket name prefix for the current data source.

packages/sync-rules/src/HydrationState.ts
```diff
@@ -0,0 +1,76 @@
...
+ * both to re-use mappings across hydrations of different sync rule versions, or to generate new mappings.
+ */
+export interface HydrationState<
+  T extends BucketSourceState = BucketSourceState,
+  U extends ParameterLookupScope = ParameterLookupScope
+> {
+  /**
+   * Given a bucket data source definition, get the bucket prefix to use for it.
+   */
+  getBucketSourceState(source: BucketDataSourceDefinition): T;
```
simolus3: See [this comment](https://github.com/powersync-ja/powersync-service/pull/432/changes#r2611409349) which would potentially split this into a function for creating evaluated rows and a function for looking up bucket names (if we want to be fancy, only giving each step access to the function it needs).

Following that, I wonder if we should add a utility like `lookupBucketName(source: BucketDataSourceDefinition, parameters: SqliteJsonValue[])` that queriers would use instead.

packages/sync-rules/src/BucketSource.ts
```diff
@@ -69,13 +83,153 @@ export interface BucketSource {
...
+
+export interface BucketParameterLookupSource {
+  /**
+   * Given a row in a source table that affects sync parameters, returns a structure to index which buckets rows should
+   * be associated with.
+   *
+   * The returned {@link ParameterLookup} can be referenced by {@link pushBucketParameterQueriers} to allow the storage
+   * system to find buckets.
+   */
+  evaluateParameterRow(sourceTable: SourceTableInterface, row: SqliteRow): EvaluatedParametersResult[];
```
simolus3: Here I'm also wondering if passing a consumer accepting the results might be easier. That would create `EvaluatedParameter`s by applying the one and only scope accessible to this lookup source, making it impossible to get this wrong.

packages/sync-rules/src/BucketSource.ts
rkistner: You're right that adding the prefix manually is a risk. And there's a further issue: for bucket storage, we ideally want to store all data per BucketDataSource together, for example to easily delete all data for one source. Theoretically the bucketPrefix can be used for that, but ideally we want to use a different field, and the BucketDataSource itself shouldn't care about that.

On the idea of the consumer - personally I prefer the style of returning the values rather than passing to a consumer. But either way we can get the same "type safety", without the source needing to use the bucketPrefix itself.

Also, we don't actually need a spit between the definition and the hydrated source in this case - that could be handled by the sync rules hydration.

Now where this gets more complicated is in the BucketParameterQuerier: each querier may use multiple parameter lookup sources, so needs to distinguish the output for each of them. I'll check if there is a better way to structure that.

packages/sync-rules/src/BucketSource.ts
rkistner: Yeah, I'm not a fan of the names either. Will look at these options.

packages/sync-rules/src/BucketSource.ts
rkistner: Could you please check the new structure I have now?

I removed the `BucketDataSourceDefinition` -> `BucketDataSource` split - no hydration is required on that anymore, and it doesn't deal with bucket names anymore.

I'm not using the consumer pattern, but the new APIs should make it much more difficult to make mistakes with the bucket prefix.

It is still a little ugly on the querier sides, I'll look into that next.

packages/sync-rules/src/HydrationState.ts
rkistner: I refactored the BucketDataSource part now (see my other comments), but kept this the same aside from some renaming. For queriers, I specifically want to keep it a multipart process for `BucketDataSource -> BucketDataScope -> bucketName`, since that first part of computing the scope could be more expensive, and we may need to compute the list of referenced scopes upfront.

packages/sync-rules/src/BucketSource.ts
rkistner: I went with ParameterIndexLookupCreator here now. I still want to keep the "parameter" part: Even though "parameter query" not a user-facing concept in sync streams, the name is still used in the implementation in many places, and the concept still translates to sync stream subqueries.

packages/sync-rules/src/BucketSource.ts
rkistner: I implemented the same refactoring as for BucketDataSource here:
1. Merged BucketParameterQuerierSourceDefinition + BucketParameterQuerierSource into a single ParameterIndexLookupCreator - no hydration required anymore.
2. The ParameterIndexLookupCreator now returns results without a scope. The HydratedSyncRules then adds the scope.

packages/sync-rules/src/BucketSource.ts
rkistner: I removed the concept of BucketParameterQuerierSourceDefinition now. These don't persist any data, and we never need to evaluate them individually, so it made more sense to have the functionality on HydratedBucketSource.

I kept the implementation for these queriers mostly the same, but the structure is an implementation detail now, rather than a public API.

packages/sync-rules/src/HydrationState.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Hydration state information for a source.
+ *
+ * This is what keeps track of bucket name and parameter lookup mappings for hydration. This can be used
+ * both to re-use mappings across hydrations of different sync rule versions, or to generate new mappings.
+ */
+export interface HydrationState<
+  T extends BucketDataScope = BucketDataScope,
+  U extends ParameterLookupScope = ParameterLookupScope
+> {
```
simolus3: What's the reason for these being generic? If we ever need to refer to an implementation returning specific subtypes, I assume we could introduce a subinterface returning those?

packages/sync-rules/src/streams/parameter.ts
```diff
@@ -38,12 +31,17 @@ export interface BucketParameter {
...
   parameterTable: TablePattern;
 
-  lookupsForParameterRow(sourceTable: SourceTableInterface, row: SqliteRow): SubqueryLookups | null;
-  lookupsForRequest(params: RequestParameters): ParameterLookup[];
+  indexLookupCreators(): ParameterIndexLookupCreator[];
+  // TODO: Is there a better design here?
+  // This is used for parameter _queries_. But the queries need to know which lookup scopes to
+  // use, and each querier may use multiple lookup sources, each with their own scope.
+  // This implementation here does "hydration" on each subquery, which gives us hydrated function call.
+  // Should this maybe be part of a higher-level class instead of just a function, i.e. a hydrated subquery?
```
simolus3, line 39: This is a good point, I'll revisit this for the planned sync stream improvements.

packages/sync-rules/test/src/parameter_queries.test.ts
```diff
@@ -1,24 +1,44 @@
 import { describe, expect, test } from 'vitest';
-import { ParameterLookup, SqlParameterQuery } from '../../src/index.js';
-import { BASIC_SCHEMA, identityBucketTransformer, normalizeTokenParameters, PARSE_OPTIONS } from './util.js';
+import { UnscopedParameterLookup, SqlParameterQuery, SourceTableInterface } from '../../src/index.js';
 import { StaticSqlParameterQuery } from '../../src/StaticSqlParameterQuery.js';
+import { BASIC_SCHEMA, EMPTY_DATA_SOURCE, normalizeTokenParameters, PARSE_OPTIONS } from './util.js';
 
 describe('parameter queries', () => {
```
simolus3, line 18: Maybe also add a test similar to `variants with custom hydrationState` in `streams.test.ts` here and in `data_queries.test.ts`?

Or at least add some e2e tests in `sync_rules.test.ts` to ensure sync rules also work correctly under custom bucket source and parameter index lookup scopes.

simolus3 (commented):
I have only reviewed the `packages/sync-rules` changes so far, will continue with the service and storage implementations next. The refactoring looks great to me, I only have a few minor comments.

packages/sync-rules/src/HydrationState.ts
rkistner: The idea was that the storage layer can attach additional data. But that requires a lot more work, and I'm not even sure generics here would actually help, so I'll remove it for now.

PR #433, opened 2025-12-10
Update `pgwire` to 0.8.1, improve type-safety when decoding rows

simolus3:
This updates our dependency on `pgwire` to the latest released version. Updating that package helps make our type decoding more consistent and to take custom types into account whenever we're decoding types.

Currently, we're monkey patching `pgwire` to use our type decoder (important to represent postgres integers as JS `BigNum`s). In versions `0.8.x`, this is technically no longer necessary since row values are decoded on-demand and the raw strings sent by postgres are available through `row.raw`.
I believe that parsing postgres values when they're actually accessed is a better option because it allows us to explicitly pass a registry of custom types when reading values (something we can't do in `pgwire` itself). Parsing values prematurely can cause issues for types such as `box[]`, where the old parser misinterpreted results.

So while we could technically remove our patches and remember to consistently use `row.raw[i]` instead of `row[i]`, I've added a patch throwing an error for the former. This ensures that we have to be explicit about when we support custom types.

In this PR, I've refactored all reads to behave like before (so tests continue to pass without changes). But since parsing types is now more explicit, fixing the inconsistencies can be a small follow-up PR.

Some other changes in version 0.8.x are:

- The package is no longer publishing type declarations, so I had to add my own (copied from the last 0.7.0 release with some changes that are marked with `POWERSYNC START`).
- Some parameters have been changed, e.g. `hostname` is now `host` and `_connectOptions` is now `_socketOptions`.
- The package has been migrated to proper `exports` in `package.json`, meaning that we can't import `mod.js` anymore. We have to import `index.js`, but can keep our patches for `_net`.

### DIFF
--- a/packages/jpgwire/src/util.ts
+++ b/packages/jpgwire/src/util.ts
@@ -133,7 +133,6 @@ export async function connectPgWire(
     throw e;
   }
 
-  await connection.query();
   return connection;
 }
 


modules/module-postgres/src/api/PostgresRouteAPIAdapter.ts
```diff
@@ -108,7 +108,8 @@ export class PostgresRouteAPIAdapter implements api.RouteAPI {
         results: {
           columns: result.columns.map((c) => c.name),
           rows: result.rows.map((row) => {
-            return row.map((value) => {
+            function mapAt(index: number) {
```
rkistner: Just FYI, you can do `array.map((value, index) => { }` in JS.

modules/module-postgres/src/replication/WalStream.ts
```diff
@@ -575,7 +571,8 @@ WHERE  oid = $1::regclass`,
         const rows = chunk.rows.map((row) => {
           let q: DatabaseInputRow = {};
           for (let c of columns) {
-            q[c.name] = row[c.i];
+            q[c.name] = pgwire.PgType.decode(row.raw[c.i], c.typeOid);
+            row.raw[c.i];
```
rkistner: Looks like this line should be removed?

packages/jpgwire/src/pgwire_node.js
```diff
@@ -1,37 +1,14 @@
...
-    return Uint8Array.from(buf);
-  }
-});
-
 Object.assign(_net, {
-  connect(options) {
-    return SocketAdapter.connect(options);
+  connectTcp({ host, port, keepAlive }) {
+    return SocketAdapter.connect({ host, port, keepAlive });
+  },
```
rkistner: We pass custom `tlsOptions` and `lookup` options through here to SocketAdapter - this change removes support for that.

Additionally, `SocketAdapter` expects a `hostname` parameter, while this uses `host`. That makes it always connect to localhost, instead of the provided host.

Side note: Is it perhaps feasible at this point to convert this file to TS? It would detect at least the latter issue.

packages/jpgwire/src/util.ts
```diff
@@ -109,45 +109,63 @@ export async function connectPgWire(
...
-    tlsOptions = makeTlsOptions(config);
   } else {
     connectionOptions.sslmode = 'disable';
   }
+
   const connection = pgwire.pgconnection(connectionOptions as pgwire.PgConnectOptions);
+  patchConnection(connection, config);
 
-  // HACK: Not standard pgwire options
+  await connection.query();
```
rkistner: Is this call intentional?
I assume it would be to open/initialize the connection, but it would be good to have a comment here explaining this.

Also, it appears a connection error is now reported as "query failed", rather than the underlying connection error. The `cause` still shows the underlying error, but it would be good to not have `query failed` confusing the message. Example I'm seeing:

```
error: [powertest_173_b46f] Replication error postgres query failed {"stack":"Error: postgres query failed\n    at PgConnection.stream (file:///home/ralf/src/powersync-service/node_modules/.pnpm/pgwire@0.8.1/node_modules/pgwire/mod.js:334:15)\n    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at async PgResult.fromStream (file:///home/ralf/src/powersync-service/node_modules/.pnpm/pgwire@0.8.1/node_modules/pgwire/mod.js:1063:22)\n    at async Module.connectPgWire (file:///home/ralf/src/powersync-service/packages/jpgwire/dist/util.js:72:5)\n    at async PgManager.replicationConnection (file:///home/ralf/src/powersync-service/modules/module-postgres/dist/replication/PgManager.js:36:16)\n    at async WalStream.replicate (file:///home/ralf/src/powersync-service/modules/module-postgres/dist/replication/WalStream.js:645:47)\n    at async WalStreamReplicationJob.replicateOnce (file:///home/ralf/src/powersync-service/modules/module-postgres/dist/replication/WalStreamReplicationJob.js:114:13)\n    at async WalStreamReplicationJob.replicate (file:///home/ralf/src/powersync-service/modules/module-postgres/dist/replication/WalStreamReplicationJob.js:41:13)"}
error: [powertest_173_b46f] cause self-signed certificate {"code":"DEPTH_ZERO_SELF_SIGNED_CERT","stack":"Error: self-signed certificate\n    at TLSSocket.onConnectSecure (node:_tls_wrap:1679:34)\n    at TLSSocket.emit (node:events:519:28)\n    at TLSSocket._finishInit (node:_tls_wrap:1078:8)\n    at ssl.onhandshakedone (node:_tls_wrap:864:12)"}
```

modules/module-postgres/src/api/PostgresRouteAPIAdapter.ts
simolus3: The issue is that the row itself is no longer an array, but we can do that through `row.raw`.

packages/jpgwire/src/util.ts
simolus3: It is intentional, `pgwire` has been changed to connect lazily. Running an empty `query()` is what the library does as well. I agree the "postgres query failed" error is not helpful, I've added a `catch` to recover the inner error in that case.

I've also added comments with links to the relevant sources in `pgwire`.

packages/jpgwire/src/pgwire_node.js
simolus3: Yikes 😬 Good point, thanks!

I've added a unit test ensuring hostnames and custom `lookup` functions work. I've also manually tested a service build against supabase which uses a custom certificate through `tlsOptions`.

> Is it perhaps feasible at this point to convert this file to TS?

TypeScript complains when importing an external package without types. I've tried adding a `.d.ts` file for that explicitly, but ran into errors for packages consuming `jpgwire`.

packages/jpgwire/src/util.ts
```diff
@@ -109,45 +109,78 @@ export async function connectPgWire(
...
+    if (e instanceof Error && e.message == 'postgres query failed') {
+      // We didn't actually send a query, so we report the inner cause instead:
+      // https://github.com/exe-dealer/pgwire/blob/24465b25768ef0d9048acee1fddc748cf1690a14/mod.js#L334
+      throw e.cause;
+    }
+    throw e;
+  }
 
-  // HACK: Not standard pgwire options
+  await connection.query();
```
rkistner: The block above looks good, but this looks like a duplicate?

PR #436, opened 2025-12-12
Reworked dynamic module loading, fixing startup issues for migration and compact jobs

Rentacookie:
Rework dynamic module loading, fixing startup issues for migration and compact jobs in 1.18.0 / 1.18.1

The main problem was that the jobs run in a different context to the full service, and in that context not all the required modules had been registered yet at time of execution.

### DIFF
--- a/.changeset/nasty-snakes-double.md
+++ b/.changeset/nasty-snakes-double.md
@@ -3,4 +3,4 @@
 '@powersync/service-image': patch
 ---
 
-Rework dynamic module loading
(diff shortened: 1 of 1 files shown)


.changeset/nasty-snakes-double.md
```diff
@@ -0,0 +1,6 @@
+---
+'@powersync/service-core': patch
+'@powersync/service-image': patch
+---
+
+Rework dynamic module loading
```
rkistner: Maybe specifically note this fixes startup issues in the migration and compact jobs in 1.18.0 / 1.18.1.

rkistner (approved):

PR #446, opened 2025-12-24
Optimize Postgres compactor queries

jaswdr:
### Problem Statement

In the context of #400, the current compaction query uses an OR condition with COLLATE "C" for cursor-based pagination:

```sql
SELECT
  op,
  op_id,
  source_table,
  table_name,
  row_id,
  source_key,
  bucket_name
FROM
  bucket_data
WHERE
  group_id = $1
  AND bucket_name >= $2
  AND (
    (
      bucket_name = $3
      AND op_id < $4
    )
    OR bucket_name < $3 COLLATE "C"
  )
ORDER BY
  bucket_name DESC,
  op_id DESC
LIMIT
  $5
```

The OR condition prevents Postgres from using the index bounds and removes the majority of the rows, as we can see in the query plan with EXPLAIN ANALYZE in below sample query executed during compaction:

```sql
EXPLAIN ANALYSE SELECT 
  op,
  op_id,
  source_table,
  table_name,
  row_id,
  source_key,
  bucket_name
FROM bucket_data
WHERE
  group_id = 1
  AND bucket_name >= ''
  AND ((
    bucket_name = 'global[]'
    AND op_id < 49976) 
    OR bucket_name < 'global[]' COLLATE "C")
ORDER BY bucket_name DESC, op_id DESC
LIMIT 10000
```

Output:

```
Limit  (cost=0.42..27866.77 rows=10000 width=114) (actual time=144.147..146.449 rows=10000 loops=1)
  Buffers: shared hit=4729 read=50886
  ->  Index Scan Backward using unique_id on bucket_data  (cost=0.42..160948.06 rows=57757 width=114) (actual time=144.146..146.125 rows=10000 loops=1)
        Index Cond: ((group_id = 1) AND (bucket_name >= ''::text))
        Filter: (((bucket_name = 'global[]'::text) AND (op_id < 49976)) OR (bucket_name < 'global[]'::text COLLATE "C"))
        Rows Removed by Filter: 950000 <-- this is the problem, we are scanning more than we should
        Buffers: shared hit=4729 read=50886
Planning Time: 0.070 ms
Execution Time: 146.637 ms
```

PostgreSQL scans all rows matching the Index Cond and then applies the Filter, discarding most rows as we can see in the Rows Removed by Filter section.
Performance Impact

### Solution

To optimize this query, I'm proposing to split the code into 3 specialized queries. Each query will handle a specific parameter case passed to `compact` command. Check the changes in the PostgresCompactor.ts file to see the changes. I've also created unit and integration tests to verify correctness.

### Results

By running both the code currently in master branch and this branch I got these numbers locally when compacting a database with 1 million bucket records.

| Branch | Time | Throughput |
|--------|------|------------|
| PR | 8.67s | 116,451 rec/s |
| main | 15.61s | 64,708 rec/s |

### DIFF
--- a/modules/module-postgres-storage/src/storage/PostgresCompactor.ts
+++ b/modules/module-postgres-storage/src/storage/PostgresCompactor.ts
@@ -3,11 +3,82 @@ import { logger, ReplicationAssertionError } from '@powersync/lib-services-frame
 import { InternalOpId, storage, utils } from '@powersync/service-core';
 import * as pgwire from '@powersync/service-jpgwire';
 import * as t from 'ts-codec';
-import { BIGINT_MAX } from '../types/codecs.js';
 import { models } from '../types/types.js';
 import { sql } from '../utils/db.js';
 import { pick } from '../utils/ts-codec.js';
 import { encodedCacheKey } from './batch/OperationBatch.js';
+
+export type BucketFilter =
+  | { mode: 'all' }
+  | { mode: 'exact'; bucketName: string }
+  | { mode: 'prefix'; bucketPrefix: string };
+
+export interface PaginationCursor {
+  bucket: string;
+  opId: bigint;
+}
+
+/**
+ * Validates that a bucket filter string is either:
+ * 1. A valid bucket definition name (no brackets) - e.g., "user", "global"
+ * 2. A valid full bucket name (name + JSON array) - e.g., "global[]", "user[\"u1\"]"
+ */
+export function isValidBucketNameOrPrefix(bucket: string): boolean {
+  if (bucket.length === 0) {
+    return false;
+  }
+
+  const bracketIndex = bucket.indexOf('[');
+
+  if (bracketIndex === -1) {
+    return true;
+  }
+
+  if (!bucket.endsWith(']')) {
+    return false;
+  }
+
+  const paramPart = bucket.slice(bracketIndex);
+
+  try {
+    const parsed = JSON.parse(paramPart);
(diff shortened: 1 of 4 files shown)


modules/module-postgres-storage/src/storage/PostgresCompactor.ts
```diff
@@ -3,12 +3,83 @@ import { logger, ReplicationAssertionError } from '@powersync/lib-services-frame
...
+  }
+
+  if (bucket.includes('[')) {
+    return { mode: 'exact', bucketName: bucket };
+  } else {
+    return { mode: 'prefix', bucketPrefix: `${bucket}[` };
+  }
+}
+
+const COMPACT_ROW_CODEC = pick(models.BucketData, [
```
jaswdr, line 71: Moving this to global scope since it never changes.

modules/module-postgres-storage/src/storage/PostgresCompactor.ts
```diff
@@ -3,12 +3,83 @@ import { logger, ReplicationAssertionError } from '@powersync/lib-services-frame
...
+  }
+
+  if (!bucket.endsWith(']')) {
+    return false;
+  }
+
+  const paramPart = bucket.slice(bracketIndex);
+
+  try {
+    const parsed = JSON.parse(paramPart);
```
jaswdr, line 44: I'm using `JSON.parse` to avoid writing a custom parsing function, let me know what do you think.

Issue #423, opened 2025-12-03
YAML Environment Substitution Swallows Validation Errors


stevensJourney:
# Overview

When passing in an invalid YAML config via a Base64 string (might affect other methods also), the YAML validation error is not printed to the console. Instead only the JSON parsing error is shown. This makes it difficult to debug invalid configurations. 

E.g. if passing in a YAML config which contains `!env` substitutions which have either undefined env vars or use variable names not starting with `PS_` the following error is printed.

```
Error: Unexpected token 'r', "replicatio"... is not valid JSON 
{"stack":"Error: Could not collect config using Base64 method. Caught exception: Error: Could not parse PowerSync config file content as JSON or YAML: SyntaxError: Unexpected token 'r', \"replicatio\"... is not valid JSON\n    at CompoundConfigCollector.collectBaseConfig (file:///Users/stevenontong/Documents/platform_code/powersync/service/powersync-service/packages/service-core/dist/util/config/compound-config-collector.js:164:23)\n    at async CompoundConfigCollector.collectConfig 

(file:///Users/stevenontong/Documents/platform_code/powersync/service/powersync-service/packages/service-core/dist/util/config/compound-config-collector.js:27:28)\n    at async startUnifiedRunner (file:///Users/stevenontong/Documents/platform_code/powersync/service/powersync-service/service/src/runners/unified-runner.ts:10:20)\n    at async Command.<anonymous> (file:///Users/stevenontong/Documents/platform_code/powersync/service/powersync-service/packages/service-core/dist/entry/commands/start-action.js:12:9)\n    at async Command.parseAsync (/Users/stevenontong/Documents/platform_code/powersync/service/powersync-

service/node_modules/.pnpm/commander@12.1.0/node_modules/commander/lib/command.js:1092:5)\n    at async runProgram (file:///Users/stevenontong/Documents/platform_code/powersync/service/powersync-service/packages/service-core/dist/entry/cli-entry.js:31:17)"}
``` 

The service should not swallow the YAML validation errors.

hamzzy:
@stevensJourney  i have added a pr for this issue

PR #449, opened 2026-01-07
fix: Improve config parsing error messages with YAML details and add …

hamzzy:
# Summary

This PR fixes an issue where detailed YAML validation errors (e.g., invalid !env substitutions) were being swallowed during configuration parsing. Instead of a generic and misleading JSON parsing error, users will now see the specific YAML error, making it significantly easier to debug configuration issues #423 .

# Problem
Currently, if a user provides an invalid YAML configuration (for example, via Base64 string) that contains an undefined environment variable or a variable that does not start with the mandatory PS_ prefix, the service swallows the YAML error and falls back to JSON parsing. When JSON parsing inevitably fails, only the JSON error is reported.

Example of current misleading error:

```Error: Unexpected token 'r', "replicatio"... is not valid JSON```
### Solution

1. Modified `ConfigCollector.parseContent()` in  packages/service-core to: Capture any errors thrown during the parseYaml() phase.

2. If the fallback parseJSON() phase also fails, include the captured YAML error in the final thrown exception.

### Improved error output:
```
Could not parse PowerSync config file content as JSON or YAML: SyntaxError: Unexpected token '#', " # PowerSyn"... is not valid JSON
YAML Error: Error: Could not parse YAML configuration file. Received errors: 
 Attempting to substitute environment variable INVALID_VAR is not allowed. Variables must start with "PS_" at line 6, column 15:
        type: !env INVALID_VAR
              ^^^^
```
# Verification Results

### Automated Tests
Ran the full test suite for @powersync/service-core:
- Result: 13 passed, 109 passed (total)
- Duration: ~5s

### DIFF
--- a/modules/module-mongodb-storage/src/storage/implementation/MongoPersistedSyncRules.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/MongoPersistedSyncRules.ts
@@ -1,7 +1,6 @@
-import { SqlSyncRules, HydratedSyncRules } from '@powersync/service-sync-rules';
+import { SqlSyncRules, HydratedSyncRules, versionedHydrationState } from '@powersync/service-sync-rules';
 
 import { storage } from '@powersync/service-core';
-import { versionedHydrationState } from '@powersync/service-sync-rules/src/HydrationState.js';
 
 export class MongoPersistedSyncRules implements storage.PersistedSyncRules {
   public readonly slot_name: string;
--- a/modules/module-postgres-storage/src/storage/sync-rules/PostgresPersistedSyncRulesContent.ts
+++ b/modules/module-postgres-storage/src/storage/sync-rules/PostgresPersistedSyncRulesContent.ts
@@ -1,10 +1,9 @@
 import * as lib_postgres from '@powersync/lib-service-postgres';
 import { ErrorCode, logger, ServiceError } from '@powersync/lib-services-framework';
 import { storage } from '@powersync/service-core';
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: packages/service-core/src/util/config/collectors/config-collector.ts [hunk 1]
```diff
@@ -70,13 +70,19 @@ export abstract class ConfigCollector {
...
-        } catch (ex) {}
+        } catch (ex) {
+          yamlError = ex;
+        }
         try {
           return this.parseJSON(content);
         } catch (ex) {
-          throw new Error(`Could not parse PowerSync config file content as JSON or YAML: ${ex}`);
+          throw new Error(
+            `Could not parse PowerSync config file content as JSON or YAML: ${ex}${yamlError ? `\nYAML Error: ${yamlError}` : ''
```
stevensJourney: Minor nit: Can the JSON error be prefixed simillarly to the YAML error
```suggestion
            `Could not parse PowerSync config file content as JSON or YAML.\nJSON Error: ${ex}${yamlError ? `\nYAML Error: ${yamlError}` : ''
```

stevensJourney (commented):
Thanks for the PR! I'm generally happy with this. I only have one minor nit request for the formatting of the message.

packages/service-core/src/util/config/collectors/config-collector.ts
hamzzy: sure, on it

PR #451, opened 2026-01-08
New sync streams compiler, part one

simolus3:
This adds a new compiler for sync streams. 

## Motivation

This compiler is written from scratch, using almost none of the existing tools we have to parse and implement SQL expressions. To explain this perhaps odd choice, we need to look at some of the things we want sync streams to support:

### Multiple data queries

A common complaint about sync streams is that each stream can only define a single `SELECT` statement. Users typically have sync rules like `parameter: SELECT auth.user_id() as uid; data: [SELECT * FROM notes WHERE owner = bucket.uid, SELECT * FROM another WHERE owner = bucket.uid, ...]`. With sync streams, they'd have to define one stream per query, which increases the amount of buckets.

A key challenge here is that parameters are implicit in sync streams. When we force users to write parameter and data queries separately, all data queries can trivially go into the same bucket since they're based on the same parameters. With streams, figuring out whether queries are compatible requires a better understanding of their sources. If we have queries like `SELECT * FROM notes WHERE owner = auth.user_id()` and `SELECT * FROM settings WHERE owner = auth.user_id()`, we technically see two independent anonymous parameters. To put those into the same bucket, we need to check that their instantiation (`auth.user_id()`) is the same.

### Parameter reuse

Even after supporting multiple data queries per stream, the anonymous nature of stream parameters means they can't be reused. Where sync rules allowed parameter queries like `SELECT org_id FROM org_memberships WHERE user = auth.user_id()`, sync streams requires parameters to be inlined: `SELECT * FROM repositories WHERE org_id IN (SELECT org_id FROM org_memberships WHERE user = auth.user_id())`.

If multiple streams use the same subquery, we currently create duplicate lookups. This is inefficient for storage size and, since there's a lookup limit, can restrict the usefulness of sync streams. Once again, we need a source-level understanding to figure out that the subquery is the same and can be re-used.

### Parameter merging

An easy footgun with sync streams today is that `AND` clauses can become expensive. We have seen an instance of this ourselves with queries like `SELECT * FROM comments WHERE issue_id = subscription.parameter('issue') AND issue_id IN auth.parameter('issues')`. Because there are two matches on `issue_id`, we currently create two parameters and compute the cartesian product of them during lookups. This `O(n²)` behavior causes us to hit lookup limits way too quickly. While the query can be rewritten to be more efficient, it would be nice if we could recognize that two filters on the same column don't need two independent parameters.

### Consequences

What all of these features have in common is that they're extremely hard to implement with types like `BaseClause` we have today. Because those are implemented as JavaScript objects and functions, we don't have the insight we need to figure out whether two parameters are equal. Two instances of `owner = auth.user_id()` are two `[object Object]`'s and we can't compare them.

Of course, we might want to attach source offsets or an SQL substring to each clause and compare those. But even that is tricky to do correctly:

- comparing unqualified references by name means we can't be sure they don't refer to different tables.
- qualified references might refer to the same table under a different alias in different streams, but we still want to detect that it's fundamentally the same query.

And thus, our need to de-duplicate buckets and parameter queries means we need to come up with a different scheme. Instead of using objects implementing expressions through callbacks, the new compiler uses a declarative model describing their AST. And instead of the compiler output implementing behavior, the new compiler outputs a _sync plan_, a stateless description of how data get mapped to buckets and parameter lookups as well as how buckets are to be resolved. The service will then interpret that plan during replication.

Outside of this being a requirement for features we're interested in, this has a number of other benefits:

1. The new compiler is much easier to test: While the old one could only be tested by observing sync behavior for fixed inputs, the sync plan is a serializable data structure. This means we can have a large number of expected inputs -> sync plan assertions that are much easier to write than manual behavior tests.
2. We are currently evaluating scalar expressions in JavaScript, essentially re-implementing SQLite's functionality. In the sync plan, scalar expressions are represented as dependency-free SQL strings. This means we'll be able to use SQLite to evaluate them, which is faster and guaranteed to be accurate.
3. Since the sync plan is serializable, we can store it in bucket storage and load it on startup instead of re-compiling sync streams on every service startup. Currently, there is an implicit contract between buckets that may have been created by an older service version and buckets that are resolved for connecting users today. This means we can't change the compiler without breaking existing deployments. By introducing the sync plan as a small but stable structure, we can deliver subsequent features and optimizations without breaking deployed sync rules.
4. The sync plan describes how buckets and parameter lookups are designed as elements that can be compared efficiently. This should help with plans for incremental reprocessing, since we can do a diff between plans and only re-compute changed buckets, keeping others unchanged.
5. We are considering the option to push some filters into the source database to speed up initial replication. By having the AST available, we can match the filter in queries like `SELECT * FROM notes WHERE NOT is_archived AND owner = auth.user_id()` to realize that `NOT is_archived` is a static filter we can run on the source database instead of fetching all rows.

## New features

A new feature unlocked by having more insights into SQL and data dependencies between subexpressions is that we can support nested subqueries. This is a commonly-requested feature, since it allows writing queries that traverse through a normalized relational structure. The new compiler supports queries like the following:

```SQL
select * from comments where issue_id in (
  select id from issues where owner_id in (
    select id from users where is_admin AND id = auth.user_id()
  )
)
```

Almost without no additional effort, this also allows us to handle a subset of joins. For instance, the query above is equivalent to this one, which the new compiler also supports:

```SQL
SELECT * FROM comments c
  INNER JOIN issues i ON c.issue_id = i.id
  INNER JOIN users u ON i.owner_id = u.id
  WHERE u.is_admin AND u.id = auth.user_id()
```

Allowing these queries should make sync streams much easier to write.

## Implementation sketch

The `SyncPlan` interface in `sync-rules/src/sync_plan/plan.ts` is the output from the compiler, we serialize it to JSON for tests.
Since a stream can have multiple data queries, these are added independently. The main entrypoint thus looks like this:

https://github.com/powersync-ja/powersync-service/blob/23f50d74ea6ecd4915bc12187a1047b1d5a828f2/packages/sync-rules/test/src/compiler/utils.ts#L54-L69

Compilation begins by `parser.ts` desugaring subquerie, analyzing dependencies for expressions and merging all joins into a single DNF filter condition. Filters are optimized to avoid creating parameters when that's not necessary (`filter_simplifier.ts`). Then, `querier_graph.ts` traverses filter clauses to create queriers while also instantiating parameter lookups and bucket parameters. Finally, `ir_to_sync_plan.ts` translates the mutable IR from the compiler into a sync plan.

## Follow-up work

- Schema-aware validation: Validate that table references exist and use it to emit warnings / build the expected client-side schema. Also warn when auto-subscribe streams use parameters.
- Persist in bucket storage: We want to persist the compiled sync plan and restore it instead of re-compiling from scratch every time the service starts. This allows us to update the compiler (e.g. to add new optimizations) without breaking existing streams.
- Restore support for the overlap operator (`&&`)? I'm not sure if this is that critical since we don't seem to be documenting that this exists? I have ideas on how the operator could be implemented.
- Actually use it: This just adds the compiler and tests, but parsing sync rules still uses the old logic.
- More items to get sync streams to beta, such as CTEs.

### DIFF
--- a/packages/sync-rules/src/sync_plan/plan.ts
+++ b/packages/sync-rules/src/sync_plan/plan.ts
@@ -2,6 +2,7 @@ import { BucketPriority } from '../BucketDescription.js';
 import { ParameterLookupScope } from '../HydrationState.js';
 import { TablePattern } from '../TablePattern.js';
 import { UnscopedEvaluatedParameters } from '../types.js';
+import { SqlExpression } from './expression.js';
 
 /**
  * A compiled "sync plan", a description for
@@ -160,25 +161,6 @@ export interface StreamQuerier {
    */
   bucket: StreamBucketDataSource;
   sourceInstantiation: ParameterValue[];
-}
-
-/**
- * An expression that can be evaluated by SQLite.
- *
- * The type parameter `T` describes which values this expression uses. For instance, an expression in a
- * {@link TableProcessor} would only use {@link ColumnSqlParameterValue}s since no request is available in that context.
- */
-export interface SqlExpression<T extends SqlParameterValue> {
-  /**
-   * To SQL expression to evaluate.
-   *
-   * The expression is guaranteed to not contain any column references. All dependencies to row data are encoded as SQL
-   * parameters that have an instantiation in {@link instantiation}.
-   *
-   * For instance, the stream `SELECT UPPER(name) FROM users;` would have `UPPER(?1)` as SQL and a
-   */
-  sql: string;
-  values: T[];
 }
 
 export type SqlParameterValue = ColumnSqlParameterValue | RequestSqlParameterValue;
--- a/packages/sync-rules/src/sync_plan/serialize.ts
+++ b/packages/sync-rules/src/sync_plan/serialize.ts
@@ -1,4 +1,5 @@
 import { TablePattern } from '../TablePattern.js';
+import { SqlExpression } from './expression.js';
 import {
   ColumnSource,
   ColumnSqlParameterValue,
@@ -6,7 +7,6 @@ import {
   ParameterValue,
   PartitionKey,
   RequestSqlParameterValue,
-  SqlExpression,
   StreamBucketDataSource,
   StreamDataSource,
   StreamOptions,
--- a/packages/sync-rules/src/compiler/compiler.ts
+++ b/packages/sync-rules/src/compiler/compiler.ts
@@ -6,6 +6,7 @@ import { StreamOptions, SyncPlan } from '../sync_plan/plan.js';
 import { CompilerModelToSyncPlan } from './ir_to_sync_plan.js';
 import { QuerierGraphBuilder } from './querier_graph.js';
 import { StreamQueryParser } from './parser.js';
+import { NodeLocations } from './expression.js';
 
 /**
  * State for compiling sync streams.
@@ -39,6 +40,7 @@ export class SyncStreamsCompiler {
         const parser = new StreamQueryParser({
           compiler: this,
           originalText: sql,
+          locations: new NodeLocations(),
           errors
         });
         const query = parser.parse(stmt);
--- a/packages/sync-rules/src/compiler/expression.ts
+++ b/packages/sync-rules/src/compiler/expression.ts
@@ -1,8 +1,12 @@
-import { Expr } from 'pgsql-ast-parser';
+import { Expr, NodeLocation, PGNode } from 'pgsql-ast-parser';
 import { SourceResultSet } from './table.js';
 import { EqualsIgnoringResultSet, equalsIgnoringResultSetList } from './compatibility.js';
 import { StableHasher } from './equality.js';
 import { ConnectionParameterSource } from '../sync_plan/plan.js';
+import { ExternalData, SqlExpression } from '../sync_plan/expression.js';
+import { ExpressionToSqlite } from '../sync_plan/expression_to_sql.js';
+import { RecursiveExpressionVisitor } from '../sync_plan/expression_visitor.js';
+import { getLocation } from '../errors.js';
 
 /**
  * An analyzed SQL expression tracking dependencies on non-static data (i.e. rows or connection sources).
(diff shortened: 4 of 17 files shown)


packages/sync-rules/src/sync_plan/plan.ts
```diff
@@ -0,0 +1,235 @@
...
+export interface SqlExpression<T extends SqlParameterValue> {
+  /**
+   * To SQL expression to evaluate.
+   *
+   * The expression is guaranteed to not contain any column references. All dependencies to row data are encoded as SQL
+   * parameters that have an instantiation in {@link instantiation}.
+   *
+   * For instance, the stream `SELECT UPPER(name) FROM users;` would have `UPPER(?1)` as SQL and a
+   */
+  sql: string;
```
simolus3: Instead of keeping a string forming the expression and instantiations as separate objects, I think we would have a better time with our own stable / serializable AST for expressions.

After all, there aren't that many types of expressions and having the AST available would make it easier to:

- validate schemas when only having a serialized plan available
- generate output schemas when only having a serialized plan available
- prefilter (https://github.com/powersync-ja/powersync-service/pull/469).

I think it makes sense to address that in a follow-up PR.

packages/sync-rules/src/compiler/bucket_resolver.ts
```diff
@@ -0,0 +1,159 @@
...
+    equalsIgnoringResultSetUnordered.hash(hasher, this.requestFilters);
+    this.resolvedBucket.buildInstantiationHash(hasher);
+  }
+
+  hasIdenticalInstantiation(other: StreamResolver) {
+    if (other.options != this.options) {
+      return false;
+    }
+
+    if (!equalsIgnoringResultSetUnordered.equals(this.requestFilters, this.requestFilters)) {
```
rkistner:
```suggestion
    if (!equalsIgnoringResultSetUnordered.equals(other.requestFilters, this.requestFilters)) {
```

packages/sync-rules/src/sync_plan/serialize.ts
```diff
@@ -0,0 +1,210 @@
...
+  function serializeDataSources(): SerializedDataSource[] {
+    return plan.dataSources.map((source, i) => {
+      dataSourceIndex.set(source, i);
+
+      return {
+        hash: source.hashCode,
+        table: serializeTablePattern(source.sourceTable),
+        outputTableName: source.outputTableName,
+        filters: source.filters,
+        partition_by: source.parameters,
```
rkistner: Nitpick: Other serialized fields use camelCase, so should probably use `partitionedBy` here? Same for serializeParameterIndexes below.

packages/sync-rules/src/sync_plan/serialize.ts
```diff
@@ -0,0 +1,210 @@
...
+      table: pattern.tablePattern
+    };
+  }
+
+  function serializeDataSources(): SerializedDataSource[] {
+    return plan.dataSources.map((source, i) => {
+      dataSourceIndex.set(source, i);
+
+      return {
+        hash: source.hashCode,
```
rkistner, line 44: How is the hash used in the serialized form? I assume it can be re-calculated from the rest of the serialized data?

packages/sync-rules/src/sync_plan/plan.ts
```diff
@@ -0,0 +1,235 @@
...
+  values: T[];
+}
+
+export interface SqlParameterValue {
+  /**
+   * The start index and length of the parameter in {@link SqlExpression.sql} that this parameter is instantiating.
+   *
+   * This allows merging equal parameters, which is useful to compose SQL expressions.
+   */
+  sqlPosition: [number, number];
```
rkistner: What is the position used for - only for equality checks?

simolus3: This is an optimization to be able to remove the amount of parameters when building queries to implement buckets. The idea is that the query we build for a bucket might reference the same column / input multiple times. E.g. `SELECT id, name FROM tbl WHERE LENGTH(name) > 10` would at runtime compile to something equivalent to:

```
*evaluateRow(tbl) {
  for (const row in evalSql('SELECT ?1, ?2 WHERE LENGTH(?2) > 10', tbl.id, tbl.name)) {
    yield bucket(...);
  }
}
```

Since the output column and the filter are two distinct `SqlExpression` instances, we need to know their offsets to rewrite the variable index at runtime.

But this isn't actually used and doesn't feel like a good solution. I think a better approach would be to serialize expressions in a structural way instead of serializing SQL. I'll remove the `sqlPosition` entry from this PR and will open another one for the expression refactoring.

packages/sync-rules/src/sync_plan/serialize.ts
simolus3: It can't be recalculated from serialized data at the moment, but if two elements have the same serialized form they would also have the same hash code.

At the moment, adding the hash code to the serialized form ensures that the hash code is actually stable (since all tests compare codes against snapshots). Since these codes are used a lot in the compiler, it's helpful to assert that they're fully deterministic.
I assume they might also help with incremental reprocessing (essentially using the hash as a key to compare bucket data sources and parameters between sync rules instead of using the name). But that depends on your approach to that, happy to talk about that offline.

cahofmeyr:
> Maybe we can revisit this once sync streams are stable, at which point we just call it `packages/sync-streams` and just happen to support the legacy alternative in the same package?

That sounds good to me :)

packages/sync-rules/src/sync_plan/plan.ts
simolus3: Done in https://github.com/powersync-ja/powersync-service/pull/471, marking this as resolved.

packages/sync-rules/src/sync_plan/serialize.ts
rkistner: Happy with this for now - can maybe revisit the hash again when we look into stabilizing the serialization format.

cahofmeyr:
> > Maybe we can revisit this once sync streams are stable, at which point we just call it `packages/sync-streams` and just happen to support the legacy alternative in the same package?
> 
> That sounds good to me :)

Closing the loop on this for anyone following - new proposal is to genericize the name to `sync-config`

PR #457, opened 2026-01-12
Add additional metrics metadata

Rentacookie:
Added the source database and storage types to the captured metrics metadata.


packages/service-core/src/metrics/open-telemetry/util.ts
```diff
@@ -59,6 +59,8 @@ export function createOpenTelemetryMetricsFactory(context: ServiceContext): Metr
   const resource = resourceFromAttributes({
     ['service']: 'PowerSync',
     ['service.version']: pkg.version,
+    ['source_type']: configuration.connections ? configuration.connections[0].type : 'Unknown',
+    ['storage_type']: configuration.storage.type ? configuration.storage.type : 'Unknown',
```
rkistner: Nitpick: Since the types use lower-case names, also use lower-case `unknown`.

Also handle the case of `connections` being defined, but an empty array. The `?.` operator is great for that.

PR #460, opened 2026-01-13
Restructure and validate typescript builds

rkistner:
## Restructure build commands

This restructures the root `pnpm build` to primarily rely on `tsc -b` on the root tsconfig.json to do builds, rather than a recursive `pnpm run -r build`. This is much faster, especially for incremental builds (see #435).

The limited number of cases that need custom builds (e.g. schema generation) are now handled through a `pnpm run -r build:extra`.

Tests are now also referenced in the root `tsconfig.json`, ensuring that everything is built at the same time. This now also catches type errors in `rsocket-router` tests.

## Fix and validate typescript references

This adds a script to validate that all packages use typescript project references correctly:
1. Every workspace dependency should be a typescript reference, and no extra references other than that. *This includes both the main tsconfig.json of the package, as well as test/tsconfig.json*.
   a. `dependencies` should always be in the main tsconfig.json
   b. `devDependencies` may be in either the main one or `test/tsconfig.json`. For docker build purposes, we prefer `test/tsconfig.json` for this when feasible, but that's not a strict check.
2. Every tsconfig.json file (including tests) must be referenced in the root tsconfig.json.

This scripts runs on CI. The script itself was mostly generated using Codex, but manually checked and tested.

This ensures that:
1. There are no race conditions in the build process. We've occasionally had issues only on some developer machines due to build order issues from missing references.
3. In an editor, navigating to a function implementation now always navigates to the source, instead of occasionally to a built .d.ts file.
4. `tsc -b` now always builds _everything_. No worrying about whether some packages are missed when running for example `pnpm watch:ts` in the root dir.

In the future, we may be able to semi-automate generating these references. For now, the checks at least make it difficult to make mistakes, and make it easy for an agent to fix.

### DIFF
--- a/scripts/validate-tsconfig-references.mts
+++ b/scripts/validate-tsconfig-references.mts
@@ -3,6 +3,7 @@ import path from 'node:path';
 import process from 'node:process';
 import { fileURLToPath } from 'node:url';
 import fg from 'fast-glob';
+import { findWorkspacePackages, type Project } from '@pnpm/workspace.find-packages';
 
 const ROOT_DIR = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
 const PROD_DEP_SECTIONS = ['dependencies', 'peerDependencies', 'optionalDependencies'] as const;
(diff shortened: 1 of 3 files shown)


scripts/validate-tsconfig-references.mts
```diff
@@ -0,0 +1,479 @@
...
+    if (/^\s*$/.test(line) || /^\s*#/.test(line)) {
+      continue;
+    }
+    const match = line.match(/^\s*-\s*['"]?(.+?)['"]?\s*$/);
+    if (match) {
+      patterns.push(match[1]);
+      continue;
+    }
+    if (/^\S/.test(line)) {
+      break;
```
simolus3: Perhaps add a comment mentioning that this is the start of a new block that isn't `packages:` and that we don't care about remaining file contents.

simolus3 (dismissed):
I'm happy with these changes. In the JS SDK, we're using `@pnpm/workspace.find-packages` to find packages. But I think pulling in that dependency which is kind of internal to pnpm for this is also not really necessary.

We could perhaps use a yaml parser instead of regex to parse the workspace since we already depend on one, but LGTM either way.

PR #461, opened 2026-01-13
[Incremental Reprocessing] [WIP] [MongoDB Storage] Replication storage restructure

rkistner:
Builds on #454.

This restructures storage (MongoDB storage only) to decouple storage of different parts from sync rules:
1. SourceTable is not scoped to a specific sync rules versions anymore - it is now a many-to-many relationship.
2. CurrentData is not scoped to a specific sync rules versions anymore - it is now only scoped to a SourceTable.
3. BucketState and BucketData is now scoped to a specific sync rule definition, rather than the sync rules versions. This may be shared between different sync rules versions. The mapping is stored on the sync rules version.
4. BucketParameterDocument (parameter lookups) are not scoped to a specific sync rules versions anymore - it is instead scoped to the parameter lookup index creator [PENDING] and the SourceTable.

Bucket names are now also decoupled from the definition name. And wherever we read or write bucket data, it is now required to pass in the relevant BucketDataSource. This unfortunately complicates tests a bit, since it can't rely on static bucket names anymore.

The bucket ids are generated when a sync rules version is first loaded. It is effectively just an incrementing sequence, starting from `sync_rules_version << 17`. This will change in the future.

In the current form, the PR implements changes without much regard for the old storage structure. We'll have to implement proper backwards-compatibility and/or a migration strategy to handle existing instances.

TODO:
 * [ ] Finish refactoring of BucketParameterDocument storage.
 * [ ] Fix the compact task.
 * [ ] Fix all tests.
 * [ ] Implement a migration strategy.

### DIFF
--- a/modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
@@ -633,7 +633,6 @@ export class MongoSyncBucketStorage
         }
       }
 
-      // TODO: Only delete if not used by other sync rules
       for (let [name, id] of Object.entries(doc.rule_mapping.definitions)) {
         if (keepSyncDefinitionIds.has(id)) {
           continue;


rkistner:
Superseded by #468.

PR #465, opened 2026-01-14
Sync streams evaluator

simolus3:
This builds on https://github.com/powersync-ja/powersync-service/pull/451: While that PR adds a tool compiling sync streams from SQL to a sync plan, this adds an evaluator implementing the `BucketSource` interface over sync plans.

The foundational layer to evaluate rows, parameters and queries is `ScalarExpressionEngine` (in `scalar_expression_engine.ts`). This interface evaluates SQL queries of the form `SELECT $outputs WHERE $filter`, where both outputs and filter are deterministic and scalar expressions. There are two implementations:

1. One based on our existing tools implementing SQL expression in JavaScript.
2. A new one based on turning expression trees into SQL text that is then evaluated with `node:sqlite`.

Due to some behavioral differences between SQLite and our implementation, the SQL engine is not a drop-in replacement. But having it around helps validate that streams work with both, so we could introduce an option to let users migrate in the future.

Bucket data sources and parameter index lookup creators are implemented in a fairly straightforward way: In both cases, we have scalar filter expressions as well as output columns and parameter keys. These are all evaluated and used to create bucket parameters or lookups, respectively.

The most complex part of evaluating sync streams is finding parameters for connections. In sync plans, parameter instantiations are represented as `ParameterValue`s, which can be:

1. Derived from request parameters (e.g. the scalar expression in `WHERE id = auth.user_id()`).
2. Derived from a lookup result, where a lookup is either:
   a. A table-valued function using request parameters as input (e.g. `WHERE id IN auth.param('orgs')`).
   b. An index parameter lookup (e.g. `WHERE id IN (SELECT org_id FROM orgs WHERE ...)`).

Lookups are further categorized in stages, where lookups from later stages are allowed to use lookups from earlier stages as inputs. Outputs from a lookup may be used by multiple later lookups or `ParameterValue`s, which means that the querier is a DAG structure.
This structure is evaluated in `RequestParameterEvaluators` of `parameter_evaluator.ts`. The main idea is that nodes in the lookup graph are evaluated and replaced with their instantiation (essentially a form of caching). We perform two evaluation runs:

1. First, we evaluate nodes only depending on current request/token/subscripton parameters. If no nodes in the graph reference parameter lookups, the graph is fully resolved at this point and we can return all instantiations as a static querier.
2. Otherwise, we return a dynamic querier. We can clone the partial static instantiation since parameters don't change within a connection. Then, to resolve buckets dynamically, we traverse lookup stages and replace lookups with their resolved instantiation until we've completed the graph and end up with an instantiation.

Some notes:

1. Step one could be split into two, one to evaluate parameters only depending on global request and token parameters and a second one to instantiate with subscription parameters. That might reduce some work for streams with many subscriptions, but I don't expect that to have a large impact.
2. We currently can't cache anything from step 2, any change in parameter lookups requires re-running everything. In the future, we might want to consider a querier interface that dispatches which lookups specifically have changed when `queryDynamicBucketDescriptions` is called. This would allow us to do fine-grained invalidation and potentially improve performance.

### DIFF
--- a/packages/sync-rules/src/compiler/sqlite.ts
+++ b/packages/sync-rules/src/compiler/sqlite.ts
@@ -247,8 +247,9 @@ export class PostgresToSqlite {
         let rightHandSideOfIs: SqlExpression<ExpressionInput>;
 
         switch (expr.op) {
-          case '+':
           case '-':
+            return this.invalidExpression(expr, 'Unary minus is not currently supported');
+          case '+':
             return { type: 'unary', operator: expr.op, operand: this.translateNodeWithLocation(expr.operand) };
           case 'NOT':
             return { type: 'unary', operator: 'not', operand: this.translateNodeWithLocation(expr.operand) };
--- a/packages/sync-rules/src/sync_plan/engine/javascript.ts
+++ b/packages/sync-rules/src/sync_plan/engine/javascript.ts
@@ -151,9 +151,9 @@ class ExpressionToJavaScriptFunction
         return operand;
       case 'not':
         return (input) => sqliteNot(operand(input));
-      case '~':
-      case '-':
-        throw new Error(`unary operator not supported: ${expr.operator}`);
+      // case '~':
+      // case '-':
+      //   throw new Error(`unary operator not supported: ${expr.operator}`);
     }
   }
 
@@ -173,10 +173,13 @@ class ExpressionToJavaScriptFunction
     return (input) => {
       const evaluatedValue = value(input);
 
-      return (
-        sqliteBool(evaluateOperator('>=', evaluatedValue, low(input))) &&
-        sqliteBool(evaluateOperator('<=', evaluatedValue, high(input)))
-      );
+      const geqLow = evaluateOperator('>=', evaluatedValue, low(input));
+      const leqHigh = evaluateOperator('<=', evaluatedValue, high(input));
+      if (geqLow == null || leqHigh == null) {
+        return null;
+      }
+
+      return sqliteBool(geqLow) && sqliteBool(leqHigh);
     };
   }
 
--- a/packages/sync-rules/src/sync_plan/expression.ts
+++ b/packages/sync-rules/src/sync_plan/expression.ts
@@ -26,7 +26,7 @@ export type SqlExpression<Data> =
  */
 export type ExternalData<Data> = { type: 'data'; source: Data };
 
-export type UnaryOperator = 'not' | '~' | '+' | '-';
+export type UnaryOperator = 'not' | '+'; //| '-' | '~';
 
 export type UnaryExpression<Data> = {
   type: 'unary';
--- a/packages/sync-rules/src/sync_plan/expression_to_sql.ts
+++ b/packages/sync-rules/src/sync_plan/expression_to_sql.ts
@@ -221,7 +221,7 @@ const binaryPrecedence: Record<BinaryOperator, Precedence> = {
 
 const unaryPrecedence: Record<UnaryOperator, Precedence> = {
   not: Precedence.not,
-  '~': Precedence.unary,
-  '+': Precedence.unary,
-  '-': Precedence.unary
+  //'~': Precedence.unary,
+  '+': Precedence.unary
+  //'-': Precedence.unary
 };
--- a/packages/sync-rules/test/src/sync_plan/engine/engine.test.ts
+++ b/packages/sync-rules/test/src/sync_plan/engine/engine.test.ts
@@ -3,6 +3,7 @@ import * as sqlite from 'node:sqlite';
 import { describe, expect, onTestFinished, test } from 'vitest';
 import {
   CompatibilityContext,
+  CompatibilityEdition,
   javaScriptExpressionEngine,
   nodeSqliteExpressionEngine,
   SqliteValue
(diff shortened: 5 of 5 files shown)


packages/sync-rules/src/compiler/sqlite.ts
rkistner: Discussed offline - probably better to remove the jsonb* functions until we have a more concrete use case.

rkistner: With LIKE we need to be clear about:
1. Case sensitivity.
2. Support for changing the escape character.
3. ICU support.

I'm happy if we go with the simplest/default version and these are not configurable - we just need to be clear on what we support.

rkistner: Discussed offline: We also need to make sure to support all the current functions as documented [here](https://docs.powersync.com/sync/rules/supported-sql#functions).

Datetime functions could be tricky: The current version supports a very limited subset of functionality. But most notably - `'now'` values can get tricky, and we can't easily remove that from SQLite's own functions if we expose those. So options I can think of:
1. Re-create the same behavior we currently have as new user-defined functions.
2. Similar, but use the underlying sqlite functions for the implementation, and just explicitly disallow `'now'` values.
3. Just allow `'now'` values (expose the sqlite functions as-is), and just document the expected behavior, i.e. that depending on where you use it in the query, the function could be evaluated at replication time or at sync time.

rkistner: `likely`, `unlikely` and `likelihood` would probably not be useful in sync rules/stream.s

simolus3: Removed them on the `sync-compiler-expressions` branch.

simolus3: Removed them on the `sync-compiler-expressions` branch.

simolus3: Added a note for this on `sync-compiler-expressions`.

> Re-create the same behavior we currently have as new user-defined functions.

I think this is probably the easiest solution. Explicitly disallowing `now` statically is not that easy since it could be a value from a table. So we'd need a wrapper function to throw if `now` is the first parameter, and at that point we might as well use our existing implementation to guarantee compatibility.

simolus3: I think it might even make sense to remove `LIKE` support initially since we don't seem to support it for sync definitions today.

If we later feel confident that we'll only use SQLite to evaluate expressions and can remove our JS implementation, we can still add support for it later.

simolus3: I've added an error on `LIKE` expressions for now, and support has been removed from known functions.

packages/sync-rules/src/sync_plan/engine/javascript.ts
```diff
@@ -0,0 +1,280 @@
...
+    const operand = this.compile(expr.operand);
+
+    switch (expr.operator) {
+      case '+':
+        return operand;
+      case 'not':
+        return (input) => sqliteNot(operand(input));
+      case '~':
+      case '-':
+        throw new Error(`unary operator not supported: ${expr.operator}`);
```
rkistner: Is it an issue that these are supported in the sqlite engine but not this one?

packages/sync-rules/src/sync_plan/engine/scalar_expression_engine.ts
```diff
@@ -0,0 +1,152 @@
...
+      first = false;
+    });
+  }
+
+  if (filters.length != 0) {
+    toSqlite.addLexeme('WHERE');
+    filters.forEach((expr, i) => {
+      if (i != 0) toSqlite.comma();
+      visitExpr(toSqlite, expr, null);
+    });
```
rkistner, line 89: What is the expected behavior when multiple filters are provided? The comma here would give a syntax error - should probably be AND?

packages/sync-rules/src/sync_plan/engine/javascript.ts
```diff
@@ -0,0 +1,280 @@
...
+    const value = this.compile(expr.value);
+
+    return (input) => {
+      const evaluatedValue = value(input);
+
+      return (
+        sqliteBool(evaluateOperator('>=', evaluatedValue, low(input))) &&
+        sqliteBool(evaluateOperator('<=', evaluatedValue, high(input)))
+      );
+    };
```
rkistner, line 183: This has differences in null handling compared to the SQLite version.

packages/sync-rules/src/sync_plan/engine/sqlite.ts
```diff
@@ -0,0 +1,64 @@
...
+      return impl.call(...args);
+    });
+  }
+
+  // Needed to make them deterministic / prevent passing 'now'
+  registerPowerSyncFunction('unixepoch');
+  registerPowerSyncFunction('datetime');
+
+  registerPowerSyncFunction('ST_AsGeoJSON');
+  registerPowerSyncFunction('AS_AsText');
```
rkistner: Should this be `ST_AsText`?

packages/sync-rules/src/sync_plan/engine/sqlite.ts
```diff
@@ -0,0 +1,64 @@
...
+  registerPowerSyncFunction('ST_X');
+  registerPowerSyncFunction('ST_Y');
```
rkistner: These should all be lowercase, matching the definitions from `generateSqlFunctions(compatibility)`

packages/sync-rules/src/sync_plan/engine/sqlite.ts
```diff
@@ -0,0 +1,64 @@
...
+  if (!compatibility.isEnabled(CompatibilityOption.fixedJsonExtract)) {
+    // For backwards compatibility, use the old JSON operators which parse the path argument differently.
+    db.function('->', { useBigIntArguments: true, varargs: true, deterministic: true }, (...args) => {
+      return functions.operatorJsonExtractJson.call(...args);
+    });
+
+    db.function('->>', { useBigIntArguments: true, varargs: true, deterministic: true }, (...args) => {
+      return functions.operatorJsonExtractSql.call(...args);
+    });
+  }
```
rkistner, line 48: Just to confirm - these functions would be limited to the specific db connection, and therefore the specific evaluator? Just need to make sure we can use multiple sync rules versions with different compatibility options concurrently.

simolus3: Exactly. I've also added a test using both compatibility options to ensure they're scoped to the engine in use.

packages/sync-rules/src/sync_plan/engine/javascript.ts
simolus3: It would lead to subpar error messages when used. I've just removed these unary operators from the sync plan for now.

packages/sync-rules/src/sync_plan/engine/javascript.ts
simolus3: It's worth noting that we have a few such differences in our implementation, e.g. `NOT NULL` is a fairly trivial one that evaluates to `1n` in JavaScript and `NULL` in SQLite.

I've fixed the between operator one though.

PR #467, opened 2026-01-15
Chore/replication replicas

EspressoTrip-v2:
When using two replication pods, the pod waiting for lock will continuously log:
`{"code":"PSYNC_S1003","description":"Sync rules: 3 have been locked by another process for replication, expiring at 2026-01-15T08:32:12.270Z.","name":"ServiceError"}` 

The change only allows it to log once, it will then output an info log when it regains the lock.


modules/module-mongodb-storage/src/storage/implementation/MongoSyncRulesLock.ts
```diff
@@ -36,17 +37,28 @@ export class MongoSyncRulesLock implements storage.ReplicationLock {
...
-        );
+        if (!this.lockAlerted) {
+          this.lockAlerted = true;
+          throw new ServiceError(
+            ErrorCode.PSYNC_S1003,
+            `Sync rules: ${sync_rules.id} have been locked by another process for replication, expiring at ${heldLock.lock.expires_at.toISOString()}.`
+          );
+        } else {
+          throw undefined;
+        }
```
rkistner: Rather keep the error throwing logic in this file unchanged, and only add a special case for this error where it is logged - throwing `undefined` can get confusing very quickly.

If needed, this can be changed to use a new subclass of ServiceError. Alternatively, just check the error code when logging the error.

packages/service-core/src/replication/AbstractReplicator.ts
```diff
@@ -225,12 +225,16 @@ export abstract class AbstractReplicator<T extends AbstractReplicationJob = Abst
...
-          // When new (valid) sync rules are deployed and processed, this one be disabled.
-          this.logger.error('Failed to start replication for new sync rules', e);
+          if (!this.lockAlerted) {
+            // Could be a sync rules parse error,
+            // for example from stricter validation that was added.
+            // This will be retried every couple of seconds.
+            // When new (valid) sync rules are deployed and processed, this one be disabled.
+            this.logger.error('Failed to start replication for new sync rules', e);
+            this.lockAlerted = true;
+          }
```
rkistner: 1. We should only silence the logs when it is a lock issue - not other errors (for other errors we could look into "rate limiting" the logs if the same error repeats, but that's another project).
2. For the lock error, we should log it as an informational message, not an error. Something like "[PSYNC_S1003] Replication is performed by another process, waiting until lock is released".

packages/service-core/src/replication/AbstractReplicator.ts
```diff
@@ -225,12 +225,20 @@ export abstract class AbstractReplicator<T extends AbstractReplicationJob = Abst
...
             activeJob = newJob;
           }
+          this.lockAlerted = false;
         } catch (e) {
-          // Could be a sync rules parse error,
-          // for example from stricter validation that was added.
-          // This will be retried every couple of seconds.
-          // When new (valid) sync rules are deployed and processed, this one be disabled.
-          this.logger.error('Failed to start replication for new sync rules', e);
+          if (e?.errorData.code === ErrorCode.PSYNC_S1003) {
```
rkistner:
```suggestion
          if (e?.errorData?.code === ErrorCode.PSYNC_S1003) {
```

Not all errors will have `errorData` - this change should cover it.

PR #470, opened 2026-01-19
Infer parameter lookups dynamically

simolus3:
This removes the `parameterQueryLookups` property from `BucketParameterQuerier`. That field was a static array of lookups the querier might use in `queryDynamicBucketDescriptions()`. By having that information available, we can re-use parameter lookups between checkpoints not affecting those parameters.

We want to avoid that field to prepare for nested lookups: If the output of a lookup is used as an input to another lookup, we can't declare it as a dependency ahead of time. So, this refactors `BucketParameterState` to fetch used parameters from the actual lookup invocations instead. To avoid serializing lookups multiple times, the `JSONBig`-serialized form of lookups is cached on the instance.

This is covered by existing tests, in particular the "dynamic buckets with updates" test in `BucketChecksumState.test.ts`. It's difficult to add tests for the enabled behavior of chained lookups since `BucketChecksumState.test.ts` uses real sync rules as queriers and we don't have a scenario where this is used yet.

### DIFF
--- a/packages/sync-rules/src/BucketParameterQuerier.ts
+++ b/packages/sync-rules/src/BucketParameterQuerier.ts
@@ -22,13 +22,12 @@ export interface BucketParameterQuerier {
   /**
    * True if there are dynamic buckets, meaning queryDynamicBucketDescriptions() should be used.
    *
-   * If this is false, queryDynamicBucketDescriptions() will always return an empty array,
-   * and parameterQueryLookups.length == 0.
+   * If this is false, queryDynamicBucketDescriptions() will always return an empty array.
    */
   readonly hasDynamicBuckets: boolean;
 
   /**
-   * These buckets depend on parameter storage, and needs to be retrieved dynamically for each checkpoint.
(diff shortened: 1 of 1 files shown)


rkistner (commented):
I'm trying to confirm the logic here. For simplicity, lets say these lookups are just a map of `lookup key => lookup value`.

So before, we had:

> These are the constant _lookup keys_ used for this connection. If the _lookup value_ produced by any of these lookups change, re-run the the parameter queries from scratch, and check all buckets of updates. The `update` gives us a list of lookup keys for which the lookup value may have changed.

Now, this prepares for a list of lookups keys that is not static. So we have:

> These are the lookup keys used by the last checkpoint. If the lookup value produced by any of these lookup keys change, re-run the the parameter queries from scratch, refreshing the list of keys, and check all buckets of updates.

What I'm missing here, is how do we detect whether the list of lookup keys may change? Would the change always be triggered by another lookup value changing, from a recursive lookup? In that case the logic here appears correct, I just want to confirm that.

PR #471, opened 2026-01-20
Sync streams compiler: Refactor expression handling

simolus3:
In the original compiler PR (https://github.com/powersync-ja/powersync-service/pull/451), expressions were represented via the `Expr` type from the `pgsql-ast-parser` package. That made parsing expressions trivial, but also had some downsides:

- we had to use hacks to attach scopes to references. In particular when transforming subqueries to joins (which would change scopes syntactically) we had to remember to resolve hoisted references in their original subquery scope. Now we can use our own expression tree where expressions are already resolved and can be rearranged without patching references.
- we need to do some validation for expressions, e.g. to ensure no illegal function is called. Having distinct types for original and validated expressions makes that clearer.
- we support syntax like `a IN (x, y, z)` that can't be represented in Postgres, which required hacks. With our own expression types we can add everything we need.
- by using our own serializable expression types, we can store expressions in sync plans. This makes it possible to implement type-aware schema generation and schema validation for serialized sync plans.
- the expression format is also easier to compose. External references where previously represented as SQL text + array of instantiations (e.g. `WHERE length(?) > 5`, `[row.name]`). Now, we can directly embed those references as an AST node (`WHERE length(<embed row.name>) > 5`), which means two ASTs can be combined without having to transform an instantiation array.
- because we're not just storing SQL text in sync plans, we can keep using our existing implementation for operators and functions if we want to.

This introduces an `SqlExpression` type with a generic type referencing external data (e.g. row values or connection parameters) and a visitor interface. The existing translation from Postgres to SQL strings has been split into two: First, we lower Postgres expressions to SQLite and desugar subqueries in one go. These SQLite expression nodes are only rendered into SQL text by a second iteration when necessary.

### DIFF
--- a/packages/sync-rules/src/sync_plan/expression.ts
+++ b/packages/sync-rules/src/sync_plan/expression.ts
@@ -146,9 +146,11 @@ export const supportedFunctions: Record<string, ArgumentCount> = {
   iif: { min: 2 },
   instr: 2,
   length: 1,
+  // TODO: Establish defaults for case sensitivity, changing escape characters, ICU support.
+  // We might just want to remove LIKE support since we don't seem to have it in sql_functions.ts
   like: { min: 2, max: 3 },
-  likelihood: 2,
-  likely: 1,
+  //  likelihood: 2,
+  //  likely: 1,
   lower: 1,
   ltrim: { min: 1, max: 2 },
   max: { min: 2 },
@@ -169,7 +171,7 @@ export const supportedFunctions: Record<string, ArgumentCount> = {
   unicode: 1,
   unistr: 1,
   unistr_quote: 1,
-  unlikely: 1,
+  //  unlikely: 1,
   upper: 1,
   zeroblob: 1,
   // Scalar functions from https://sqlite.org/json1.html#overview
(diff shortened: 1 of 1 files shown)


packages/sync-rules/test/src/compiler/sqlite.test.ts
```diff
@@ -26,8 +28,10 @@ describe('sqlite conversion', () => {
   });
 
   test('in values', () => {
-    expectNoErrors('"intrinsic:contains"(1)', 'FALSE');
-    expectNoErrors('"intrinsic:contains"(1, 2, 3, 4)', '1 IN (2, 3, 4)');
+    expectNoErrors('1 IN ARRAY[1,2,3]', '1 IN (1, 2, 3)');
+    expectNoErrors('1 IN ARRAY[]', 'FALSE');
```
rkistner, line 32: At what point in the process this evaluated to false? That looks like more than a simple conversion

simolus3: This happens here: https://github.com/powersync-ja/powersync-service/blob/16f449d3f3d6e8c485a54ef7adbbc7fd5d7f4c2e/packages/sync-rules/src/sync_plan/expression_to_sql.ts#L113-L117

In the sync plan, this would be represented as `{ type: 'scalar_in', target: {/* literal 1 */}, in: [] }`. The direct translation in SQLite is `1 IN ()`, but that is a syntax error. So this test is there to ensure we don't generate invalid syntax, I don't think anyone would actually write `1 IN ARRAY[]` in a sync stream since it's trivially false.

PR #473, opened 2026-01-23
Compiler: Support table-valued functions, `&&` operator

simolus3:
The new sync stream compiler supports table-valued functions in `FROM` clauses as long as these functions only use inputs that can be derived from the current connection. This is enough to support the equivalent of table-valued parameter queries from bucket definitions. However, it is not enough to support the  `&&` (JSON array overlap) operator supported by the old sync streams implementation.

This expands support for table-valued functions, and uses that to implement `&&`.

### Table-valued functions

Table-valued functions like `json_each` now support having a synced table or parameter table as inputs. To support that, the compiled sync plan is expanded as follows:

1. `TableProcessor`s (bucket data sources and parameter index lookup creators) now have a `tableValuedFunctions` property describing which table-valued functions to add as inner joins when evaluating a source row. For each function, that describes the function name (e.g. `json_each`), input expressions derived from the source row and filters.
2. Parameter values (represented as the `PartitionKey` interface) are allowed to reference outputs from joined table-valued functions.
3. For parameter queries, outputs are also allowed to reference outputs from table-valued functions.

A few examples might help to demonstrate how table-valued functions work and how they're implemented:

#### Example 1: Static filter

Consider the stream `SELECT * FROM posts WHERE 'important' IN posts.descriptions`. As usual with sync streams, the `IN` is desugared to `SELECT posts.* FROM posts, json_each(posts.descriptions) AS subquery0 WHERE subquery0.value = 'important'`.

In `querier_graph.ts`, we resolve table-valued functions along with their input table (`posts` int this case). So this stream compiles into a single `TableProcessor` with no parameters and:

- `tableValuedFunctions: [{name: 'json_each', inputs: [/* reference row.descriptions */], filters: []}]`
- `filters: [{/* literal('important') == tableValuedOutput(functionIndex: 0, column: 'value') */}]`.

#### Example 2: Partition on data

The stream `SELECT * FROM posts WHERE subscription.parameter('tag') IN (SELECT value FROM json_each(posts.tags))` compiles to a table processor with no filters and:

- `tableValuedFunctions: [{name: 'json_each', inputs: [/* reference row.tags */], filters: []}]`
- `parameters: [/* tableValuedOutput(functionIndex: 0, column: 'value') */]`

For connections, the parameter is instantiated with `subscription.parameter('tag')` as usual.

#### Example 3: Filter on array values

This query is a bit exotic, but we support `SELECT * FROM customers, json_each(customers.active_regions) AS regions WHERE region.value LIKE 'eu-%'`. This translates into a table-valued function like

- `tableValuedFunctions: [{name: 'json_each', inputs: [/* reference row.active_regions */], filters: [/* reference row.value LIKE literal('eu-%') */]}]`

Even though the table-valued function isn't used, it still needs to be evaluated to skip customers without a matching element in the `active_regions` array.

#### Example 4: Partition on parameters

Partitions based on table-valued functions are also supported on parameter creators: `SELECT users.* FROM users, orgs, json_each(orgs.members) as members WHERE users.id = members.value AND orgs.id = auth.parameter('org')`. This creates a parameter lookup creator with a table-valued function and an output column referencing that function.

### The `&&` operator

Similar to how the `IN` operator is desugared into a joined subquery and an equals clause, we do the exact same thing for `&&`. The only difference is that while `IN` only expands the right-hand side, `&&` expands both. `&&` is consistently lowered into `=` and not considered after parsing queries.

### DIFF
--- a/packages/sync-rules/test/src/sync_plan/evaluator/evaluator.test.ts
+++ b/packages/sync-rules/test/src/sync_plan/evaluator/evaluator.test.ts
@@ -2,7 +2,6 @@ import { describe, expect } from 'vitest';
 import { syncTest } from './utils.js';
 import {
   HydratedSyncRules,
-  RequestParameters,
   ScopedParameterLookup,
   SourceTableInterface,
   SqliteJsonRow,
@@ -392,7 +391,7 @@ streams:
 `);
 
     const { querier, errors } = desc.getBucketParameterQuerier({
-      globalParameters: new RequestParameters({ sub: 'user', a: ['a1', 'a2'], b: ['b1', 'b2'] }, {}),
+      globalParameters: requestParameters({ sub: 'user', a: ['a1', 'a2'], b: ['b1', 'b2'] }, {}),
       hasDefaultStreams: true,
       streams: {}
     });
--- a/packages/sync-rules/test/src/sync_plan/evaluator/table_valued.test.ts
+++ b/packages/sync-rules/test/src/sync_plan/evaluator/table_valued.test.ts
@@ -1,7 +1,7 @@
 import { describe, expect } from 'vitest';
 import { syncTest } from './utils.js';
-import { TestSourceTable } from '../../util.js';
-import { RequestParameters, ScopedParameterLookup, SqliteJsonRow } from '../../../../src/index.js';
+import { requestParameters, TestSourceTable } from '../../util.js';
+import { ScopedParameterLookup, SqliteJsonRow } from '../../../../src/index.js';
 
 describe('table-valued functions', () => {
   syncTest('as partition key', ({ sync }) => {
@@ -64,7 +64,7 @@ streams:
     ]);
 
     const { querier } = desc.getBucketParameterQuerier({
-      globalParameters: new RequestParameters({ sub: 'user' }, {}),
+      globalParameters: requestParameters({ sub: 'user' }, {}),
       hasDefaultStreams: false,
       streams: {
         stream: [


packages/sync-rules/src/sync_plan/evaluator/parameter_index_lookup_creator.ts
```diff
@@ -21,20 +23,22 @@ export class PreparedParameterIndexLookupCreator implements ParameterIndexLookup
...
     for (const parameter of source.parameters) {
-      expressions.push(mapExpressions.transform(parameter.expr));
+      expressions.push(translationHelper.mapper.transform(parameter.expr));
     }
     this.numberOfParameters = source.parameters.length;
 
     this.evaluator = engine.prepareEvaluator({
       outputs: expressions,
-      filters: source.filters.map((f) => mapExpressions.transform(f))
+      filters: filterExpressions,
```
rkistner: The `filterExpressions` array is never populated - should this be using `translationHelper.filterExpressions` instead?

Failing test:

```ts
  syncTest('parameter index lookup respects filters', ({ sync }) => {
    const desc = sync.prepareSyncStreams([
      {
        name: 'stream',
        queries: [
          `SELECT users.* FROM users, orgs WHERE users.org_id = orgs.id AND orgs.name = subscription.parameter('org') AND orgs.is_active = 1`
        ]
      }
    ]);

    const orgs = new TestSourceTable('orgs');

    const active = desc.evaluateParameterRow(orgs, { id: 'a', name: 'org-a', is_active: 1 });
    const inactive = desc.evaluateParameterRow(orgs, { id: 'b', name: 'org-b', is_active: 0 });

    expect(active.length).toBe(1);
    expect(inactive.length).toBe(0); // Fails on this PR
  });
```

packages/sync-rules/src/sync_plan/evaluator/table_processor_to_sql.ts
```diff
@@ -0,0 +1,41 @@
...
+    for (const fn of source.tableValuedFunctions) {
+      const mapped: TableValuedFunction = {
+        name: fn.functionName,
+        inputs: fn.functionInputs.map((i) => this.mapper.transformWithoutTableValued(i))
+      };
+      this.mapper.tableValuedFunctions.set(fn, mapped);
+
+      for (const filter of fn.filters) {
+        this.filterExpressions.push(this.mapper.transform(filter));
+      }
```
rkistner, line 43: This appears to use values from the source row, rather than the table valued function output.

Potential fix (not sure if this is how things are supposed to be used):

```suggestion
      const mapFnFilter = new MapSourceVisitor<plan.ColumnSqlParameterValue, TableValuedFunctionOutput>(
        ({ column }) => ({ function: mapped, column })
      );

      for (const filter of fn.filters) {
        this.filterExpressions.push(visitExpr(mapFnFilter, filter, null));
      }
```

Failing test:

```ts
  syncTest('table-valued function filters reference function output', ({ sync }) => {
    const desc = sync.prepareSyncStreams([
      {
        name: 'stream',
        queries: [
          `SELECT customers.id as id FROM customers, json_each(customers.active_regions) AS region WHERE region.value < 'm'`
        ]
      }
    ]);

    const customers = new TestSourceTable('customers');
    const rows = desc.evaluateRow({
      sourceTable: customers,
      record: { id: 'c1', active_regions: JSON.stringify(['us-east', 'eu-west']) }
    });

    expect(rows).toMatchObject([{ bucket: 'stream|0[]', data: { id: 'c1' }, id: 'c1', table: 'customers' }]); // Fails on this PR
  });
```

rkistner:
@simolus3 Looks like there is a merge conflict with some changes I recently made, affecting tests

rkistner (approved):

PR #475, opened 2026-01-28
feat: public api client

stevensJourney:
# Overview

When started with the `RouterEngine` enabled - the PowerSync service currently exposes HTTP routes  which are used by the PowerSync management service (our cloud services orchestrator) to perform orchestrating tasks. 

These routes can also be useful for self-hosting tooling. 

This PR adds a client SDK for these routes. This is based of the Journey Micro tooling which is now publicly available from [JourneyApps labs](https://github.com/journeyapps-labs).

### DIFF
--- a/.changeset/modern-pugs-kiss.md
+++ b/.changeset/modern-pugs-kiss.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-client': patch
+---
+
+Trusted publishing release


packages/service-client/src/client.ts
```diff
@@ -0,0 +1,196 @@
...
+   * //   results: {
+   * //     columns: ['?column?'],
+   * //     rows: [[1]]
+   * //   }
+   * // }
+   * ```
+   */
+  executeSql = this.createEndpoint<internal_routes.ExecuteSqlRequest, internal_routes.ExecuteSqlResponse>({
+    path: '/api/admin/v1/execute-sql'
+  });
```
rkistner: I think we should exclude this one, and deprecate & remove the API eventually. The others look fine.

rkistner (dismissed):

rkistner (approved):

PR #478, opened 2026-01-31
feat(test-client): support RS256 and configurable aud/exp in generate…

imbrahiam:
## Summary

This PR adds RS256 support to the `generate-token` command and improves
JWT generation ergonomics by allowing configurable audience and expiration.

## Changes

- Support RS256 signing keys alongside HS256
- Validate private RSA keys before attempting to sign
- Pass algorithm explicitly to `jose.importJWK`
- Add `--aud` (repeatable) and `--exp` CLI options
- Keep existing behavior fully backward-compatible

## Motivation

The documentation suggests RS256 is supported, but the test-client was
previously hard-coded to HS256. This change aligns implementation with
documentation and enables realistic authentication testing.

## Testing

- Generated tokens using HS256 and RS256 keys
- Verified RS256 tokens against public JWKS configuration
- Confirmed existing HS256 workflows remain unchanged

### DIFF
--- a/test-client/src/auth.ts
+++ b/test-client/src/auth.ts
@@ -7,6 +7,8 @@ export interface CredentialsOptions {
   endpoint?: string;
   config?: string;
   sub?: string;
+  aud?: string[];
+  exp?: string;
 }
 
 export interface Credentials {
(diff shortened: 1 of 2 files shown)


test-client/src/auth.ts
```diff
@@ -18,48 +20,65 @@ export async function getCredentials(options: CredentialsOptions): Promise<Crede
...
+  let endpoint = options.endpoint;
+  if (endpoint == null) {
+    endpoint = `http://127.0.0.1:${parsed.port ?? 8080}`;
+  }
+
+  const aud = [
+    ...(parsed.client_auth?.audience ?? []),
+    ...(options.aud ?? []),
+    endpoint
+  ].filter(Boolean);
```
simolus3, line 64: This is clever but I think the original `(a) => a != null` was a bit easier to read :D

test-client/src/auth.ts
```diff
@@ -18,48 +20,65 @@ export async function getCredentials(options: CredentialsOptions): Promise<Crede
...
+    })
+    .setSubject(sub)
+    .setIssuedAt()
+    .setIssuer('test-client')
+    .setAudience(aud)
+    .setExpirationTime(exp)
+    .sign(key);
+
+  return { token, endpoint };
+}
```
simolus3, line 84: Since this file is missing a trailing newline: Could you run `pnpm format` in the workspace directory (or `pnpm exec prettier --write .` in `test-client`) to apply formatting?

simolus3 (commented):
These changes look good to me apart from two minor comments. Could you also run `pnpm -w changeset add` to add a changeset entry (which we use to auto-release packages)?

PR #479, opened 2026-02-02
Fix truncate on Postgres storage

rkistner:
When we truncate a table, we need to find all the existing rows we have in that table, and create relevant REMOVE operations for each.

A bug in the Postgres storage implementation caused the wrong `source_key` value to be used for these. This causes the client to see it as an op for a new row, instead of removing the one already synced.

While we had tests for many scenarios requiring a truncate, it only checked for the presence of REMOVE operations, and not that the source_key / subkey matched. This now adds tests using the `reduceBucket()` function. This matches the logic on the client, producing the final output. So while this does not give us the intermediate state, it allows us to accurately test the final output as the client would see it.

For the storage itself, the fix is simple: We need to deserialize the binary source_key before using it.

If the issue was already triggered, this fix does will not remove the data already orphaned - re-replicating is required to recover.

## Impact

If the bug is triggered, it could result in "orphaned" data being synced, and remaining on devices. Re-syncing a device would not resolve the issue. Re-replicating from scratch (for example by making a sync rules change) is the only thing that would reliably resolve the issue.

The issue could be triggered if Postgres storage is used, and:
1. A table is dropped or truncated.
2. A table is renamed.
3. The replica identity of a table is changed.

This can effect any source database - Postgres, MongoDB, MySQL or SQL Server.

MongoDB storage is not affected.

I am not aware of any reports of this happening in practice, but it could be that users just recovered by making sync rule changes.

### DIFF
--- a/.changeset/brown-turtles-repair.md
+++ b/.changeset/brown-turtles-repair.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-module-postgres-storage': patch
+---
+
+[Postgres Storage] Fix data being orphaned after truncating table.


stevensJourney (approved):
Looks good to me. thanks for fixing this!

PR #487, opened 2026-02-04
Storage version

rkistner:
This introduces a "storage_version" config, for both MongoDB and Postgres storage.

The basic idea is to move away from migrations that need to be run upfront, to a storage version that is specific to a sync rules version. So when upgrading the service version, there is no need to run large migrations on existing data. Instead, when you deploy a new sync rules version, the new storage version is used for that.

Pros:
1. We can make significant changes to large collections such as bucket_data, without running any expensive migrations upfront.
2. There is no conflict between service versions before/after the migration. For example, if the migration runs before running the new service version, the old version had to be compatible with the post-migration data, otherwise it would cause issues.
3. Storage version downgrades are possible, by first deploying sync rules with an older storage version (not implemented yet).

Cons:
1. We need to support older storage versions for a significant period. We can eventually deprecate old storage versions in new major releases.

This initial implementation uses the storage version for two features:
1. A guarantee that checksums are always stored as Long, which gives a small performance improvement for checksum calculations. See #282 for context on the original issue.
2. Auto-enabling versioned bucket names, instead of requiring an opt-in in the sync rules file.

These are fairly minor storage changes to start with. However, the plan is to use this for incremental reprocessing (#468), which may introduce much larger storage changes.

## Collections

There are no actual changes to collections here, but generally going forward, we'd have "static" collections (no change based on version), and "versioned" collections - where we'd use different collections based on the storage version, _if affected by the storage version_. This is not final - we can always introduce more changes based on storage_version, but this generally explains the expected changes.

Still TBD how we manage changes to static collections. For now we can keep on using migrations, but we may eventually replace that mechanism with something that can handle downgrades better.

Static collections:
1. migrations
2. instance
3. sync_rules (the actual fields used may change based on storage version)
4. locks
5. op_id_sequence (some of this usage may change based on storage version)
6. connection_report_events (TBD)

Versioned collections:
1. bucket_data
2. bucket_parameters
3. bucket_state
4. current_data
5. checkpoint_events
6. source_tables
7. write_checkpoints
8. custom_write_checkpoints

## Downgrades

A downgrade to a lower storage version would always need a sync rule reprocessing. In theory, this can happen before or after downgrading the service version (to one that doesn't support the latest storage version). So you generally have these options:

### 1. Don't upgrade storage version 

Upgrade service version, but don't upgrade storage version. Can downgrade without issues.

The caveat is that storage version upgrade is currently automatic when you update the sync rules - can't opt out of that yet. We can add support for opting out in the future.

### 2. Downgrade storage before downgrading service.

1. Upgrade service.
2. Update sync rules, triggering storage version upgrade.
3. Realize you want to downgrade.
4. Update sync rules, with lower storage version. _Not supported yet, but we can support this in the future_
5. Downgrade service, with no downtime.

### 3. Downgrade storage after downgrading service

1. Upgrade service.
2. Update sync rules, triggering storage version upgrade.
3. Downgrade service.
4. Storage version unsupported - replication and API processes fail, causes downtime.
5. Re-process sync rules, resulting in the lower storage version. _We can consider triggering this automatically, but have to think through the implications._
6. Once reprocessing completed, users can sync again.

### 4. Downgrading to lower migration version

_Just for reference_

Currently this blocks the migration process. Theoretically it's possible to down-migrate first using the newer service version, but I'm not sure that process actually works. Not supported at all on the cloud service at the moment - have to stop and start the instance, which re-creates the storage from scratch and causes downtime.

Depending on the actual migrations, there may be consistency issues in the process.

## General comments

There is remaining work to implement to make downgrading possible without downtime. We'd have to think how exactly we expose this - for example config options, or an API. Either approach would likely require documenting storage version compatibility.

However, the current state with storage is no worse than we have with the migration system.

## Test changes

This now runs most tests with both storage version 1 and 2. This affects bucket names, so the tests have some refactoring to cater for that.

### DIFF
--- a/modules/module-mongodb-storage/src/storage/implementation/MongoPersistedSyncRules.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/MongoPersistedSyncRules.ts
@@ -1,12 +1,14 @@
 import {
   CompatibilityOption,
+  DEFAULT_HYDRATION_STATE,
   HydratedSyncRules,
+  HydrationState,
   SyncConfigWithErrors,
   versionedHydrationState
 } from '@powersync/service-sync-rules';
 
 import { storage } from '@powersync/service-core';
-import { DEFAULT_HYDRATION_STATE, HydrationState } from '@powersync/service-sync-rules/src/HydrationState.js';
+
 import { StorageConfig } from './models.js';
 
 export class MongoPersistedSyncRules implements storage.PersistedSyncRules {
--- a/packages/service-core/src/sync/BucketChecksumState.ts
+++ b/packages/service-core/src/sync/BucketChecksumState.ts
@@ -1,8 +1,10 @@
 import {
   BucketDescription,
+  BucketParameterQuerier,
   BucketPriority,
   BucketSource,
   HydratedSyncRules,
+  QuerierError,
   RequestedStream,
   RequestParameters,
   ResolvedBucket
@@ -18,7 +20,6 @@ import {
   ServiceAssertionError,
   ServiceError
 } from '@powersync/lib-services-framework';
-import { BucketParameterQuerier, QuerierError } from '@powersync/service-sync-rules/src/BucketParameterQuerier.js';
 import { JwtPayload } from '../auth/JwtPayload.js';
 import { SyncContext } from './SyncContext.js';
 import { getIntersection, hasIntersection } from './util.js';


modules/module-mongodb-storage/src/storage/implementation/models.ts
```diff
@@ -204,8 +204,41 @@ export interface SyncRuleDocument {
     id: string;
     expires_at: Date;
   } | null;
+
+  storage_version?: number;
```
stevensJourney, line 208: I initially thought this might not need to be optional, due to it being set in the migrations. But I assume we can't really guarantee that all migrations have actually been executed in some circumstances - like self-hosted environments, or is there another reason for declaring it as optional?

modules/module-mongodb-storage/src/storage/implementation/models.ts
```diff
@@ -204,8 +204,41 @@ export interface SyncRuleDocument {
...
+
+  /**
+   * Whether versioned bucket names are automatically enabled.
+   *
+   * If this is false, bucket names may still be versioned depending on the sync config.
+   */
+  versionedBuckets: boolean;
+}
+
+export const LEGACY_STORAGE_VERSION = 1;
```
stevensJourney: Just for my understanding. It looks like the versions are a single integer, which we only increment for breaking changes? Would any non-breaking changes or additions to the storage still retain the same version?

modules/module-mongodb-storage/src/storage/implementation/models.ts
```diff
@@ -204,8 +204,41 @@ export interface SyncRuleDocument {
...
+export const CURRENT_STORAGE_VERSION = 2;
+
+export const STORAGE_VERSIONS: Record<number, StorageConfig | undefined> = {
```
stevensJourney: nit: It looks like this is mapping the storage version to the corresponding StorageConfig. Maybe we could call this `STORAGE_VERSION_CONFIGS`.

modules/module-mongodb-storage/src/storage/implementation/MongoPersistedSyncRulesContent.ts
```diff
@@ -34,14 +36,32 @@ export class MongoPersistedSyncRulesContent implements storage.PersistedSyncRule
...
+  /**
+   * Load the storage config.
+   *
+   * This may throw if the persisted storage version is not supported.
+   */
+  getStorageConfig() {
+    const storageConfig = STORAGE_VERSIONS[this.storage_version];
+    if (storageConfig == null) {
+      throw new ServiceError(
+        ErrorCode.PSYNC_S1403,
```
stevensJourney: What would the flow be if a user did downgrade the service and this has been reached?

Would we always recommend performing a sync rules change when downgrading the service? Otherwise, it seems like a downgrade would essentially take-down the instance for both replication and api services (if I understand this correctly)?

modules/module-mongodb-storage/src/storage/implementation/models.ts
rkistner: Initially this was from before having a migration. Now, it's mostly a case of we're not guaranteed that the migration has run, and having the fallback is simple to implement.

modules/module-mongodb-storage/src/storage/implementation/MongoPersistedSyncRulesContent.ts
rkistner: Correct - a downgrade would take down the instance. I feel that's better than attempting to continue, which could result in obscure errors or even silent consistency issues.

I added a section in the PR description on the available downgrade options.

packages/sync-rules/src/SyncConfig.ts
```diff
@@ -33,17 +33,12 @@ export abstract class SyncConfig {
   /**
    * Hydrate the sync rule definitions with persisted state into runnable sync rules.
    *
-   * @param params.hydrationState Transforms bucket ids based on persisted state. May omit for tests.
+   * @param params.hydrationState Transforms bucket ids based on persisted state.
```
simolus3, line 39: Maybe also mention that the compatibility option is not checked in this method.

packages/sync-rules/test/src/compatibility.test.ts
```diff
@@ -1,7 +1,7 @@
 import { describe, expect, test } from 'vitest';
 import { DateTimeValue, SqlSyncRules, TimeValuePrecision, toSyncRulesValue } from '../../src/index.js';
 
-import { versionedHydrationState } from '../../src/HydrationState.js';
+import { DEFAULT_HYDRATION_STATE, versionedHydrationState } from '../../src/HydrationState.js';
```
simolus3, line 12: We should remove `versioned_bucket_ids: false` from the `streams can disable new format` test and remove the `can use versioned bucket ids` test, these tests are no longer relevant now that the compatibility option is ignored.

Maybe we could replace them with tests asserting that the option is marked as enabled via `expect(rules.compatibility.isEnabled(CompatibilityOption.versionedBucketIds))` depending on YAML since we're using that property when loading sync rules from storage.

PR #490, opened 2026-02-05
[Fix]: Ensure JWT sub is a string

EspressoTrip-v2:
## Background
We previously assumed the JWT `sub` claim was always a string; however, it can also be a boolean, object, or array.

## Issue
- Client analytics fails when `sub` is not a string.
- Other code paths that rely on `context.user_id` or token payload string semantics may also be affected.

## Fix
Normalize `sub` immediately after JWT verification (in `KeyStore.verifyJwt`): keep strings as-is, use `JSON.stringify` for everything else, and use the normalized value for `context.user_id` and `token_payload`.

## Impact of the fix
- `request.user_id()` / `auth.user_id()` now consistently return a string.
- String `sub` values are unchanged; non-string `sub` values are normalized with `JSON.stringify`.
- Sync rules that expect string IDs continue to behave as before; rules that assume strict non-string typing should be reviewed.


packages/service-core/src/routes/auth.ts
```diff
@@ -47,17 +47,23 @@ export async function authorizeUser(context: Context, authHeader: string = ''):
...
     const maxAge = configuration.token_max_expiration;
-    tokenPayload = await configuration.client_keystore.verifyJwt(token, {
+    const parsedToken = await configuration.client_keystore.verifyJwt(token, {
       defaultAudiences: configuration.jwt_audiences,
       maxAge: maxAge
     });
+    const safeToken = {
+      ...parsedToken,
+      // Ensure the sub is a string as it can be a number
+      sub: parsedToken.sub.toString()
```
rkistner: `toString()` won't work well if the sub is a boolean, object or array. I recommend:
1. Keep strings as-is.
2. Use JSON.stringify for everything else.

Could you then also add tests for each of those cases?

Moving the logic into the verifyJwt method may help to make testing easier.

PR #492, opened 2026-02-09
Add `SyncConfig` superclass for sync configurations

simolus3:
This adds the abstract `SyncConfig` class which is extended by the existing `SqlSyncRules` and the new `PrecompiledSyncConfig` class. This then refactors places where `SqlSyncRules` have previously been used to now use the common `SyncConfig` class.

The idea here is that can keep `SqlSyncRules` unchanged, and then have a separate but compatible class representing a compiled sync plan generated by the new compiler. In the future, it will also enable bucket storage implementations to store sync plans and return a `PrecompiledSyncConfig` instance instead of recompiling to `SqlSyncRules` when a sync plan is available. 

Since precompiled sync plans don't have errors, I've moved the `errors` field to a separate interface (`SyncConfigWithErrors`).
`SyncConfig` follows the [updated terminology](https://github.com/powersync-ja/powersync-service/pull/451#issuecomment-3847359534) we want to use to describe both sync rules and sync streams going forward.

### DIFF
--- a/packages/sync-rules/src/sync_plan/evaluator/index.ts
+++ b/packages/sync-rules/src/sync_plan/evaluator/index.ts
@@ -47,8 +47,4 @@ export class PrecompiledSyncConfig extends SyncConfig {
       this.bucketSources.push(new StreamBucketSource(stream, streamInput));
     }
   }
-
-  protected asSyncConfig(): this {
-    return this;
-  }
 }


packages/sync-rules/src/schema-generators/SwiftSchemaGenerator.ts
```diff
@@ -9,7 +8,7 @@ export class SwiftSchemaGenerator extends SchemaGenerator {
   readonly mediaType = 'text/x-swift';
   readonly fileName = 'schema.swift';
 
-  generate(source: SqlSyncRules, schema: SourceSchema, options?: GenerateSchemaOptions): string {
+  generate(source: SyncConfig, schema: SourceSchema, options?: GenerateSchemaOptions): string {
```
stevensJourney, line 12: Nit: should this one also be `BaseSyncConfig`?

packages/sync-rules/src/schema-generators/TsSchemaGenerator.ts
```diff
@@ -47,7 +46,7 @@ export class TsSchemaGenerator extends SchemaGenerator {
     }
   }
 
-  generate(source: SqlSyncRules, schema: SourceSchema, options?: GenerateSchemaOptions): string {
+  generate(source: SyncConfig, schema: SourceSchema, options?: GenerateSchemaOptions): string {
```
stevensJourney, line 50: This one also `BaseSyncConfig`?

packages/sync-rules/src/BaseSyncConfig.ts
```diff
@@ -0,0 +1,126 @@
...
+   * The (YAML-based) source contents from which these sync rules have been derived.
+   */
+  content: string;
+
+  constructor(content: string) {
+    this.content = content;
+  }
+
+  // Ensure asSyncConfig is not implemented externally
+  protected abstract asSyncConfig(): SyncConfig & this;
```
stevensJourney: It took me a little bit to figure out what was going on here.

My understanding of the current layout is that we have `SqlSyncRules` and `PrecompiledSyncConfig` which extend from this `BaseSyncConfig` abstract class. Where `SyncConfig` is a union of `SqlSyncRules | PrecompiledSyncConfig`
Then, in `BaseSyncConfig::hydrate` we need to pass in `SyncConfig` to `HydratedSyncRules`'s constructor. Which seems to be why this base class returns an instance of its possible known/limited implementations. If that is the reasoning, I think the comment above could add a little more context.

Is this necessary? Could we pass `BaseSyncConfig` to `HydratedSyncRules`?

packages/sync-rules/src/SyncConfig.ts
```diff
@@ -0,0 +1,126 @@
...
+    this.content = content;
+  }
+
+  // Ensure asSyncConfig is not implemented externally
+  protected abstract asSyncConfig(): SyncConfig & this;
+
+  /**
+   * If this sync config is fully described by a serializable sync plan, returns that plan.
+   */
+  abstract extractSyncPlan(): SyncPlan | null;
```
stevensJourney: If I understand correctly - only `PrecompiledSyncConfig` will return a result from this method. It feels a little strange for the `BaseSyncConfig` class to cater for one of its possible implementations.
It doesn't look like this method is currently used in this PR... Would it be feasible for consuming code to do `instanceof` checks instead - and call `extractSyncPlan` from instances of `PrecompiledSyncConfig`?

packages/sync-rules/src/BaseSyncConfig.ts
simolus3: My original idea was to make `BaseSyncConfig` an implementation detail, and only allow two concrete instances (through the `SyncConfig` union type).

There's no strong benefit of that though, so I've removed the `SyncConfig` type and renamed `BaseSyncConfig` to `SyncConfig`. So now we get to use `SyncConfig` consistently.

packages/sync-rules/src/sync_plan/evaluator/index.ts
```diff
@@ -1,40 +1,54 @@
...
   }
 
-  const streamInput: StreamInput = {
-    ...context,
-    preparedBuckets,
-    preparedLookups
-  };
-  for (const stream of plan.streams) {
-    rules.bucketSources.push(new StreamBucketSource(stream, streamInput));
+  protected asSyncConfig(): this {
```
stevensJourney: This method can be removed now, I believe

PR #493, opened 2026-02-09
Make sync plans fully serializable

simolus3:
To be able to evolve the compiler without affecting compiled sync plans, it's important that a sync plan loaded from a serialized form behaves exactly as a sync plan compiled from YAML originally.

This is already mostly the case, but there are two remaining aspects that make this tricky:

1. We expect to be able to set the default schema when loading sync configurations from bucket storage, and there's no guarantee that the default schema used when compiling sync plans is the same as the one used to later restore load from a serialized form.
2. The behavior of sync plans depends on compatibility options, which needs to be serializable.

To fix the first issue, this introduces the `ImplicitSchemaTablePattern` class with an optional connection tag and schema. It is used in the compiler instead of applying the default. This allows the implicit pattern to be serialized, we apply the default later when loading the serialized plan.
For the second issue, this introduces methods on `CompatibilityContext` allowing it to be serialized and deserialized.

### DIFF
--- a/packages/sync-rules/src/compatibility.ts
+++ b/packages/sync-rules/src/compatibility.ts
@@ -129,7 +129,7 @@ export class CompatibilityContext {
     }
 
     let maxTimeValuePrecision: TimeValuePrecision | undefined;
-    if (serialized.maxTimeValuePrecision) {
+    if (serialized.maxTimeValuePrecision != null) {
       for (const option of Object.values(TimeValuePrecision.byName)) {
         if (option.subSecondDigits == serialized.maxTimeValuePrecision) {
           maxTimeValuePrecision = option;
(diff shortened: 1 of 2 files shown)


packages/sync-rules/src/compatibility.ts
```diff
@@ -105,6 +105,46 @@ export class CompatibilityContext {
...
+    const overrides = new Map<CompatibilityOption, boolean>();
+    for (const [option, enabled] of Object.entries(serialized.overrides)) {
+      const knownOption = CompatibilityOption.byName[option];
+      if (knownOption) {
+        overrides.set(knownOption, enabled);
+      }
+    }
+
+    let maxTimeValuePrecision: TimeValuePrecision | undefined;
+    if (serialized.maxTimeValuePrecision) {
```
rkistner: Picked up by Codex: If `serialized.maxTimeValuePrecision == 0`, this logic won't be triggered correctly.

```suggestion
    if (serialized.maxTimeValuePrecision != null) {
```

PR #496, opened 2026-02-09
feat: export JSON schema from @powersync/service-schema

stevensJourney:
We currently include a JSON schema for the PowerSync service configuration. This schema can be used for type checking and autocomplete when editing PowerSync self hosted YAML configuration files.

This version of the configuration merges all possible combinations of different module Replication connection and bucket storage types.

We don't currently export any other version of a merged configuration typing. This PR updates the existing package to additionally explicitly export the JSON schema.

I've also added the MSSSQL connection type to the exported merged schema.

### DIFF
--- a/packages/schema/src/index.ts
+++ b/packages/schema/src/index.ts
@@ -1,1 +1,1 @@
-// This project does not yet include TypeScript source, but it does use TypeScript to depend on the build off @powersync/service-types.
+// This project does not yet include TypeScript source, but it does use TypeScript to depend on the build of @powersync/service-types.


packages/schema/src/index.ts
```diff
@@ -1 +1 @@
-// This project does not yet include TypeScript source, but it does use TypeScript to depend on the build of @powersync/service-types.
+// This project does not yet include TypeScript source, but it does use TypeScript to depend on the build off @powersync/service-types.
```
rkistner: This change looks wrong?

modules/module-mongodb/src/types/types.ts
```diff
@@ -54,7 +54,7 @@ export interface NormalizedMongoConnectionConfig {
   postImages: PostImagesOption;
 }
 
-export const MongoConnectionConfig = lib_mongo.BaseMongoConfig.and(service_types.configFile.DataSourceConfig).and(
+export const MongoConnectionConfig = service_types.configFile.DataSourceConfig.and(lib_mongo.BaseMongoConfig).and(
```
stevensJourney, line 57: The order of this union seemed to affect the json schema generation. Previously `mongodb` was not listed as a direct `type` option. 
I don't think we need to add a changeset for this internal change.

packages/schema/src/index.ts
stevensJourney: Good catch! 🙏

PR #498, opened 2026-02-10
Store sync plan in bucket storage

simolus3:
Sync configurations derived from the new compiler have a stable serialized format, meaning that we can store them as JSON data in bucket storage and retrieve it from there instead of recompiling every time the replicator starts. This allows us to evolve the compiler without breaking existing sync plans (which would have been the case before since the parameter format needs to match exactly between replication and the querier).

When deploying a sync config from YAML, it will get parsed and passed through `updateSyncRulesFromParsed` which attaches a sync plan if possible. When reading sync configurations (in `PersistedSyncRulesContent.parsed()`, we now take the sync plan into account and prefer to deserialize instead of re-compiling).

### Testing

I've tested this locally with the both mongo and Postgres storage implementations. First, I've deployed this sync config:

```yaml
streams:
  lists:
    query: SELECT * FROM lists
    auto_subscribe: true
  todos:
    query: SELECT * FROM todos WHERE list_id = subscription.parameter('list')

config:
  edition: 3
```

After deploying that, I've stopped the sync service and applied this patch to prevent the service from properly compiling new streams:

```patch
diff --git a/packages/sync-rules/src/from_yaml.ts b/packages/sync-rules/src/from_yaml.ts
index 124d341e..5252519a 100644
--- a/packages/sync-rules/src/from_yaml.ts
+++ b/packages/sync-rules/src/from_yaml.ts
@@ -222,7 +222,8 @@ export class SyncConfigFromYaml {
       streamCompiler.finish();
     }
 
-    const config = new PrecompiledSyncConfig(compiler.output.toSyncPlan(), {
+    const emptyCompiler = new SyncStreamsCompiler(this.options);
+    const config = new PrecompiledSyncConfig(emptyCompiler.output.toSyncPlan(), {
       defaultSchema: this.options.defaultSchema,
       engine: javaScriptExpressionEngine(compatibility),
       sourceText: this.yaml
```

After restarting the sync service, I've connected with the supabase todolist sync streams client to ensure data is there. If the service had re-compiled sync streams for the querier, I would not have seen any data since no streams would be available with that patch.

### DIFF
--- a/packages/service-core/src/storage/PersistedSyncRulesContent.ts
+++ b/packages/service-core/src/storage/PersistedSyncRulesContent.ts
@@ -85,21 +85,21 @@ export abstract class PersistedSyncRulesContent implements PersistedSyncRulesCon
     if (this.compiled_plan != null) {
       const plan = deserializeSyncPlan(this.compiled_plan.plan);
       const compatibility = CompatibilityContext.deserialize(this.compiled_plan.compatibility);
-
-      const precompiled = new PrecompiledSyncConfig(plan, {
-        defaultSchema: options.defaultSchema,
-        engine: javaScriptExpressionEngine(compatibility),
-        sourceText: this.sync_rules_content
-      });
-
+      const eventDefinitions: SqlEventDescriptor[] = [];
       for (const [name, queries] of Object.entries(this.compiled_plan.eventDescriptors)) {
         const descriptor = new SqlEventDescriptor(name, compatibility);
         for (const query of queries) {
           descriptor.addSourceQuery(query, options);
         }
 
-        precompiled.eventDescriptors.push(descriptor);
+        eventDefinitions.push(descriptor);
       }
+
+      const precompiled = new PrecompiledSyncConfig(plan, compatibility, eventDefinitions, {
+        defaultSchema: options.defaultSchema,
+        engine: javaScriptExpressionEngine(compatibility),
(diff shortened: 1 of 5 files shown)


packages/service-core/src/storage/PersistedSyncRulesContent.ts
```diff
@@ -72,11 +79,37 @@ export abstract class PersistedSyncRulesContent implements PersistedSyncRulesCon
...
+    let config: SyncConfigWithErrors;
+    if (this.compiled_plan != null) {
+      const plan = deserializeSyncPlan(this.compiled_plan.plan);
+      const compatibility = CompatibilityContext.deserialize(this.compiled_plan.compatibility);
+
+      const precompiled = new PrecompiledSyncConfig(plan, {
+        defaultSchema: options.defaultSchema,
+        engine: javaScriptExpressionEngine(compatibility),
+        sourceText: this.sync_rules_content
+      });
```
rkistner: Unless I'm missing something, it does not look like this ever sets `precompiled.compatibility`. It is set on the engine, but not on the compatibility object used by for example the `versionedBucketIds` check below.

simolus3: Good point, compatibility options should be set here. I've made them a constructor parameter to make this harder to miss.

PR #502, opened 2026-02-11
[Incremental Reprocessing] Prefiltering - Adds support for filtering data during initial replication

raffidahmad:
Enables filtering of rows during the initial snapshot phase of binlog replication, based on a configurable SQL WHERE clause.

This allows for partial snapshots, replicating only a subset of data based on specified criteria, which is particularly useful for large tables or scenarios where only recent data is needed.

Example included in the changeset file


raffidahmad:
> With someone not that familiar with the replication process, it may appear that these filters are applied per-definition. In reality, the filters are applied globally, and both definitions will end up with the same data. It is this mismatch between where the filters are specified versus how it is used that can cause confusion and potential issues.

Ah I see it now. For phase 1 we can keep them global and revisit more granular control in a separate issue, does that sound okay?

> Yes, that should work, just have to check for edge cases around special cases, such as special characters in the table name.

Okay, i will look into it.

> Since we support wildcard table names in queries, we should support the same for the filters. A specific use case example is Postgres partitioned tables.

Got it

> Technically you could use subqueries in the snapshot filters if we forward as-is, for example: `[WHERE] user_id NOT IN (SELECT user_id FROM deleted_users)`. However, I'd discourage this in most cases - the same subqueries or joins (if we did add support) won't work in the binlog streaming. So the only thing you can use it for is for optimizing the data that is loading initially, and I'd recommend rather denormalizing data to do that.

 i would assume the key being named "snapshot_filter" is self explanatory, that this will only apply to initial replication.
 Alternatively, we could call it "initial_snapshot_filter" ?

> Supporting JOINs in sync streams is something that we're working towards (with limitations), but that is different from using JOINs in the replication process.

I will leave JOIN support out for now since we can use sub-queries.

> So since those row filters must already be configured on the source database, I think it's best to just re-use the same ones for the snapshot queries, rather than duplicating in sync rules. That would also match how Postgres itself does logical replication.
> 
> MySQL and MongoDB don't have this same functionality, so there it does make more sense to configure it in the sync config.

I'm struggling to understand what you mean by duplication and setting it up in the sync config, could you provide an example for this please?

raffidahmad:
@rkistner fixed the tests, though the prettier yaml one is still failing for me locally, is that a known issue or is it one from my PR?
could you please run the workflow again

raffidahmad:
@rkistner The tests failing are they because of this PR? 
"Error: Error: Table 'users' already exists"

henriquekraemer:
We've been running a Postgres port of this in staging (schema-per-tenant, largest table 13.6M rows) and the results are good: on our largest tenant the initial snapshot went from reading all 13.6M rows to only the ~282k that match the filter.

A few things we found along the way that may be useful for this PR. I've opened #754 with the Postgres port and the fixes below, in case any of it helps here.

1. The filters get silently dropped when the config uses `edition: 3`. They live on `SyncConfig`, but edition-3 deploys persist a compiled sync plan, and at runtime `parsePersistedSyncConfigContent` rebuilds the config from the stored plan (`PrecompiledSyncConfig`), which never restores `initialSnapshotFilters`. Only the `compiledPlan == null` fallback goes through `SqlSyncRules.fromYaml` and picks them up. In practice the filters parse and validate fine but never reach the snapshot, so you get full-table snapshots with no warning. This sits in service-core, so it affects every source module. The fix persists the filters next to the plan (same treatment `eventDescriptors` already get) and restores them when parsing a persisted plan. One detail: they need to be stored as an ordered list of `{pattern, filter}` pairs instead of a map, since matching is first-match-wins and JSONB doesn't preserve object key order.

2. On tests: our first tests passed with the feature completely broken, so two traps worth avoiding here. If the data query filters the same rows as the snapshot filter, the bucket ends up with the same count either way; the stream in the test should select all rows and let the count measure the filter. And the test needs to reload the sync config from storage before replicating, like the replicator does on startup. Using the in-memory config from the deploy hides exactly the persistence bug above. This trap bit us twice: it also hid that our port never applied the filters with MongoDB bucket storage.

3. Probably worth a line in the docs: a selective filter on a big table needs an index matching the predicate. The chunked snapshot query is `WHERE <filter> ORDER BY pk LIMIT n`, and without an index Postgres walks the PK filtering row by row. On our 13.6M row table (~2% selectivity, matching rows clustered at high ids) the first chunk took 8m38s, which blew the snapshot socket timeout and left the table stuck in a retry loop. After creating a partial index on `(id)` with the same predicate, 2.6s.

PR #505, opened 2026-02-12
Refactor JWT payload and user id handling.

rkistner:
Follow-up to #490.

After doing more thorough checks, I think that we cannot change the output of `request.user_id()` for numbers. We do have users relying on `request.user_id()` being a number, so changing this to always be a string is a breaking change.

This updates user id handling, so that:
1. `token_parameters.user_id` and `request.user_id()` preserves numbers in `sub`, staying compatible with earlier versions.
2. `token_parameters.user_id` and `request.user_id()` does convert other types to a number or string, such as booleans, objects and arrays. These should not be used in practice, but this ensures we have well-defined behavior if it does come up.
3. Logs use the same value JSON-compatible value as `request.user_id()`.
4. Write checkpoints and user metrics use a stringified version of the above.

To achieve this, this does a refactoring of JWT payload handling:
1. The core JWT parsing logic is moved into the `sync-rules` package.
2. All tests now test using a raw JWT payload, rather than the old weird `token_parameters`.

These changes ensure that we can test the JWT payload handling directly inside the sync-rules package, rather than having conversions spread through other unspecified places.

## Alternative considered

I'm not aware of any valid use cases using anything other than a string or number for the `sub`. We could raise a hard error in that case, same as with other token validation issues.

However, it is theoretically possible that there are cases using for example a nested object for `sub`, and are either not using it, or using for example `request.jwt() ->> '$.sub.email'`, which would likely have worked.

We can consider making these type of breaking changes in the next major version.

### DIFF
--- a/packages/sync-rules/test/src/static_parameter_queries.test.ts
+++ b/packages/sync-rules/test/src/static_parameter_queries.test.ts
@@ -213,7 +213,7 @@ describe('static parameter queries', () => {
     ]);
   });
 
-  test.only("request.jwt() ->> '$.sub.email'", function () {
+  test("request.jwt() ->> '$.sub.email'", function () {
     const sql = "SELECT request.jwt() ->> 'sub.email' as email";
     const query = SqlParameterQuery.fromSql(
       'mybucket',


packages/sync-rules/src/types.ts
```diff
@@ -192,18 +216,18 @@ export class RequestParameters implements ParameterValueSet {
...
     // compatibility.
-    this.parsedTokenPayload = tokenPayload;
+    this.parsedTokenPayload = tokenPayload.parsedPayload;
     this.legacyTokenParameters = toSyncRulesParameters(
       tokenParameters,
       CompatibilityContext.FULL_BACKWARDS_COMPATIBILITY
     );
-    this.userId = tokenPayload.sub;
-    this.rawTokenPayload = JSONBig.stringify(tokenPayload);
+    this.userId = tokenPayload.userIdJson;
```
rkistner, line 229: This single line is the core change compared to the behavior pre #490. Since `sub` could take any JSON value, this lead to unsupported value types leaking into sync rule processing, causing weird behavior and errors.

packages/service-core/src/routes/configure-fastify.ts
```diff
@@ -61,7 +61,7 @@ export function configureFastifyServer(server: fastify.FastifyInstance, options:
 
   const generateContext: ContextProvider = async (request, options) => {
     return {
-      user_id: undefined,
+      user_id_string: undefined,
```
EspressoTrip-v2, line 64: I dont see `user_id_string` in the context type can this be removed?

rkistner: Good catch, that's from an older version of my refactor. I'll remove in a follow-up.

rkistner: Addressed in #513.

PR #510, opened 2026-02-17
[SQL Server] Schema change handling

Rentacookie:
### Summary

This change set adds schema change detection and handling for the SQL Server adapter.

Schema changes are handled differently to the other database adapters because the SQL Server CDC mechanism itself is designed to protect downstream consumers from schema changes. This can take the form where SQL Server does not allow certain schema changes to be done at all if CDC is enabled for a table, or when schema changes _are_ allowed, the corresponding capture instance schema does not get updated. Their is no way to update the schema of the capture instance, without replacing it. This involves:
Creating a new capture instance for the source table. PowerSync automatically uses the newest capture instance that is available.
OR
Disabling CDC for the source table and then re-enabling it. This effectively deletes the capture instance and starts with a fresh one.

Both of the above unfortunately require DB admin privileges, and involve manual intervention by the user. Therefore schema change handling for SQL server mainly takes the form of detection and then recovery once a capture instance has been updated.

### Schema Change Detection & Handling

The CDC poller checks for schema changes before each poll cycle and handles:
**Table create** – New tables matching sync rules are discovered and replicated
**Table drop** – Dropped tables are removed from the replicated set
**Table rename** – Renames are handled by dropping the old table and adding the new one. Renaming a table has no effect on the capture instance so we can continue replication without user intervention.
**New capture instance** – When a new capture instance is detected, the table is re-snapshotted before continuing replication.
**Missing capture instance** – When CDC is disabled or was never enabled, replication for that table is paused with a warning
**Column schema changes** – Pending DDL changes are detected and logged; Users are advised to disable and re-enable CDC for the table to bring the capture instance up to date with the schema changes.

Known Issues:
We cannot currently detect that a table has been dropped if PowerSync was stopped while the drop happened.
To remove the bucket data for the dropped table, the sync rules will need to be redeployed.

### Error Handling

Since PowerSync actively uses CDC stored procedures and the capture instances to replicate, it is susceptible to race conditions and deadlock errors when CDC is modified on the source database while PowerSync is running.
A retry mechanism was added so that some of these errors(especially deadlocks) can be recovered from.

### Testing

New tests were added to check the schema change handling. These required more careful state management than usual to handle the non-determinism introduced when manipulating CDC whilst PowerSync is running

### DIFF
--- a/.changeset/clean-spoons-confess.md
+++ b/.changeset/clean-spoons-confess.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Support a global `with` block defining common table expressions usable in all Sync Streams.
--- a/packages/sync-rules/src/from_yaml.ts
+++ b/packages/sync-rules/src/from_yaml.ts
@@ -17,6 +17,8 @@ import { ParsingErrorListener, SyncStreamsCompiler } from './compiler/compiler.j
 import { syncStreamFromSql } from './streams/from_sql.js';
 import { PrecompiledSyncConfig } from './sync_plan/evaluator/index.js';
 import { javaScriptExpressionEngine } from './sync_plan/engine/javascript.js';
+import { PreparedSubquery } from './compiler/sqlite.js';
+import { TablePattern } from './TablePattern.js';
 
 const ACCEPT_POTENTIALLY_DANGEROUS_QUERIES = Symbol('ACCEPT_POTENTIALLY_DANGEROUS_QUERIES');
 
@@ -76,11 +78,22 @@ export class SyncConfigFromYaml {
     // Bucket definitions using explicit parameter and data queries.
     const bucketMap = parsed.get('bucket_definitions') as YAMLMap | null;
     const streamMap = parsed.get('streams') as YAMLMap | null;
+    const globalCtes = parsed.get('with') as YAMLMap | null;
 
     let result: SyncConfig;
     if (compatibility.edition >= CompatibilityEdition.COMPILED_STREAMS) {
-      result = this.#compileSyncPlan(bucketMap, streamMap, compatibility);
+      result = this.#compileSyncPlan(bucketMap, streamMap, globalCtes, compatibility);
     } else {
+      if (globalCtes != null) {
+        // We don't support CTEs at all in this compiler implementation.
+        this.#errors.push(
+          this.#yamlError(
+            globalCtes as Node,
+            'Common table expressions are not supported without the `sync_config_compiler` option.'
+          )
+        );
+      }
+
       result = this.#legacyParseBucketDefinitionsAndStreams(bucketMap, streamMap, compatibility);
     }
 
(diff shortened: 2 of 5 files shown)


modules/module-mssql/src/replication/CDCStream.ts
```diff
@@ -472,6 +451,14 @@ export class CDCStream {
         skipExistingRows: true
       },
       async (batch) => {
+        if (status === SnapshotStatus.RESNAPSHOT_REQUIRED) {
+          for (const table of specificTablesToResnapshot!) {
+            await batch.drop([table.sourceTable]);
+            // Update table in the table cache
+            await this.processTable(batch, table.sourceTable, table.captureInstance, false);
+          }
+        }
```
rkistner: When a resnapshot is required, this resnapshots certain tables, then resets snapshotLSN to `getLatestReplicatedLSN()`. If changes are made to _other_ tables during this resnapshot, wouldn't those be skipped then? Is it possible to do this without resetting the LSN?

modules/module-mssql/src/replication/CDCPoller.ts
```diff
@@ -186,61 +234,192 @@ export class CDCPoller {
...
+          table,
+          newCaptureInstance: latestCaptureInstance
+        });
+      }
+    }
+
+    return schemaChanges;
+  }
+
+  private async checkForNewTables(): Promise<CaptureInstanceDetails[]> {
```
rkistner: There is no asynchronous work in this function - can be a normal synchronous function.

modules/module-mssql/src/replication/CDCPoller.ts
```diff
@@ -107,14 +123,41 @@ export class CDCPoller {
       }
 
       try {
+        // Refresh capture instance details
+        this.captureInstances = await getCaptureInstances({ connectionManager: this.connectionManager });
+        // Check and handle any schema changes before polling
+        const schemaChanges = await this.checkForSchemaChanges();
```
rkistner: This runs a _lot_ of work on every polling loop:
1. Get all capture instances (`sys.sp_cdc_help_change_data_capture`) - typically 1 or 2 per table.
2. Get all schema changes for each capture instance (`sys.sp_cdc_get_ddl_history`).
3. `checkForSchemaChanges()` -> check whether each replicated table still exists (`sys.tables`).

Especially if a lot of tables are being replicated, this can slow down replication significantly, both in increased latency and reduced throughput. Do we have other options here?

For example:
1. Can we detect some of the changes in the existing polling loop? For example, does a dropped table or missing capture instance generate an error there, that we can then use as a trigger these type of checks?
2. Can we reduce the number of work we need to do for the case where there are no changes?
3. For unavoidable work, can we reduce the frequency we do the checks? For example, run when starting replication, and then perhaps on fixed intervals, instead of in every polling check.

modules/module-mssql/src/replication/CDCStream.ts
Rentacookie: Yikes, yes you're right. I'll handle this case.

modules/module-mssql/src/replication/CDCPoller.ts
Rentacookie: The current approach was greatly influenced by the unpredictability of the exceptions that were thrown when modifying CDC for tables whilst in a polling cycle. This made it difficult to latch onto specific errors, because you would get a different error depending what stage the CDC DB modification was in. So pre-checking the state of the capture instances and reacting accordingly proved to be more deterministic which was good for the tests.

That being said, I realise now that we are still open to same problem if a schema change happens right after a schema change check happened and the polling cycle already started. 😬 

A pragmatic approach would probably be a mix between (1) and (3). Let me see what I can do.
Basically check for schema changes on startup, or when an exception happened, or when a certain amount of time has passed.

PR #522, opened 2026-02-20
Sync Streams: Report warnings for dangerous parameters

simolus3:
`connection.parameter` and `subscription.parameter` values can be set to arbitrary values by clients, so they shouldn't be used for authorization.

For bucket definitions and sync streams in alpha, we have a check to emit warnings for that. This ports the logic to the new compiler. The basic logic for this is is unchanged from bucket definitions:

1. If any parameter depends on authenticated values, good.
2. Otherwise, if any parameter depends on unauthenticated values, bad.
3. If there are no parameters, warn if a request filter (e.g. `SELECT * FROM notes WHERE auth.parameter('is_admin')`) consists only of unauthenticated values.

For sync streams, the easiest place to implement this check is to use the intermediate representation just before we throw source locations away to generate the sync plan (this allows recursing through nested parameter lookups as well).

One complication is that we don't have a single source query on which we could report those errors: At this point, the model has been derived from a mix of CTEs and (potentially merged) data queries with different source YAML nodes. So a small refactoring is to track an error listener for each expression in `NodeLocations`, so that we can report warnings at the right location. This allows us to report the warning at the specific location an insecure connection/subscription parameter is used.

### DIFF
--- a/packages/sync-rules/src/compiler/detect_dangerous_parameters.ts
+++ b/packages/sync-rules/src/compiler/detect_dangerous_parameters.ts
@@ -30,7 +30,8 @@ export class DangerousParameterDetector {
 
       errors.report(
         'Clients can send any value for this unauthenticated parameter, so this is unsuitable for authorization. Consider further constraining this with an authenticated parameter.',
-        location
+        location,
+        { isWarning: true }
       );
     }
   }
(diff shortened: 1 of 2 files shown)


packages/sync-rules/src/compiler/detect_dangerous_parameters.ts
```diff
@@ -0,0 +1,100 @@
...
+      // We have at least one authenticated parameter, all good.
+      return;
+    } else {
+      // Unauthenticated parameter, warn.
+      const { location, errors } = classification.unauthenticated;
+
+      errors.report(
+        'Clients can send any value for this unauthenticated parameter, so this is unsuitable for authorization. Consider further constraining this with an authenticated parameter.',
+        location
+      );
```
rkistner, line 35: This appears to report these as a hard error. For the other parsers, we just treat these as a warning, and still allow deploying.

rkistner (approved):

PR #527, opened 2026-02-25
feat: ability to specify sync_config

stevensJourney:
This renames the current `sync_rules` entry in the PowerSync config to `sync_config`. We keep and deprecate `sync_rules` for backwards compatibility.

The changes here are more focussed on the user-facing declarations of `sync_rules`/`sync_config`. This does not rename sync rules internally throughout the entire codebase.

An example of YAML configuration

```yaml
replication:
  connections: []
storage:
  type: mongodb
sync_config:
  path: somepath to file
```

The configuration collection will fail validation if both `sync_config` and `sync_rules` are present.

We also currently allow for the sync config to be specified by environment variables. The current `POWERSYNC_SYNC_RULES_B64` is now deprecated, with a new `POWERSYNC_SYNC_CONFIG_B64` introduced.

### DIFF
--- a/.changeset/thin-paws-battle.md
+++ b/.changeset/thin-paws-battle.md
@@ -1,0 +1,7 @@
+---
+'@powersync/service-module-postgres-storage': minor
+'@powersync/service-module-mongodb-storage': minor
+'@powersync/service-core': minor
+---
+
+Store compiled sync plans in bucket storage.
--- a/modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
+++ b/modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
@@ -179,6 +179,7 @@ export class MongoBucketStorage extends storage.BucketStorageFactory {
         _id: id,
         storage_version: storageVersion,
         content: options.config.yaml,
+        serialized_plan: options.config.plan,
         last_checkpoint: null,
         last_checkpoint_lsn: null,
         no_checkpoint_before: null,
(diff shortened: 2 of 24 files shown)


packages/service-core/src/util/config/sync-rules/impl/filesystem-sync-rules-collector.ts
```diff
@@ -1,15 +1,15 @@
...
-import { configFile } from '@powersync/service-types';
 
 export class FileSystemSyncRulesCollector extends SyncRulesCollector {
   get name(): string {
     return 'FileSystem';
   }
 
   async collect(baseConfig: configFile.PowerSyncConfig, runnerConfig: RunnerConfig): Promise<SyncRulesConfig | null> {
-    const sync_path = baseConfig.sync_rules?.path;
+    const sync_path = baseConfig.sync_config?.path ?? baseConfig.sync_rules?.path;
```
simolus3: I don't think this is actually possible (and we are validating for it), but perhaps it could be written like this to avoid parsing an inconsistent state if `sync_config` has an inline definition while `sync_rules` has a `path` key:


```suggestion
    const sync_path = (baseConfig.sync_config ?? baseConfig.sync_rules)?.path;
```

Same for the inline collector and the `exit_on_error` option.

stevensJourney: True, we could use either one of `sync_config` or `sync_rules` at this stage. Since the collector allows for a lazy migration from these fields. If that lazy migration approach is deemed acceptable, then we could simplify this even further.

PR #529, opened 2026-02-25
Sync plans: Allow using table-valued functions in more places

simolus3:
When processing row or parameter data with sync plans, we essentially evaluate SQL statements of the form `SELECT $output, $parameters FROM $addedTableValuedFunctions WHERE $filter`, where values from the row to sync are embedded as prepared statement parameters.

This is pretty flexible and should give us everything we need. In sync plans however, `$filter` and `$output` could not depend on outputs from table-valued functions. This PR lifts that restriction.

### Supported before

1. Table-valued functions backing bucket parameters: `SELECT products.* FROM products, json_each(products.regions) AS regions WHERE regions.value = auth.parameter('region')`.
2. Table-valued functions backing inputs on parameter indexes (essentially imagine example 1 as a subquery somewhere).
3. Outputs of parameter indexes (`SELECT * FROM users WHERE id IN (SELECT members.value FROM conversations, json_each(conversations.members) AS members WHERE conversations.id = subscription.parameter('chat'))`).
4. Filters only depending on table-valued data: `SELECT * FROM customers WHERE 'synced' IN customers.properties` (which expands to joining `json_each(customers.properties)` and doing `'synced' = json_each.value`).

### Additionally supported now

Note: These aren't yet supported in the compiler, and in case 1. it's not clear to me whether we should. But enabling this in sync plans and adding tests doesn't hurt.

1. Using table-valued functions as columns to sync (`SELECT products.id || '_' || expanded.region as id, products.name FROM products, json_each(region.regions) AS expanded`).
2. Using a filter that depends on both table-valued functions and a column in the row to process (`SELECT * FROM tasks WHERE state IN '["active", "completed"]'`). We should add support for that to the compiler, this prepares support for that in sync plans.

Note: Skipping a changeset entry for this since there are pending `patch` entries on the sync rules package already.

### DIFF
--- a/packages/sync-rules/src/sync_plan/plan.ts
+++ b/packages/sync-rules/src/sync_plan/plan.ts
@@ -47,9 +47,9 @@ export interface TableProcessor {
   /**
    * All of these expressions exclusively depend on the {@link sourceTable}.
    *
-   * All of the filters must evaluate to a "true-ish" value for the row to be processed.
+   * All of the filters must evaluate to a truthy (according to SQL definitions) value for the row to be processed.
    */
-  filters: SqlExpression<ColumnSqlParameterValue>[];
+  filters: SqlExpression<TableProcessorData>[];
   /**
    * How to partition rows.
    *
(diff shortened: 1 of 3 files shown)


rkistner (approved):
The example of expanding arrays into separate synced rows is interesting. I feel it could be useful in some cases, for example converting from a typical MongoDB-type structure with embedded arrays, to a more SQL-type structure using separate rows. That said, it could be more efficient to do that type of conversion using views on the client, rather than syncing these as separate rows.

Issue #528, opened 2026-02-25
[MySQL] Avoid storing copies of row data


rkistner:
Related to #398.

For Postgres, we handle partial updates (TOAST columns excluded) by storing a copy of each row in `current_data`.

For MySQL, we currently do the same, but don't actually need that. In MySQL, there is a [binlog_row_image](https://dev.mysql.com/doc/refman/8.4/en/replication-options-binary-log.html#sysvar_binlog_row_image) setting:

> full (Log all columns)
> minimal (Log only changed columns, and columns needed to identify rows)
> noblob (Log all columns, except for unneeded BLOB and TEXT columns)

So for `full` (the default), we get all columns for every change, and don't need to store a copy of the data in `current_data`.

For `minimal` and `noblob`, some columns could be excluded from the binlog. Right now, we don't support that either way - those columns would end up as `null`.

So I propose:
1. Always use `storeCurrentData: false` for MySQL.
2. Add an additional validation check when connecting, ensuring that `binlog_row_image=full`.

The validation would not catch all cases: It is possible to change the value of `binlog_row_image` for a specific session. However, it should catch common cases of global misconfiguration.

I don't think it's worth specifically supporting `minimal` or `noblob` configuration: Any savings the source database gets from that would just be transferred to overhead in the PowerSync storage database. I prefer that we just document and check that it should always be `full`.

rkistner:
Implemented in #530.

PR #530, opened 2026-02-25
[MySQL] Avoid storing copies of row data

raffidahmad:
Introduces a pre-flight check to validate that the global `binlog_row_image` variable is set to `FULL`. This ensures MySQL always provides complete row post-images in the binary log for all changes.

Leverages this guarantee to disable internal storage of current row data (`storeCurrentData: false`) within the replication stream. This reduces memory usage and simplifies data handling, as the full state of a row after a change can be directly derived from the binlog event.

Provides clear error messages for misconfigured `binlog_row_image` settings or privilege issues.

Relates to #528

### DIFF
--- a/.changeset/orange-shoes-teach.md
+++ b/.changeset/orange-shoes-teach.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-module-mysql': minor
+---
+
+[MySQL] Add check for binlog_row_image=FULL.
--- a/.changeset/silent-dingos-march.md
+++ b/.changeset/silent-dingos-march.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-module-mysql': minor
+---
+
+[MySQL] Avoid storing a copy of all replicated data - we do not need this for MySQL.


modules/module-mysql/src/replication/BinLogStream.ts
```diff
@@ -251,6 +251,30 @@ export class BinLogStream {
     // Replication will be performed in a single transaction on this connection
     const connection = await this.connections.getStreamingConnection();
     const promiseConnection = (connection as mysql.Connection).promise();
+
+    // Validate global binlog_row_image is FULL
+    try {
+      const [rows] = await promiseConnection.query<mysqlPromise.RowDataPacket[]>(
```
rkistner: There is an existing `checkSourceConfiguration` method that contains a lot of similar checks, including checking `binlog_format`. Could you please move the check there?

modules/module-mysql/src/replication/BinLogStream.ts
```diff
@@ -251,6 +251,30 @@ export class BinLogStream {
...
+        "SHOW GLOBAL VARIABLES LIKE 'binlog_row_image'"
+      );
+      const image = rows?.[0]?.Value ?? rows?.[0]?.value; // MySQL vs MariaDB casing
```
rkistner: There's at least one other place where we just use `.Value` - not sure if that means MariaDB is not supported?

Rentacookie: Yeah that's right - we do not currently support MariaDB.

modules/module-mysql/src/replication/BinLogStream.ts
Rentacookie: Yes, adding the check in the [checkSourceConfiguration](https://github.com/powersync-ja/powersync-service/blob/main/modules/module-mysql/src/common/check-source-configuration.ts) is pretty much all that is needed to ensure that the configuration is checked in all the right places.

`checkSourceConfiguration` is also used when testing a connection.

PR #533, opened 2026-02-25
New sync compiler: Support `NOT IN` operator for scalar JSON arrays

simolus3:
Unlike the alpha sync streams implementation, the beta sync streams compiler doesn't support `NOT IN` at all. The reason is that we try to transform `IN` into a bucket parameter (e.g. `SELECT * FROM notes WHERE state IN subscription.parameter('requested_state')` splits by `state`, then array-expands the subscription parameter when a user connects to find buckets) and that logic doesn't work for `NOT IN` operators.

For the special case where the right-hand side of the `NOT IN` operator is a scalar expression (not a subquery or CTE), this introduces the `json_contains` function to evaluate the operator. We don't want to do that for regular `IN` operators because it's a restriction: Arguments to a function must all come from the same table. But for `NOT IN` operators, it's an improvement:

1. We continue to support `IN` and `&&` operators with all existing targets.
2. We continue to support weird cases like `WHERE NOT (x NOT IN (SELECT id FROM ...))`  where boolean normalization removes the `NOT` (ugh, why did I ever add support for that...).
3. We now support `WHERE state NOT IN '["foo", "bar"]'`.

Using `NOT IN` across tables is still not supported because that requires a parameter, but that has never been supported by an older service version either.

### DIFF
--- a/packages/sync-rules/src/compiler/sqlite.ts
+++ b/packages/sync-rules/src/compiler/sqlite.ts
@@ -458,7 +458,7 @@ export class PostgresToSqlite {
     if (translateRightResult.isScalar) {
       return this.negate(original, {
         type: 'function',
-        function: 'json_contains',
+        function: 'ps_json_contains',
         parameters: [translatedLeft, translateRightResult.expr]
       });
     } else {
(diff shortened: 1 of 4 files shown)


rkistner (approved):
Just a note that the IN behavior here differs a bit from standard SQL/SQLite semantics, especially when considering NULL or numeric values. But this matches our sync rules IN semantics, and our usage of `IN [JSON TEXT]` is very non-standard anyway, so I think that's fine.

It may be worth removing this weird behavior and going for a more SQL-compatible approach in a future config.edition, especially as our support for alternative approaches like `json_each` is expanding, but this looks good for now.

PR #534, opened 2026-02-25
Update description and author in package.json

cahofmeyr:


package.json
```diff
@@ -2,10 +2,10 @@
   "name": "powersync-service",
   "version": "0.0.0",
   "preview": true,
-  "description": "PowerSync source available monorepo",
+  "description": "PowerSync Service source available monorepo",
   "license": "FSL-1.1-ALv2",
   "keywords": [],
```
kobiebotha, line 7: should we add a couple keywords while we're at it?

PR #538, opened 2026-03-02
Generate type-safe wrappers for sync streams

simolus3:
As an early adopter pointed out, developers need to be careful when subscribing to sync streams because typos in stream or parameter names lead to missing data that can only be debugged by looking at the sync status to check resolved streams.

E.g. with streams defined like this:

```yaml
streams:
  notes:
    query: SELECT * FROM notes WHERE project = subscription.parameter('project')
```

It's fairly easy to do something like `db.syncStream('note', { 'project_id': ... })` which would fail. To make sync streams easier to subscribe to, this generates type-safe wrappers along with the schema. The updated tests give a good example of how the generated code looks like, I'll give an overview of how the generated code would be _used_ here:

```dart
// dart
final subscription = await TypedSyncStreams(db).notes(project: 'my_project_id').subscribe();
```

```Kotlin
// Kotlin
val subscription = TypedSyncStreams(db).notes(project="my_project_id").subscribe();
```

```Swift
// Swift
let subscription = try await TypedSyncStreams(db).notes(project: "my_project_id").subscribe();
```

```TypeScript
// TypeScript
const subscription = await typedStreams(db).notes({ project: 'my_project_id' }).subscribe();
```

```csharp
// C#, the new is kind of ugly but at least it's a ref struct
var subscription = await new TypedStreams(db).Notes(name: "name").Subscribe();
```

In all languages, we have typed parameters reflecting names and types of parameters.

### Implementation

While we can already infer the _return type_ of expressions, inferring parameters requires also inferring the type of subexpressions from an outer context. When looking at sync plans, that outer context is defined by the bucket or parameter definition: In `WHERE project = subscription.parameter('project')` instance, we can look at the partition keys of the bucket and infer their types from the schema. We then go through the instantiation to infer the corresponding type for each parameter.

Currently, only these simple `=` comparisons are implemented. The technique can be expanded to more complex parameters (WHERE project = subscription.parameter('project')), but that requires inferring the type of arguments from a function call. We don't have that information available at the moment, so we'd just return an `ANY` type in that case. Since the schema generation is a one-off action that doesn't require strict backwards compatibility, we can always revisit this in the future.

### DIFF
--- a/packages/sync-rules/test/src/generate_schema.test.ts
+++ b/packages/sync-rules/test/src/generate_schema.test.ts
@@ -151,14 +151,16 @@ export const AppSchema = new Schema({
 
 export type Database = (typeof AppSchema)['types'];
 
-export const typedStreams = Object.freeze({
-  assetsOne(db: PowerSyncDatabase): SyncStream {
-    return db.syncStream('assets_one', {});
-  },
-  assets2(db: PowerSyncDatabase, params: { name: string }): SyncStream {
-    return db.syncStream('assets_2', params);
-  }
-});
+export function typedStreams(db: PowerSyncDatabase) {
+  return {
+    assetsOne(): SyncStream {
+      return db.syncStream('assets_one', {});
+    },
+    assets2(params: { name: string }): SyncStream {
+      return db.syncStream('assets_2', params);
+    }
+  };
+}
 `
     );
 
@@ -194,14 +196,16 @@ export const AppSchema = new Schema({
 
 export type Database = (typeof AppSchema)['types'];
 
-export const typedStreams = Object.freeze({
-  assetsOne(db: PowerSyncDatabase): SyncStream {
-    return db.syncStream('assets_one', {});
-  },
-  assets2(db: PowerSyncDatabase, params: { name: string }): SyncStream {
-    return db.syncStream('assets_2', params);
-  }
-});
+export function typedStreams(db: PowerSyncDatabase) {
+  return {
+    assetsOne(): SyncStream {
+      return db.syncStream('assets_one', {});
+    },
+    assets2(params: { name: string }): SyncStream {
+      return db.syncStream('assets_2', params);
+    }
+  };
+}
 `
     );
   });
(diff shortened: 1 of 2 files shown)


packages/sync-rules/test/src/generate_schema.test.ts
```diff
@@ -138,12 +150,21 @@ export const AppSchema = new Schema({
 });
 
 export type Database = (typeof AppSchema)['types'];
+
+export const typedStreams = Object.freeze({
+  assetsOne(db: PowerSyncDatabase): SyncStream {
```
stevensJourney: I like how in the other languages one doesn't have to pass the `db` instance to the individual sync stream methods.

If we want to use an Object for this, we could still store the `db` in the object and reference it based off the `this` context. The key is to use a regular `function` definition for the streams.

e.g.

```javascript
const test = Object.freeze({
  db: {getAll: () => Promise.resolve(["test"])},
  streamone: async function (params) {
    return this.db.getAll();
  },
});

// `test` is used as the `this` context in `streamone`
//  this is the expected usage, so it should be fine to expect the context type.
const result = await test.streamone({});
```

simolus3: I'm not sure I understand that example :D `db` would need to be passed into the object with a real PowerSync database to be able to use shared streams (since `.subscribe()` on a sync stream needs access to the database, we can't use a mock object).

I could imagine doing this:

```TypeScript
function typedStreams(db: PowerSyncDatabase) {
  return {
    streamOne: (params: {name: string}) => db.syncStream('stream_one', params),
  }
}
```

Users could then use `typedStreams(db).streamOne()`. This allows storing `typedStreams(db)` in a global variable somewhere to avoid calling it multiple times.

simolus3: I've updated TypeScript code and the example in the PR description. Let me know if that's what you had in mind or if I'm missing something.

stevensJourney: My bad :D. I assumed there might be some reasoning for using a static object as the export, I just wanted to point out having the `db` as an object member could help avoid needing to pass it in as a pram. 

Ideally, you would want to expose a function or class to generate that - and a closure is an even better option in that case. 

This LGTM.

PR #541, opened 2026-03-03
[COMPACT_ERROR] Leads to race conditon and fail compact job

varanasiaditya:
Context: ReplicationAssertionError: [PSYNC_S1101] Unexpected PUT operation at 1:1#bucket_by_patient_ids["..."]:<op_id>

Root Cause
A race condition existed between compactSingleBucket and clearBucket:

compactSingleBucket processes operations in descending order and queues UPDATE statements to convert PUT operations to MOVE operations
These UPDATEs are batched and only executed when flush() is called
clearBucket is called after flush(), but runs in a separate transaction with REPEATABLE READ isolation level
Due to transaction timing, clearBucket could query the database before all PUT→MOVE conversions were visible, encountering unconverted PUT operations
Since clearBucket only expects MOVE/REMOVE/CLEAR operations, it throws an error when it encounters a PUT
Solution
Added AND op != 'PUT' filter to the clearBucket query to explicitly exclude PUT operations. This ensures that only already-converted operations (MOVE, REMOVE, CLEAR) are processed, preventing the race condition.

Changes
Modified PostgresCompactor.clearBucket() query to filter out PUT operations
This aligns with the method's documented purpose: processing operations after the main compaction has converted PUTs
Testing
Verified no duplicate operations exist in bucket_data after fix
Compaction should now complete successfully without encountering unconverted PUT operations

### DIFF
--- a/.changeset/eleven-pets-speak.md
+++ b/.changeset/eleven-pets-speak.md
@@ -1,0 +1,13 @@
+---
+'@powersync/service-module-postgres-storage': minor
+'@powersync/service-module-mongodb-storage': minor
+'@powersync/service-core-tests': minor
+'@powersync/service-module-postgres': minor
+'@powersync/service-errors': minor
+'@powersync/service-module-mongodb': minor
+'@powersync/service-core': minor
+'@powersync/service-module-mysql': minor
+'@powersync/lib-services-framework': minor
+---
+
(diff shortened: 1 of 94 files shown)


rkistner:
Covered in #572.

PR #543, opened 2026-03-04
Track bucket and parameter index source

rkistner:
This adds a `source` property throughout sync rules, to track the specific definitions used.

The main motivation is to use the specific source definition for storage when using incremental reprocessing: Instead of persisting data for all buckets and parameter indexes together, we'd partition storage by the source definition.

During replication:
1. `evaluateRow()` includes the `BucketDataSource` on the result `EvaluatedRow`.
2. `evaluateParameterRow()` includes the `ParameterIndexLookupCreator` on the result `ScopedParameterLookup`.

During sync:
1. `BucketParameterQuerier` includes the `BucketDataSource` on `ResolvedBucket`, in both `staticBuckets` and `queryDynamicBucketDescriptions`.
2. Parameter index lookups in `BucketParameterQuerier` all include the `ParameterIndexLookupCreator` via `ScopedParameterLookup`.

Additional refactoring:
1. In sync rules, move some "cast"-related functions to a separate module to avoid circular imports (caused issues on this branch).
2. In `BucketChecksumState`, use `ResolvedBucket` instead of `BucketDescription` internally to keep track of buckets to fetch (simplifies code slightly).

Overall, this PR should have no user-visible impact on its own.

Right now, the tests mostly use dummy values for the source. This may change if we start using the source more.

These changes are largely extracted from #468, but simplified slightly versus that version:
1. Use a normal `source` property instead of a symbol.
2. Make the property non-enumerable, to avoid significant changes to tests, and avoid requiring a symbol to exclude it from sync output.

## Alternatives considered

Technically we don't need the entire source reference for partitioning. One alternative I considered is using the assigned definition number (see #468) instead of the object. While that simplifies some things like tests, it ties the types in sync rules to the specific storage structure (e.g. the fact that we use numbers to identify the definitions). By embedding the source itself, the storage implementation can be flexible in how it uses it.

### DIFF
--- a/.changeset/light-days-attack.md
+++ b/.changeset/light-days-attack.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+When generating schema code, also generate helpers to subscribe to Sync Streams.
--- a/.changeset/slow-plants-deliver.md
+++ b/.changeset/slow-plants-deliver.md
@@ -1,0 +1,7 @@
+---
+'@powersync/service-module-postgres-storage': patch
+'@powersync/service-module-mongodb-storage': patch
+'@powersync/service-core-tests': patch
+---
+
+[Postgres Storage] Fix concurrency issue in compacting, leading to "[PSYNC_S1101] Unexpected PUT operation".
--- a/modules/module-mongodb-storage/src/storage/implementation/MongoCompactor.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/MongoCompactor.ts
@@ -86,18 +86,18 @@ export class MongoCompactor {
   constructor(
     private storage: MongoSyncBucketStorage,
     private db: VersionedPowerSyncMongo,
-    options?: MongoCompactOptions
+    options: MongoCompactOptions
   ) {
     this.group_id = storage.group_id;
-    this.idLimitBytes = (options?.memoryLimitMB ?? DEFAULT_MEMORY_LIMIT_MB) * 1024 * 1024;
-    this.moveBatchLimit = options?.moveBatchLimit ?? DEFAULT_MOVE_BATCH_LIMIT;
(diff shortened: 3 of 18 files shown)


packages/sync-rules/src/BucketSource.ts
```diff
@@ -171,12 +170,18 @@ export function hydrateEvaluateRow(hydrationState: HydrationState, source: Bucke
       if (isEvaluationError(result)) {
         return result;
       }
-      return {
-        bucket: buildBucketName(scope, result.serializedBucketParameters),
+      const evaluated = {
```
simolus3: nit: Could we export and use `withBucketSource` from `utils.ts` here?

packages/sync-rules/src/BucketDescription.ts
```diff
@@ -19,7 +21,29 @@ export const isValidPriority = (i: number): i is BucketPriority => {
...
+ * We never instantiate this class - we only use the type.
+ */
+class NonEnumerableSourceClass {
+  private constructor() {}
+
+  /**
+   * This is specifically not enumerable - must be excluded from tests and serialization.
+   */
+  get source(): BucketDataSource {
+    throw new Error('Do not use');
```
simolus3: nit: Since we only use the interface of this class, can we make it and `get source` abstract to avoid the stub definition?

packages/sync-rules/test/src/streams.test.ts
```diff
@@ -104,6 +98,49 @@ describe('streams', () => {
     ]);
   });
 
+  test('tracks source metadata on stream APIs', async () => {
```
simolus3, line 101: This tests the legacy/alpha sync streams compiler. It would probably be worth adding a similar test to `test/src/sync_plan/evaluator/evaluator.test.ts`?

packages/sync-rules/src/BucketDescription.ts
rkistner: Good point, that does make it cleaner.

PR #549, opened 2026-03-06
Refactor tests for startBatch() -> createWriter()

rkistner:
This ports a refactoring from #468:
1. Replace `startBatch()` using a callback with `createWriter()` as an AsyncDisposable.
2. Pull that refactoring throughout the tests.
3. In tests, replace a hardcoded TEST_TABLE with `await resolveTestTable(writer)`, which will eventually return a real SourceTable instead of a mocked one.
4. There are also other minor test changes to align the tests more with #468.

This does not remove the internal usage of `startBatch()` yet - it mostly focuses on the test refactor. The internal usage of `startBatch()` can be tackled next.

The main goal is to reduce the diff size and potential for merge conflicts from #468.

### DIFF
--- a/.changeset/dirty-points-explode.md
+++ b/.changeset/dirty-points-explode.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Fix a compiler crash when using a column of a CTE in another subquery.
--- a/.changeset/dirty-ties-itch.md
+++ b/.changeset/dirty-ties-itch.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Improve error messages when a query attempts to partition a table-valued function depending on request data.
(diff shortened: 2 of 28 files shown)


modules/module-mongodb-storage/src/storage/implementation/MongoBucketBatch.ts
```diff
@@ -668,6 +668,10 @@ export class MongoBucketBatch
     super.clearListeners();
   }
 
+  async dispose() {
+    await this[Symbol.asyncDispose]();
```
stevensJourney, line 678: Previously a batch would be created using the `startBatch` method - which would call a `flush` automatically after the provided `callback` had successfully executed. 
It's probably not feasible to automatically flush on `dispose` - since that is also called even if there has been an error in the call trace, but, maybe we could warn or print a log if the batch is being disposed with pending items still present?

stevensJourney (approved):
I like this new addition. Happy with the changes.

modules/module-mongodb-storage/src/storage/implementation/MongoBucketBatch.ts
rkistner: Good call - I added the warning, and confirmed that it does not show up in the current tests.

PR #552, opened 2026-03-06
Fix crash for subqueries referenced in another subquery

simolus3:
This fixes a [report from Discord](https://discord.com/channels/1138230179878154300/1479042473316847746/1479049290830839928).

In `StreamQueryParser`, `subqueryResultSets` stores expanded subqueries which is used to inline column references. When we parse another subquery with access to a subquery that has already been joined, we need to pass `subqueryResultSets` down to the inner parser to avoid `internal error: result set from scope has not been registered` errors. This PR fixes that, and adds a regression test with a query derived from the original report.


simolus3:
I've simplified the regression test a bit, the original query had some other issues that made the generated sync plan very complicated (it was correct in both cases though, but it took me quite some time to double-check).

rkistner (approved):

PR #554, opened 2026-03-07
feat(module-postgres): early warning and mitigation for WAL slot invalidation during snapshot

Sleepful:
## Summary

- Detect replication slot invalidation mid-snapshot and abort early instead of completing a doomed hours-long snapshot
- Log WAL budget consumption with rate and ETA during snapshot — warn at 50% remaining
- Block futile replication retries when slot is lost during snapshot (retrying would repeat the same long snapshot)
- Surface actionable error messages with error code `PSYNC_S1146` and docs link
- Expose WAL budget fields in the diagnostics API for dashboard/operator visibility

## Problem

When PowerSync runs a full snapshot (triggered by sync rule changes), the replication slot can be silently invalidated if WAL growth exceeds `max_slot_wal_keep_size`. The operator only discovers this after the snapshot completes and streaming fails — often hours or days later. The fix is simple (increase the limit) but discovery is painful.

### Mid-snapshot slot health check

After each chunk flush in `snapshotTable()`, query `pg_replication_slots` to check if the slot is still valid. If `wal_status = 'lost'`, abort immediately with an enriched `MissingReplicationSlotError` carrying `walStatus` and `phase` fields.

The check hits shared memory (~1-2ms per round-trip), negligible next to per-chunk storage flush.

### Conditional retry

New `shouldRetryReplication()` function controls whether `WalStreamReplicationJob` retries after slot invalidation:

| Condition | Action |
|---|---|
| Slot lost during **snapshot** | **Block** retry — would repeat the same long snapshot |
| Slot lost during **streaming** | **Allow** retry — streaming invalidation is often transient |
| Slot **missing** | **Allow** retry — may have been dropped externally |
| `invalidation_reason = rows_removed` | **Allow** retry — not a WAL budget issue |

### WAL budget reporting

Time-throttled logging (every 2 min) of WAL budget during snapshot:
- Budget remaining (bytes and %)
- Consumption rate (GB/hr) computed from successive samples
- ETA to exhaustion
- Warning at 50% remaining

Computation extracted into exported pure functions (`computeWalBudgetReport`, `formatWalBudgetLine`, `formatBytes`, `formatDuration`) for testability.

### Actionable error messages

Slot invalidation errors now include:
- Error code `PSYNC_S1146`
- Fix guidance ("Increase max_slot_wal_keep_size on the source database")
- Docs link (`docs.powersync.com/self-hosting/troubleshooting/replication-slot-invalidated`)
- Observed WAL budget context when available (limit, time to exhaustion)

Applied to both `checkSlotHealth()` (mid-snapshot) and `initSlot()` (between replication cycles).

### Diagnostics API

Three new optional fields on the `SyncRulesStatus` connection object:
- `wal_status` — slot status from `pg_replication_slots` (PG 13+)
- `safe_wal_size` — bytes remaining before potential invalidation
- `max_slot_wal_keep_size` — configured limit in bytes

New optional `getSlotWalBudget()` on `RouteAPI`, implemented by `PostgresRouteAPIAdapter`. Non-Postgres adapters are unchanged.

### Bug fix

`withMaxWalSize()` test helper was ignoring its `size` parameter and hardcoding `'100MB'`. Now uses the parameter.

## Files changed

| File | Change |
|------|--------|
| `modules/module-postgres/src/replication/WalStream.ts` | `MissingReplicationSlotError` enrichment, `checkSlotHealth()`, WAL budget tracking, `shouldRetryReplication()`, `formatWalBudgetContext()`, pure functions |
| `modules/module-postgres/src/replication/WalStreamReplicationJob.ts` | Conditional retry wiring |
| `modules/module-postgres/src/api/PostgresRouteAPIAdapter.ts` | `getSlotWalBudget()` implementation |
| `packages/service-core/src/api/RouteAPI.ts` | `SlotWalBudgetInfo` interface, optional method |
| `packages/service-core/src/api/diagnostics.ts` | WAL budget fields in diagnostics response |
| `packages/service-errors/src/codes.ts` | `PSYNC_S1146` error code |
| `packages/types/src/definitions.ts` | `wal_status`, `safe_wal_size`, `max_slot_wal_keep_size` fields |
| `modules/module-postgres/test/src/wal_stream.test.ts` | 2 integration tests (slot loss detection + diagnostic context) |
| `modules/module-postgres/test/src/replication_retry.test.ts` | 4 pure function tests for retry logic |
| `modules/module-postgres/test/src/wal_budget.test.ts` | 17 pure function tests for budget computation/formatting |
| `modules/module-postgres/test/src/wal_budget_api.test.ts` | 3 integration tests for diagnostics adapter |
| `modules/module-postgres/test/src/wal_stream_utils.ts` | `withMaxWalSize()` bug fix |

## Testing

- 26 new tests total (4 retry + 17 budget + 3 diagnostics API + 2 integration)
- All existing tests pass
- Pure function tests run without Postgres in milliseconds
- Integration tests use `onSnapshotChunkFlushed` hook for deterministic WAL generation (no race conditions)

### DIFF
--- a/modules/module-postgres/src/replication/MissingReplicationSlotError.ts
+++ b/modules/module-postgres/src/replication/MissingReplicationSlotError.ts
@@ -8,11 +8,18 @@
  */
 export type WalStatus = 'reserved' | 'extended' | 'unreserved' | 'lost' | 'missing';
 
+/**
+ * Replication phase when the error was detected.
+ * - `snapshot`: during an active full-table scan (snapshotTable) — long-running, retry may be futile
+ * - `streaming`: during WAL streaming, at startup, or between replication cycles — retry is safe
+ */
+export type ReplicationPhase = 'snapshot' | 'streaming';
+
 export class MissingReplicationSlotError extends Error {
   /** Slot WAL status at the time the error was detected. */
   walStatus: WalStatus;
   /** Replication phase when the error occurred — controls retry behavior. */
-  phase: 'snapshot' | 'streaming';
+  phase: ReplicationPhase;
   /** PG 14+ invalidation reason from `pg_replication_slots.invalidation_reason`. */
   invalidationReason?: string;
 
@@ -20,7 +27,7 @@ export class MissingReplicationSlotError extends Error {
     message: string,
     options: {
       walStatus: WalStatus;
-      phase: 'snapshot' | 'streaming';
+      phase: ReplicationPhase;
       invalidationReason?: string;
       cause?: any;
     }


modules/module-postgres/src/replication/WalStream.ts
```diff
@@ -334,8 +478,15 @@ export class WalStream {
...
+        const fixGuidance =
+          slot.invalidation_reason === 'idle_timeout'
+            ? `Increase idle_replication_slot_timeout on the source database.`
+            : `Increase max_slot_wal_keep_size on the source database.`;
         throw new MissingReplicationSlotError(
-          `Replication slot ${slotName} is not valid anymore. invalidation_reason: ${slot.invalidation_reason ?? 'unknown'}`
+          `[PSYNC_S1146] Replication slot ${slotName} was invalidated ` +
+            `(reason: ${slot.invalidation_reason ?? 'unknown'}). ` +
+            `${fixGuidance} ` +
+            `https://docs.powersync.com/self-hosting/troubleshooting/replication-slot-invalidated`
```
Sleepful: this is 404 atm, use another URL or create this one eventually?

modules/module-postgres/test/src/wal_stream.test.ts
```diff
@@ -470,6 +470,101 @@ bucket_definitions:
...
+    }
+
+    await using _walSize = await withMaxWalSize(baseContext.pool, '100MB');
+
+    const walPool = baseContext.pool;
+    await using context = await openContext({
+      walStreamOptions: {
+        snapshotChunkLength: 100,
+        onSnapshotChunkFlushed: async () => {
+          // Generate ~16MB WAL per call. Slot lost after ~7 chunks.
```
Sleepful, line 490: this bit here is the key to intentionally lose the slot during tests

modules/module-postgres/test/src/wal_stream.test.ts
```diff
@@ -470,6 +470,101 @@ bucket_definitions:
...
+
+    await pool.query(`CREATE TABLE test_data(id uuid primary key default uuid_generate_v4(), description text)`);
+    await pool.query(`INSERT INTO test_data(description) SELECT 'row ' || g FROM generate_series(1, 1000) g`);
+
+    // The hook generates WAL synchronously after each chunk flush.
+    // Post-implementation, the WAL check (at the same injection point)
+    // detects slot invalidation and aborts the snapshot.
+    await expect(async () => {
+      await context.replicateSnapshot();
+    }).rejects.toThrowError(MissingReplicationSlotError);
```
Sleepful, line 513: This is the key test: starts snapshot, validates that it reports the error -> it means the code to identify 'slot was lost' is correct and that the test succeeds in losing the slot.

modules/module-postgres/src/replication/WalStream.ts
```diff
@@ -99,11 +113,125 @@ export const sendKeepAlive = async (db: pgwire.PgClient) => {
 };
 
 export class MissingReplicationSlotError extends Error {
-  constructor(message: string, cause?: any) {
+  walStatus?: string;
+  phase?: 'snapshot' | 'streaming';
```
rkistner: These are not consistently set everywhere the error is thrown, which could lead to assuming wrong default values. Can we make these values required?

modules/module-postgres/src/replication/WalStream.ts
```diff
@@ -99,11 +113,125 @@ export const sendKeepAlive = async (db: pgwire.PgClient) => {
...
+      this.phase = options.phase;
+    }
+  }
+}
+
+export interface SlotInvalidationContext {
+  walStatus: string; // 'lost' | 'reserved' | 'extended' | 'missing' | ...
+  phase: 'snapshot' | 'streaming';
+  invalidationReason?: string; // PG 14+: 'wal_removed', 'rows_removed', etc.
+}
```
rkistner: This repeats values that are (or should be) present on the MissingReplicationSlotError. Can we just pass the MissingReplicationSlotError object through directly, or make it implement this interface?

modules/module-postgres/src/replication/WalStream.ts
```diff
@@ -99,11 +113,125 @@ export const sendKeepAlive = async (db: pgwire.PgClient) => {
...
+ * Allows retry in all other cases.
+ */
+export function shouldRetryReplication(context: SlotInvalidationContext): boolean {
+  if (context.walStatus === 'lost' && context.phase === 'snapshot' && context.invalidationReason !== 'rows_removed') {
+    return false;
+  }
+  return true;
+}
+
+export function formatBytes(bytes: number): string {
```
rkistner: The file is becoming quite big with all the new utility functions - could we move them out to separate file(s)?

modules/module-postgres/src/replication/WalStreamReplicationJob.ts
```diff
@@ -91,8 +91,16 @@ export class WalStreamReplicationJob extends replication.AbstractReplicationJob
       }
 
       if (e instanceof MissingReplicationSlotError) {
-        // This stops replication on this slot and restarts with a new slot
-        await this.options.storage.factory.restartReplication(this.storage.group_id);
+        if (
+          shouldRetryReplication({
+            walStatus: e.walStatus ?? 'missing',
+            phase: e.phase ?? 'streaming',
+            invalidationReason: (e as any).invalidationReason
```
rkistner: Mentioned in the other comment - these default values could be incorrect. And right now, `invalidationReason` is actually never set on the error.

It would be better to make these required on the error, and pass through the error object as-is.

modules/module-postgres/src/replication/WalStream.ts
rkistner: Rather leave out the URL. The pattern we currently use is to just add the error code in the message, and the operator can use that to locate the relevant docs.

modules/module-postgres/src/replication/WalStream.ts
```diff
@@ -605,6 +902,10 @@ WHERE  oid = $1::regclass`,
 
       // Important: flush before marking progress
       await batch.flush();
+      if (this.onSnapshotChunkFlushed) {
+        await this.onSnapshotChunkFlushed();
+      }
+      await this.checkSlotHealth();
```
rkistner: This may happen multiple times per second, which could add some overhead to replication. And we don't need very fast detection here - detecting after a minute or two is fine (compare to currently only seeing the issue hours or days layer).

So I think we should throttle the check here already, instead of only throttling the log. Can add an option to override the behavior or period for tests.

modules/module-postgres/src/replication/WalStream.ts
```diff
@@ -355,6 +506,152 @@ export class WalStream {
...
+    }
+    const parts: string[] = [];
+    parts.push(` (limit: ${formatBytes(this.maxSlotWalKeepSize)})`);
+
+    if (this.prevWalBudgetSample != null) {
+      const elapsed = performance.now() - this.prevWalBudgetSample.timestamp;
+      if (elapsed > 0 && this.prevWalBudgetSample.safeWalSize > 0) {
+        const elapsedHours = elapsed / 3_600_000;
+        if (elapsedHours >= 0.1) {
+          parts.push(`, exhausted in ~${formatDuration(elapsedHours)}`);
```
rkistner: This appears to just log the elapsed time, not any ETA?

Sleepful: True! This is an odd line, removed. Good catch

modules/module-postgres/src/replication/WalStream.ts
Sleepful: Throttled ✅

modules/module-postgres/src/replication/WalStream.ts
Sleepful: made these values required, and addressed the other comments that relate to those values

modules/module-postgres/src/replication/MissingReplicationSlotError.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Replication slot WAL status from `pg_replication_slots.wal_status` (PG 13+).
+ * - `reserved`: WAL needed by the slot is within `max_wal_size`
+ * - `extended`: WAL needed exceeds `max_wal_size` but is protected by `max_slot_wal_keep_size`
+ * - `unreserved`: WAL may be removed at next checkpoint — slot is at risk
+ * - `lost`: WAL has been removed — slot is invalidated and unusable
+ * - `missing`: synthetic value — the slot row is absent from `pg_replication_slots`
+ */
+export type WalStatus = 'reserved' | 'extended' | 'unreserved' | 'lost' | 'missing';
+
+export class MissingReplicationSlotError extends Error {
```
Sleepful, line 18: Here is the better typing for the errors

modules/module-postgres/src/replication/MissingReplicationSlotError.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Blocks retry when walStatus is 'lost' during snapshot phase (unless the
+ * invalidation reason is 'rows_removed', which is not a WAL budget issue).
+ * Allows retry in all other cases.
+ */
+export function shouldRetryReplication(error: MissingReplicationSlotError): boolean {
+  if (error.walStatus === 'lost' && error.phase === 'snapshot' && error.invalidationReason !== 'rows_removed') {
+    return false;
+  }
+  return true;
+}
```
Sleepful, line 59: For posterity just in case: 

The retry decision matrix — why each combination matters:

| walStatus | phase | invalidationReason | Retry? | Rationale |
|---|---|---|---|---|
| `lost` | `snapshot` | (any except `rows_removed`) | **Block** | WAL budget exhausted during a hours-long snapshot. Retrying restarts the same snapshot, consumes the same WAL, loses the slot again. Futile loop. |
| `lost` | `snapshot` | `rows_removed` | Allow | Catalog rows were vacuumed — not a WAL budget issue. A fresh slot won't need those rows, so retry with a new slot works. |
| `lost` | `streaming` | (any) | Allow | Slot died during WAL streaming or was found dead at startup. Streaming is incremental, not hours-long. Retry with a new slot is the standard recovery. |
| `missing` | (any) | — | Allow | Slot row absent from `pg_replication_slots` — dropped externally (operator, cleanup tool). Creating a fresh slot is the correct fix. |
| `reserved` / `extended` / `unreserved` | — | — | Allow | Slot is still valid (or at risk but not dead). These statuses never appear in `MissingReplicationSlotError` in practice — the error is only thrown when the slot is `lost` or `missing`. |

The single blocking case (`lost` + `snapshot`) exists because it's the only scenario where retrying is provably wasteful: the snapshot that caused the WAL growth will run again, generate the same WAL, and exhaust the budget again. The operator must increase `max_slot_wal_keep_size` before retrying.

modules/module-postgres/src/replication/wal-budget-utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  safeWalSize: number;
+  maxSize: number;
+  walStatus: string;
+  ratePerHour: number | null;
+  etaHours: number | null;
+  isWarning: boolean;
+}
+
+export function computeWalBudgetReport(opts: {
+  safeWalSize: number;
```
Sleepful, line 37: **`computeWalBudgetReport` — input → output breakdown:**

### Inputs

| Argument | Source | What it represents |
|---|---|---|
| `safeWalSize` | `pg_replication_slots.safe_wal_size` (cast to number) | Bytes of WAL budget remaining before the slot gets invalidated. PG computes this as `max_slot_wal_keep_size - WAL_consumed_since_slot_restart_lsn`. Shrinks as WAL grows during snapshot. |
| `maxSize` | `pg_settings.max_slot_wal_keep_size` (converted to bytes) | The configured hard limit. When `safe_wal_size` hits 0, PG marks the slot `lost`. |
| `walStatus` | `pg_replication_slots.wal_status` | Current slot health — `reserved`, `extended`, `unreserved`. Included in the report for the log message. |
| `prevSample` | Previous call's `{ safeWalSize, timestamp }` | The last reading. Null on the first call (no rate possible). Used to compute how fast the budget is shrinking. |
| `now` | `performance.now()` | Current timestamp in ms. Compared against `prevSample.timestamp` to get elapsed time. |

### Outputs

| Field | Derived from | Meaning |
|---|---|---|
| `budgetRemainingPct` | `round(safeWalSize / maxSize * 100)` | How much budget is left as a percentage. 80% = healthy, 50% = warning territory. |
| `isWarning` | `budgetRemainingPct <= 50` | Triggers `logger.warn` instead of `logger.info` in the caller. |
| `ratePerHour` | `(prevSample.safeWalSize - safeWalSize) / elapsedMs * 3_600_000` | How fast WAL is being consumed in bytes/hour. Null if no previous sample or no WAL consumed. |
| `etaHours` | `safeWalSize / ratePerHour` | Hours until budget hits zero at current rate. Null if rate is null or ETA > 48h (unreliable extrapolation over long timeframes). |
| `safeWalSize`, `maxSize`, `walStatus` | Pass-through | Included in the report so `formatWalBudgetLine` can format the log message without needing the original inputs. |

packages/service-errors/src/codes.ts
```diff
@@ -186,6 +186,19 @@ export enum ErrorCode {
    */
   PSYNC_S1145 = 'PSYNC_S1145',
 
+  /**
+   * Replication slot invalidated.
+   *
+   * The replication slot was invalidated by PostgreSQL, typically because
+   * WAL retention exceeded max_slot_wal_keep_size during a long-running
+   * snapshot. Increase max_slot_wal_keep_size on the source database and
+   * redeploy sync rules.
```
benitav, line 195: Can we rename this to Sync Streams please? Sync rules are legacy. Assuming this is the message shown to end users.

PR #559, opened 2026-03-09
Add .NET class based schema generation

LucDeCaf:
Adds a schema generator for the new(ish) .NET class-based schema syntax. The old syntax remains the same.

### DIFF
--- a/.changeset/blue-mirrors-greet.md
+++ b/.changeset/blue-mirrors-greet.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-module-mongodb': patch
+---
+
+Remove a 1s delay in MongoDB initial replication.
--- a/.changeset/mean-impalas-walk.md
+++ b/.changeset/mean-impalas-walk.md
@@ -1,0 +1,9 @@
+---
+'@powersync/service-module-postgres-storage': patch
+'@powersync/service-module-mongodb-storage': patch
+'@powersync/service-core-tests': patch
+'@powersync/service-module-mongodb': patch
+'@powersync/service-core': patch
+---
+
+[Internal] Add a createWriter() API to replace startBatch().
--- a/.changeset/soft-bikes-collect.md
+++ b/.changeset/soft-bikes-collect.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Fix sync streams ignoring parts of `OR` conditions for subqueries not contributing bucket parameters.
--- a/.changeset/strong-bobcats-crash.md
+++ b/.changeset/strong-bobcats-crash.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-core': patch
+---
+
+Tighten Supabase URI regular expression.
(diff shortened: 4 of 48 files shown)


packages/sync-rules/src/schema-generators/DotNetSchemaGenerator.ts
```diff
@@ -20,13 +20,16 @@ ${tables.map((table) => this.generateTable(table.name, table.columns, options)).
 
     public static Schema PowerSyncSchema = new Schema(${tables.map((table) => this.toUpperCaseFirstLetter(table.name)).join(', ')});
 }`;
-    const streamHelper = this.generateStreamHelper(source, schema);
+
+    const optionalStreams = this.getOptionalStreams(source, schema);
+    const streamHelper = generateCSharpStreamHelper(optionalStreams);
     generated += `\n${streamHelper}\n`;
```
stevensJourney: This was picked up by an AI review. Seems like it could be an issue:

 `generateCSharpStreamHelper()` returns `string | undefined`, but the result is concatenated directly into the output string without a null
  check:

```typescript
   const streamHelper = generateCSharpStreamHelper(optionalStreams);
   generated += `\n${streamHelper}\n`;  // "undefined" gets written into the C# output!

  The sibling class DotNetClassSchemaGenerator (line 82) and other generators (Swift, Ts) all guard against this correctly. The fix:

   if (streamHelper) {
     generated += `\n${streamHelper}\n`;
   }
```

  This would silently inject the literal string `"undefined"` into generated C# code, causing compilation errors for users with no optional streams.

packages/sync-rules/src/schema-generators/DotNetSchemaGenerator.ts
```diff
@@ -20,13 +20,16 @@ ${tables.map((table) => this.generateTable(table.name, table.columns, options)).
...
+
     return generated;
   }
 
   private toUpperCaseFirstLetter(str: string): string {
-    return str.charAt(0).toUpperCase() + str.slice(1);
+    return str.charAt(0).toUpperCase() + (str.length > 1 ? str.slice(1) : '');
```
stevensJourney: The change looks unnecessary. `str.slice(1)` on a single-character (or empty) string already returns `''`. JavaScript's `slice` handles out-of-bounds indices gracefully, returning an empty string rather than throwing.

.changeset/lemon-lemons-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/service-sync-rules': minor
```
stevensJourney: Since this package is in version `0` and this is new non-breaking functionality, this can be a `patch` release.  Other additions, like [this](https://github.com/powersync-ja/powersync-service/pull/538/changes) one, have followed this pattern.

packages/sync-rules/src/schema-generators/DotNetSchemaGenerator.ts
LucDeCaf: Didn't know that, thanks!

stevensJourney (approved):
:shipit:

PR #564, opened 2026-03-13
Add "instr" SQLite function to Sync Streams

kobiebotha:
# Why

It might be desirable to do some substring detection in Sync Streams. 

One specific case is [Convex Auth](https://github.com/get-convex/convex-auth/blob/b46cd99a27a538f445241acfa08b4fee8b14fdac/src/server/implementation/tokens.ts#L41), where the "sub" claim is a compound string of the form:

`"k5764nz5tn54hncf94wh9458bn82nx89|jh76fr94jnqm0z5sqbc4wbx2vn82vmke"`

Note the `|`. Conceptually this string is `userId|sessionId`

In Sync Streams, a developer might want to extract the userId component of that compound string. 

# What
This PR enables substring detection using code based on SQLite's `instr` function: https://www.sqlite.org/lang_corefunc.html#instr

# TODO: 
[X] Testing that it actually works as advertised

### DIFF
--- a/packages/sync-rules/src/sql_functions.ts
+++ b/packages/sync-rules/src/sql_functions.ts
@@ -282,7 +282,7 @@ const instr: DocumentedSqlFunction = {
     }
     // Both BLOBs: byte-level search
     if (x instanceof Uint8Array && y instanceof Uint8Array) {
-      const pos = Buffer.from(x).indexOf(Buffer.from(y));
+      const pos = Buffer.from(x.buffer, x.byteOffset, x.byteLength).indexOf(y);
       return BigInt(pos < 0 ? 0 : pos + 1);
     }
     // Neither BLOB, or mixed: cast both to text


packages/sync-rules/src/sql_functions.ts
```diff
@@ -271,6 +271,31 @@ const iif: DocumentedSqlFunction = {
...
 };
 
+// This is the same behavior as the instr function in SQLite, but only supports TEXT values
+const instr: DocumentedSqlFunction = {
+  debugName: 'instr',
+  call(x: SqliteValue, y: SqliteValue) {
+    if (x == null || y == null) {
+      return null;
+    }
+    let pos: number = x.toString().indexOf(y.toString());
```
simolus3: We should try to keep our function implementations close to what SQLite would do (there are ideas to potentially use `node:sqlite` instead of JS implementations in the future). We already have a few functions that don't get this exactly right, but we should still try to keep SQLite compatibility with new functions.

Specifically, SQLite also does this:

>  Or, if X and Y are both BLOBs, then instr(X,Y) returns one more than the number of bytes prior to the first occurrence of Y, or 0 if Y does not occur anywhere within X.

kobiebotha: Implemented. I had to confirm in [the source](https://github.com/sqlite/sqlite/blob/7539d2f1ba7d30d6ec14ddb3c6cefdf6afbd88f2/src/func.c#L275) what the behavior is if the parameters are mixed types 😅

packages/sync-rules/src/sql_functions.ts
```diff
@@ -271,6 +271,37 @@ const iif: DocumentedSqlFunction = {
...
+// Otherwise, both arguments are interpreted as text.
+const instr: DocumentedSqlFunction = {
+  debugName: 'instr',
+  call(x: SqliteValue, y: SqliteValue) {
+    if (x == null || y == null) {
+      return null;
+    }
+    // Both BLOBs: byte-level search
+    if (x instanceof Uint8Array && y instanceof Uint8Array) {
+      const pos = Buffer.from(x).indexOf(Buffer.from(y));
```
simolus3: nit: Small optimization avoiding two copies:

```suggestion
      const pos = Buffer.from(x.buffer, x.byteOffset, x.byteLength).indexOf(y);
```

PR #568, opened 2026-03-17
Fix SQL errors being reported incorrectly in YAML source

### LINKED ISSUE #562: SQL error reporting location is wrong when queries span multiple lines
See the error reporting for the example below:

```yaml
config:
  edition: 3
streams:
  stream:
    auto_subscribe: true
    queries:
      - SELECT json_each.value AS "id" FROM json_each(subscription.parameters() ->> 'ids') WHERE something is broken
      - "SELECT json_each.value AS \"id\" FROM json_each(subscription.parameters() ->> 'ids') WHERE something is broken"
      - SELECT json_each.value AS "id"
        FROM json_each(subscription.parameters() ->> 'ids')
        WHERE something is broken
      - |
          SELECT id
          FROM unknown_table
```

Image

The current SQL parser can give a reasonable error location within the SQL string (see the first query, correctly highlighting the "broken" keyword). The issue is that the mapping of the SQL error location back to the YAML error location falls apart when the SQL string spans multiple lines, since the YAML parser removes the indentation automatically when parsing that string. It is not only an issue with indentation - there are other cases such as YAML escape characters that cause similar issues (see the examples using quotes above).

We're currently using the JS [yaml](https://www.npmjs.com/package/yaml) package for YAML parsing. We need to check whether that has the ability to map the position from the parsed string back to the source string, or if we perhaps need another parser or workaround for this issue.

The current code for doing the error location mapping is here:
https://github.com/powersync-ja/powersync-service/blob/fc0974f3847b33c2b2177404c09733efdd1b450f/packages/sync-rules/src/from_yaml.ts#L489-L510

Note the current workaround for handling a quoted yaml string, which only handles very simple cases.

LucDeCaf:
Fixes #562 by creating a map from a position in the parsed SQL query to a position in the source YAML file, which accounts for escape sequences, multiline blocks, and indentation.

### DIFF
--- a/.changeset/great-months-think.md
+++ b/.changeset/great-months-think.md
@@ -1,0 +1,6 @@
+---
+'@powersync/service-module-mongodb-storage': patch
+'@powersync/service-core': patch
+---
+
+[MongoDB Storage] Add option to explicitly enable/disable parameter compacting, and improve abort behavior when compacting.
--- a/.changeset/sixty-dolphins-look.md
+++ b/.changeset/sixty-dolphins-look.md
@@ -1,0 +1,6 @@
+---
+'@powersync/service-module-mongodb-storage': patch
+'@powersync/service-core': patch
+---
+
+[MongoDB Storage] Fix compact failing for buckets over 2GB in size.
--- a/modules/module-mongodb-storage/src/storage/implementation/MongoCompactor.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/MongoCompactor.ts
@@ -1,4 +1,4 @@
-import { mongo, MONGO_OPERATION_TIMEOUT_MS } from '@powersync/lib-service-mongodb';
+import { isMongoServerError, mongo, MONGO_OPERATION_TIMEOUT_MS } from '@powersync/lib-service-mongodb';
 import { logger, ReplicationAssertionError, ServiceAssertionError } from '@powersync/lib-services-framework';
 import {
   addChecksums,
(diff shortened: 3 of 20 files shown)


packages/sync-rules/src/from_yaml.ts
```diff
@@ -535,29 +535,160 @@ export class SyncConfigFromYaml {
...
-      // TODO: Is there a better way to do this?
-      sourceOffset += 1;
-    }
+    const srcToken = scalar.srcToken!;
+    // For block scalars (| and >), srcToken.offset points to the header character; skip past
+    // the header line to reach the first line of actual content.
+    const tokenStart = isBlockScalar(scalar.type) ? this.yaml.indexOf('\n', srcToken.offset) + 1 : srcToken.offset;
+    // For quoted scalars, the opening quote is part of the token but not the parsed value.
+    // tokenStart and valueStart are only different for quoted scalars.
+    const valueStart = isQuotedScalar(scalar.type) ? tokenStart + 1 : tokenStart;
```
rkistner: Is it feasible for this logic to also be included in the `buildParsedToSourceValueMap` function?

LucDeCaf: Yeah, probably

LucDeCaf: Actually maybe not, `valueStart` is only used in the fallback paths, not in `buildParsedToSourceValueMap` - I'll see if it's really necessary.

.changeset/silver-cups-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Fix SQL errors being reported incorrectly in YAML source
```
rkistner:
```suggestion
Fix SQL errors being reported in incorrect locations in YAML source
```

rkistner (approved):
Happy with the changes. Can fix the identified issues in a follow-up PR, since this is already way better than what we currently have.

PR #570, opened 2026-03-17
[MongoDB Storage] Compact using secondaries

rkistner:
This changes the compact process to read from secondaries (`readPreference: 'secondaryPreferred'`).

The main goal here is to reduce load on primary MongoDB nodes, by shifting these bulk reads to secondaries.

In general, reading stale data is not an issue for compacting. There are cases where we want to read our own writes, so we use a session with [causal consistency](https://www.mongodb.com/docs/manual/core/read-isolation-consistency-recency/#causal-consistency) for this. This specifically means setting the `readConcern: { level: 'majority' }` and `writeConcern: { w: 'majority' }` as well, to get all the guarantees we need.

We may be able to relax some of the guarantees later to improve throughput, for example by using a separate session for reads vs writes. That gets more complex though, so leaving that for another project.

We can also consider making the read preference configurable in the future, for example to allow allow these reads on secondaries, and rather fail if a secondary isn't available.

TODO:
 * [ ] Test on actual replicaset with replication lag.

### DIFF
--- a/modules/module-mongodb-storage/src/storage/implementation/MongoCompactor.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/MongoCompactor.ts
@@ -68,6 +68,24 @@ const DIRTY_BUCKET_SCAN_BATCH_SIZE = 2_000;
 /** This default is primarily for tests. */
 const DEFAULT_MEMORY_LIMIT_MB = 64;
 
+interface SessionOptions {
+  session: mongo.ClientSession;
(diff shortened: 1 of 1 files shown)


rkistner:
This gets quite tricky to do correctly, due to the compact process doing both reads and writes interleaved. Additionally, the compact process is changing significantly with storage version 3 now.

Using secondaries for reading checksums and sync data may give bigger gains versus the complexity.

PR #574, opened 2026-03-19
Convex replicator module

kobiebotha:
This adds support for [Convex](https://www.convex.dev/) as a replication source. Since Convex itself is open source (technically also FSL), it was quite feasible for me to implement this. 

As with any datastore, there are many quirks. I've attempted to document pertinent ones in a README.md in the module root. Required reading is the section titled "Mutation Transaction Atomicity".

To get a feel for the system, run the `convex` self-host-demo: [https://github.com/powersync-ja/self-host-demo/tree/convex-demo](https://github.com/powersync-ja/self-host-demo/tree/convex-demo). Then simply open two instances of the  Convex React demo app (new) ([http://localhost:3030](http://localhost:3030)) side by side. 

When running the demo, to log into the Convex dashboard, you need to jump through some hoops:

1. Check the container logs for the `convex-keygen` Docker container
2. Get the "Admin key" printed to console

TODO

- [x] Test against Convex Cloud (it has been a while since I did that) 
- [x] Fix replication metrics (currently reporting per transaction "page", should report per mutation?)
- [x] Update Test Connection logic to ensure that the `powersync_checkpoints` table and write mutation function exist
- [x] Measure replication performance
- [x] Docs https://github.com/powersync-ja/powersync-docs/pull/456


stevensJourney:
On the topic of benchmarks. I added a local development benchmark test in `modules/module-convex/test/src/bench/initial_replication.test.ts`

This suite:
- Connects to a local Convex backend
- Clears the db and seeds tables with varying amounts of data
- Performs an initial replication using the Convex snapshot APIs
- Records the time taken to perform the initial replication

The results for various tests were 

| Total Rows | Estimated Size (MiB) | Initial Replication Time (ms) | Estimated Rate (MiB/s) |
| ---: | ---: | ---: | ---: |
| 10,000 | 3.42 | 7,218 | 0.47 |
| 50,000 | 17.09 | 39,477 | 0.43 |
| 100,000 | 34.18 | 123,126 | 0.28 |

modules/module-convex/src/client/ConvexApiClient.ts
```diff
@@ -0,0 +1,470 @@
...
+  private async requestJson(options: {
+    endpoint: string;
+    params?: Record<string, unknown>;
+    signal?: AbortSignal;
+    allowStreamingFallback?: boolean;
+  }): Promise<Record<string, any>> {
+    await this.assertHostAllowed();
+
+    const primaryPath = `/api/${options.endpoint}`;
+    const fallbackPath = `/api/streaming_export/${options.endpoint}`;
```
stevensJourney: I could see this being used on either cloud or self hosted. It seems like the `/api/streaming_export` fallback routes don't get used. I've removed this for this reason.

stevensJourney:
As part of reviewing the initial POC, I went through the implementation in a bit more detail and then addressed a few items that seemed worth tightening up before this lands.

The main thing I spent time on was the replication flow itself. I reviewed the snapshot + delta approach, where we pin one global Convex snapshot boundary, snapshot selected tables at that same boundary, and then resume `document_deltas` from that stored LSN. I also looked at resumable snapshots and verified the behaviour by adding integration tests. 

I verified the write checkpoint flow. The important ordering here is that we read the Convex head, create the managed PowerSync write checkpoint in the callback, and only then write the Convex marker mutation. That marker write is what gives an idle Convex deployment a later observable delta so the checkpoint can actually be acknowledged. AI generated docs have been added to give more detail about this process. We can remove those if necessary, perhaps they might help other AI agents in the future. 

I reviewed the route API adapter as well, especially `createReplicationHead`, schema/debug table handling, and connection testing. I added a connection test check for the `powersync_checkpoints` table and mutator so misconfigured Convex projects fail earlier with a more useful message.

On the testing side, I added real integration tests against a local Convex backend and wired them into CI. These cover the module connection path, route API adapter, streaming replication, and resumable snapshots. Some of the original storage-mocking tests were AI-generated. They are less important now that we have real local Convex integration tests, but I left them in place since they might still have some use.

I also did a pass on the Convex API client and value conversion. The current Convex API response typings have been checked with cloud and self-hosted integration tests.

For Convex -> SQLite conversion, I verified the JSON schema responses received from Convex backends and made some cleanup improvements. This process has some limitations which are listed as known issues. I'll mention more on these in an upcoming docs PR.

If you'd like to take this for a spin, feel free to try the React Convex Todolist demo from https://github.com/powersync-ja/powersync-js/pull/952 - this uses a development PowerSync service image. Note that this demo will be moved to its own repository soon.

AI Usage disclaimer:
I believe most of the original implementation was AI generated. Most of my review improvements were hand coded. AI (Codex GPT-5.5 medium) was used to assist with the writing of integration tests - these tests were thoroughly debugged, tweaked and verified. The README content and docs pages are all AI generated. This code has been reviewed with multiple passes by Codex GPT-5.5 medium.

modules/module-convex/src/common/convex-to-sqlite.ts
```diff
@@ -0,0 +1,191 @@
...
+          `Convex ${jsonSchemaType} value must be a string, got ${typeof value}`
+        );
+      } else {
+        return value;
+      }
+    case SupportedJSONSchemaPropertyType.UNKNOWN:
+      // TODO! It seems like Convex might not report the schema value for values which have not
+      // been populated in the DB yet. This can cause many issues - and we need to work around this.
+      // We perform runtime checks and conversions at this point.
+      if (value == null) {
```
stevensJourney: After some additional thought, I’m leaning toward disabling `json_schemas` for SQLite row conversion in the Convex replication path.

The issue is that using schema metadata makes row values inconsistent depending on whether Convex happened to report a field in `json_schemas` at the time we cached it:

- `Int64` with schema metadata becomes `bigint`
- the same `Int64` without schema metadata stays the raw JSON string
- `Bytes` with schema metadata becomes a `Uint8Array`/blob`
- the same `Bytes` without schema metadata stays the raw base64 string

That inconsistency is probably worse than preserving the raw wire types. If we only use the types from `list_snapshot` / `document_deltas`, then behavior is stable:

- Convex `Int64` is always a string
- Convex `Bytes` is always a base64 string
- `number` / `float64` is always a JS number
- booleans come through as booleans, which are already accepted by the Convex-to-SQLite conversion layer

Then users can explicitly normalize ambiguous fields in Sync Streams rules, e.g. `CAST(points AS INTEGER)` for `Int64` columns. This is predictable and avoids cases where the same column changes type depending on whether a populated value existed when `json_schemas` was fetched.

I think we can still keep `json_schemas` for table discovery and admin/diagnostic schema reporting, but avoid using it to coerce replicated row values.

stevensJourney:
After making the above changes for JSON Schema usage, some additional improvements could be made to cater for schema changes.

After digging into how the Convex replicator actually uses source metadata, we’ve narrowed the schema-change problem down quite a bit.

For most of our other replicators, schema-change detection is important because it protects against things like replica identity changes, stale relation metadata, table renames/drops, and DDL changes that require a re-snapshot. Convex is different in a few important ways:

- `_id` is always the replication identity, so there is no replica-id drift to detect.
- Runtime row conversion does not use `json_schemas`; it uses the actual JSON document payload from `list_snapshot` and `document_deltas`.
- Field additions, removals, and type changes should therefore flow through normal document mutations/deltas.
- Convex data migrations are expected to be online document writes, so they should replicate as data changes rather than schema-triggered re-snapshots.

The one remaining question was wildcard table discovery. We added an integration test to verify that `json_schemas` lists schema-defined tables even when they contain no documents. That passed, which means initial wildcard expansion can discover empty tables up front. Based on that, the stream no longer needs to snapshot a table inline when it is first observed in `document_deltas`; if the table appears later in deltas, the delta payload is the source of truth.

Code/docs changes from this:
- Removed the Convex stream’s schema cache / forced schema refresh path.
- Exact table patterns now resolve directly from Sync Streams rules.
- Wildcards still use `json_schemas` for initial expansion.
- Newly observed selected tables in `document_deltas` are resolved and marked snapshot-complete, then the delta row is applied directly. No inline snapshot.
- Added/updated tests around exact table resolution, wildcard discovery, and empty schema-defined tables.
- Added `docs/convex/schema-change-handling.md` with the rationale and limitations.
- Updated the Convex README to reflect the new behavior.

One important limitation we validated: deleting a table from the Convex dashboard does not emit per-document `_deleted` rows in `document_deltas`. That means previously replicated rows can remain synced to clients. The docs now recommend using the dashboard “Clear Table” action before deleting a table, or deleting documents through mutation paths that emit document deltas. Otherwise, dashboard/schema-only table removal needs to be treated as a sync-rule/deployment state change where affected PowerSync state may need to be cleared or re-replicated.

docs/convex/schema-change-handling.md
```diff
@@ -0,0 +1,148 @@
...
+snapshot-complete at the current delta cursor.
+
+The replicator does not run an inline snapshot for that newly observed table.
+The reasoning is:
+
+1. existing wildcard-selected tables were already discovered through
+   `json_schemas` during initial replication,
+2. a table that appears later in `document_deltas` has appeared because a source
+   write occurred,
+3. the delta payload contains the full document state needed for replication,
```
Rentacookie, line 78: I'm wondering if this is still the case if a table full of data that is not in the sync streams is renamed to a table that matches the sync streams? 
If not then we end up in a situation where only new updates to the renamed table are added to the bucket storage.

stevensJourney: Yea, there certainly might still be some schema change edge cases out there. I should mention that proper schema change handling is an item scheduled for after the initial alpha release.

In this case though, I don't believe Convex supports a direct table rename operation. The recommended method is to create a new table and manually migrate the rows - which should create deltas.

modules/module-convex/src/replication/ConvexReplicationJob.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export class ConvexReplicationJob extends replication.AbstractReplicationJob {
+  private readonly connectionFactory: ConvexConnectionManagerFactory;
+  private lastStream: ConvexStream | null = null;
+
+  constructor(options: ConvexReplicationJobOptions) {
+    super(options);
+    this.connectionFactory = options.connectionFactory;
+    this.logger = defaultLogger.child({ prefix: `[powersync_${this.options.storage.group_id}] ` });
```
Rentacookie: The other replication jobs all have the logger set as the storage logger:
`this.logger = options.storage.logger;`

I think we could improve on that, but for consistency's sake this should probably  be made the same as the other replication jobs for now.

modules/module-convex/src/replication/ConvexStream.ts
```diff
@@ -0,0 +1,686 @@
...
+
+    let resumeFromLsn = batch.resumeFromLsn;
+    if (resumeFromLsn == null) {
+      throw new ReplicationAssertionError(`No LSN found to resume replication from.`);
+    }
+
+    // Resolve source tables up-front to warm table metadata and sync-rule matching.
+    await this.resolveAllSourceTables(batch);
+
+    let cursor = BigInt(resumeFromLsn);
```
Rentacookie: Is the cast to BigInt necessary? As I understand it resumeFromLsn is a string, so could that not be directly passed into the cursor value on line 158 instead?

modules/module-convex/src/replication/ConvexStream.ts
```diff
@@ -0,0 +1,686 @@
...
+    return resolved;
+  }
+
+  private async getOrResolveTables(
+    batch: storage.BucketStorageBatch,
+    tableName: string,
+    snapshotCursor: string
+  ): Promise<SourceTable[] | null> {
+    if (!this.isTableSelectedBySyncRules(tableName)) {
+      return null;
```
Rentacookie: This method could just return an empty array in cases like this to simplify the logic. The upstream call already checks for `(tables == null || tables.length == 0)`

modules/module-convex/src/replication/ConvexStream.ts
```diff
@@ -0,0 +1,686 @@
...
+
+    return false;
+  }
+
+  private async processTables(
+    batch: storage.BucketStorageBatch,
+    descriptor: SourceEntityDescriptor
+  ): Promise<SourceTable[]> {
+    const resolved = await batch.resolveTables({
+      connection_id: Number.parseInt(this.connections.connectionId) || 1,
```
Rentacookie: Might be worth storing the connectionId in the `ConvexConnectionManager` as  a number already with this default value.
At the moment if there is no connectionId it defaults to the string `default` in the manager which then defaults to `1` here.

modules/module-convex/src/types/types.ts
```diff
@@ -0,0 +1,104 @@
...
+import * as t from 'ts-codec';
+
+export const CONVEX_CONNECTION_TYPE = 'convex' as const;
+const DEFAULT_POLLING_INTERVAL_MS = 1_000;
+const DEFAULT_REQUEST_TIMEOUT_MS = 60_000;
+
+export interface NormalizedConvexConnectionConfig {
+  id: string;
+  tag: string;
+  type: typeof CONVEX_CONNECTION_TYPE;
```
Rentacookie: Doesn't look like the type field is used or necessary since `NormalizedConvexConnectionConfig` already implies that it is only for Convex.

modules/module-convex/src/replication/ConvexReplicationJob.ts
stevensJourney: That's a good catch. I've updated this to use the Storage's logger now.

modules/module-convex/src/replication/ConvexStream.ts
stevensJourney: This is a fair point. This was a bit janky due to handling both the resumeLSN string value from the storage and the bigint next page cursor received from the document deltas. If we instead track `cursor` as a string, then we only need to do a conversion from the next page cursor (bigint to string). I've made this change now.

modules/module-convex/src/replication/ConvexStream.ts
stevensJourney: This is better.

modules/module-convex/src/replication/ConvexStream.ts
stevensJourney: I've opted to implement the same logic used by the CDCStream, which is to resolve this in a small getter.

modules/module-convex/test/src/util.ts
Rentacookie, line 1: Is this still used? Looks like there is already a correctly implemented ConvexStreamTestContext in the test-utils folder.

stevensJourney: That's a good catch, this was a straggler from porting some of the initial tests. This test context and the `slow_tests` are actually covered by newer implementations, so they've been removed now.

modules/module-convex/src/types/types.ts
stevensJourney: Yeah this is fair. This was a result of some unclear type resolving. Some code was using `NormalizedConvexConnectionConfig` to satisfy `ResolvedConvexConnectionConfig`. Where `ResolvedConvexConnectionConfig` is an intersection of the Normalized and input config (making `type` required). I've removed the `type` field and then updated those usages accordingly.

modules/module-convex/src/module/ConvexModule.ts
```diff
@@ -0,0 +1,135 @@
...
+  async testConnection(config: types.ConvexConnectionConfig) {
+    this.decodeConfig(config);
+    const normalizedConfig = this.resolveConfig(this.decodedConfig!);
+    return await ConvexModule.testConnection(normalizedConfig);
+  }
+
+  static async testConnection(normalizedConfig: types.ResolvedConvexConnectionConfig): Promise<ConnectionTestResult> {
+    const connectionManager = new ConvexConnectionManager(normalizedConfig);
+    const missingMutatorErrorFragment = `
+Define a mutator for the PowerSync service to use
```
Rentacookie: I see the other modules extract his logic to a `checkSourceConfiguration` function. That function is then also used in the RouteAPIAdapter's `getConnectionStatus` implementation to return more info on the connection.

It is probably worth doing the same for this 🤔

Rentacookie (dismissed):
Looks good to me 😁

modules/module-convex/src/client/ConvexApiClient.ts
```diff
@@ -0,0 +1,281 @@
...
+
+    await new Promise<void>((resolve, reject) => {
+      this.config.lookup!(url.hostname, {}, (error) => {
+        if (error) {
+          reject(error);
+        } else {
+          resolve();
+        }
+      });
+    });
```
rkistner: Not a blocker, but this has some issues:
1. This does a DNS lookup on every request.
2. `fetch` itself may do another DNS lookup, which may return different results, bypassing validation.

For a better approach, see how `RemoteJWKSCollector` handles this using an `agent` for fetch:
https://github.com/powersync-ja/powersync-service/blob/3d465634fc4af2c1eeb5b2b9a460467c4d9b3814/packages/service-core/src/auth/RemoteJWKSCollector.ts#L142

rkistner (dismissed):
I did not review all the details, but the approach looks good. The info in `docs/convex` is very useful! Should add similar docs for other dbs over time.

modules/module-convex/src/client/ConvexApiClient.ts
stevensJourney: That is nicer. I've updated the implementation to use an HTTP agent now instead :)

rkistner (approved):

PR #601, opened 2026-04-10
Configurable storage versions

rkistner:
This adds `config.storage_version`, which can currently be 2 or 3.

For background, see:
 * #487

In most cases, the storage version should not be configured explicitly - the latest stable version will be used automatically. This adds a `config.storage_version` property to cater for some specific cases:
1. Downgrade: Say the current storage version is 4, but you want to downgrade to an older service version that only support up to version 2. To do that, first deploy with `config.storage_version: 2` in the sync config, wait for reprocessing to complete, then downgrade the service.
2. Experimental versions: Allows testing of storage changes in #585.
3. Holding back on an upgrade. In some cases, you may want to make sync config changes but stay on an older storage version.

## Version strategy

Together with this, I propose a storage version strategy:
1. Even numbers are always stable: Once support for an even number is added, future versions should continue supporting that storage version as-is, until officially deprecated.
3. Odd numbers are explicitly unstable: The underlying format may change at any time, and support for the version may be removed at any time. It should only be used for testing, never for production.

So right now we have version 2 as stable, and version 3 as unstable. Once all changes in version 3 has been finalized, we can stabilize the format as version 4, and remove support for version 3.

This strategy is not explicitly captured by the code yet - other than specifically marking version 2 as stable and 3 as unstable.

Version 1 is a special legacy case: This is used for all instances where the version was not defined. We support _using_ storage version 1, but the user cannot configure sync config for version 1. We could consider doing the same for other deprecated storage versions in the future.

## Implementation

The supported storage versions are now kinda duplicated between sync-rules and service-core packages. It does need to be available in sync-rules package, since we need to be able to validate the sync config without having access to service-core.

~We could consider merging the service-core functionality into that.~
This is now split in a way that the sync-rules package purely does validation and parsing a version number, while the service-core package assigns more specific behavior.


packages/sync-rules/src/StorageVersion.ts
```diff
@@ -0,0 +1,27 @@
+export interface ParsedStorageVersion {
```
simolus3: I think the duplication is fine for now, but maybe add a comment pointing to `service-core/src/storage/StorageVersionConfig.js` to make it clearer what a storage version is supposed to be.

simolus3 (approved):
Using this to downgrade storage versions is clever, I also like the approach of odd/even numbers for stability 👍 

I've added the product-visibility label because we probably want to document this when it gets released.

modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
```diff
@@ -156,7 +156,8 @@ export class MongoBucketStorage extends storage.BucketStorageFactory {
   }
 
   async updateSyncRules(options: storage.UpdateSyncRulesOptions): Promise<MongoPersistedSyncRulesContent> {
-    const storageVersion = options.storageVersion ?? storage.CURRENT_STORAGE_VERSION;
+    const storageVersion =
+      options.storageVersion ?? options.config.parsed.config.storageVersion?.version ?? storage.CURRENT_STORAGE_VERSION;
```
simolus3: It seems like `options.config.parsed.config` might not contain the storage version reliably when created through `PersistedSyncRulesContent.asUpdateOptions`?

I think we should set `storageVersion` in `PersistedSyncRulesContent.parsed()` (we should have it available there through `getStorageConfig()`) to make sure the field has a consistent value.

rkistner: Thanks, I missed that. I updated that now, and added some comments to explain the different cases.

packages/sync-rules/src/StorageVersion.ts
rkistner: I also tweaked this further now to make it clear it's only used for the initial validation. SyncConfig now only stores the number itself, not this interface anymore (partially to support the changes in `asUpdateOptions`).

simolus3 (dismissed):

simolus3 (approved):

PR #605, opened 2026-04-14
feat: experimental Azure Document DB support

Sleepful:
## Summary

Add experimental support for Azure Cosmos DB for MongoDB vCore (Azure DocumentDB) as a source database. Auto-detects Cosmos DB via `hello.internal.cosmos_versions` or `hello.internal.documentdb_versions` and applies workarounds for compatibility gaps in change stream behavior.

### Cosmos DB detection
- `detectCosmosDb()` method runs on both `initReplication()` and `streamChangesInternal()` — survives service restarts
- Guards `isdbgrid` and `setName` checks (Cosmos DB returns `msg: 'isdbgrid'` and no `setName`)
- Throws on post-images configuration (unsupported by Cosmos DB)

### wallTime as clusterTime substitute

Standard MongoDB change events include `clusterTime` — a BSON Timestamp with seconds and an increment (a monotonic counter within each second). Two operations in the same second get different increments (e.g., `Timestamp(1718457600, 1)` and `Timestamp(1718457600, 2)`), providing a total order. Cosmos DB change events don't include `clusterTime`. Instead, they provide `wallTime` — a `Date` used at second precision with the increment hardcoded to 0: `Timestamp.fromBits(0, Math.floor(wallTime.getTime() / 1000))`. All events within the same wall-clock second produce identical LSNs.

This loss of ordering precision directly drives two other changes in this PR:

- **The .lte() dedup guard must be disabled on Cosmos DB.** With `clusterTime`, events at the boundary have unique increments and `.lte()` correctly identifies duplicates. With second-precision wallTime, `.lte()` matches all events in the same second as the last checkpoint, silently dropping real data. (See [.lte() dedup guard](#lte-dedup-guard) below.)
- **Sentinel checkpoint matching uses document content instead of LSN comparison.** Multiple checkpoint events in the same second produce the same wallTime-derived LSN, making `>=` ambiguous for identifying a specific checkpoint. Content matching sidesteps the ordering problem entirely. (See [Sentinel checkpoint resolution](#sentinel-checkpoint-resolution) below.)

<details>
<summary>Call sites and files changed</summary>

`getEventTimestamp()` is a private method on `ChangeStream` in `modules/module-mongodb/src/replication/ChangeStream.ts`. It returns `clusterTime` on standard MongoDB and falls back to `wallTime` on Cosmos DB. It replaces direct `clusterTime!` access at these call sites, all in the same file:

- `getSnapshotLsn()` — LSN construction from the initial checkpoint event
- `streamChangesInternal()` — the `.lte(startAfter)` dedup guard
- `streamChangesInternal()` — LSN construction when a checkpoint event is committed
- `streamChangesInternal()` — checkpoint-out-of-order safety check (error message)
- `streamChangesInternal()` — periodic resume LSN persistence during catchup

`MongoRelation.ts` references the method in a comment explaining why sentinel mode uses `$inc` counters rather than relying on wallTime-derived LSNs for ordering.

</details>

### Sentinel checkpoint resolution

The replication loop in `streamChangesInternal` groups change stream events into batches. A batch is committed when a **checkpoint event** arrives — a document that PowerSync writes to `_powersync_checkpoints` and then observes coming back through the change stream. The variable `waitForCheckpointLsn` gates this: it holds a value after a checkpoint is created, and clears to `null` when the matching event arrives, triggering a commit. This creates a natural batching rhythm where data accumulates between checkpoints.

On standard MongoDB, the checkpoint is resolved by comparing LSN strings. `createCheckpoint` captures `session.operationTime` (the server-assigned timestamp of the write) and serializes it via `MongoLSN` into a hex string. When the same write appears on the change stream, the event carries `clusterTime` (the same server timestamp) plus a resume token. Both are serialized into the `MongoLSN` comparable format — a hex timestamp optionally followed by `|<base64 resume token>`. Since the checkpoint's LSN has no resume token suffix, the event's LSN is lexicographically greater for the same timestamp, and `>=` resolves correctly. It also resolves for any later checkpoint event, making the comparison defensive.

Cosmos DB breaks this: `session.operationTime` is not returned on regular commands, and `clusterTime` is absent from change events. Without a timestamp to capture or compare, the LSN comparison can't work. Instead, `createCheckpoint` returns a sentinel string (`sentinel:<id>:<i>`) encoding the document's `_id` and its `$inc` counter value. The streaming loop matches by deserializing `fullDocument` from the checkpoint event and checking for an exact content match on `_id` and `i`. After the sentinel resolves, the commit path is identical — `batch.commit(lsn)` is called with a wallTime-derived LSN in both cases.

<details>
<summary>Implementation details</summary>

- `createCheckpoint()` in `MongoRelation.ts` accepts `mode: 'lsn' | 'sentinel'`. In sentinel mode, it returns `sentinel:<id>:<i>` from the upserted document instead of reading `session.operationTime`.
- Branching in `streamChangesInternal` is via `waitForCheckpointLsn.startsWith('sentinel:')` — the same variable holds either a hex LSN or a sentinel string.
- `MongoLSN.comparable` serializes as `<8-hex-high><8-hex-low>` (timestamp only) or `<8-hex-high><8-hex-low>|<base64-bson>` (with resume token). The resume token is carried for stream resumption on restart, not for this comparison.
- Standard MongoDB resolution: `lsn >= waitForCheckpointLsn` where both sides share the same timestamp but differ in resume token suffix.
- Cosmos DB resolution: content match on `changeDocument.documentKey._id` + `BSON.deserialize(changeDocument.fullDocument).i`.
- After resolution (`waitForCheckpointLsn = null`), both paths fall through identically to `batch.commit(lsn, ...)`.

</details>

### Snapshot LSN acquisition

Before the initial snapshot, `getSnapshotLsn()` establishes a change stream position to resume from after the snapshot completes. It writes a checkpoint document and waits to observe it on the change stream — the event's LSN becomes the starting point for streaming.

On standard MongoDB, the checkpoint's `operationTime` is used with `startAtOperationTime` to open the stream from that exact point. On Cosmos DB, `startAtOperationTime` is not supported, so the stream opens from the current position (`lsn: null`). The checkpoint may have been written before the stream opened, in which case its event is missed. To handle this, the loop re-creates checkpoints every second for up to 60 seconds until one is observed. A later checkpoint producing a slightly later snapshot LSN is harmless — any changes between the missed and caught checkpoints are captured by the snapshot itself, which reads the full current state of all collections. Unlike the streaming loop's sentinel matching, `getSnapshotLsn` doesn't need content matching — it just looks for any checkpoint event from the right `checkpointId` and returns a wallTime-derived LSN.

<details>
<summary>Implementation details</summary>

- `getSnapshotLsn()` in `ChangeStream.ts` calls `detectCosmosDb()` first, then branches on `this.isCosmosDb` for checkpoint creation and stream options.
- On Cosmos DB: `createCheckpoint(mode: 'sentinel')` is called but the returned sentinel string is not used — `streamLsn` is set to `null` and the stream opens without `resumeAfter` or `startAtOperationTime`.
- The retry loop creates new checkpoints every `LSN_CREATE_INTERVAL_SECONDS` (1s) with `mode: this.isCosmosDb ? 'sentinel' : 'lsn'`.
- Resolution is by `checkpointId` match only (`this.checkpointStreamId.equals(checkpointId)`), not by sentinel content. The return value is a real `MongoLSN` comparable string, not a sentinel.
- Timeout is `LSN_TIMEOUT_SECONDS` (60s) — if no checkpoint is observed, a `PSYNC_S1301` error is thrown.

</details>

### Write checkpoint HEAD capture

Write checkpoints provide read-your-own-writes consistency for sync clients. After a client writes to the source database, replication may not have delivered the change to PowerSync's storage yet. A write checkpoint ensures the sync response includes all writes up to a known replication position, so the client doesn't see stale data that's missing its own recent changes.

The flow works by capturing a **replication HEAD** — the current position of the change stream — then creating a checkpoint that the sync stream resolves once replication advances past that HEAD. On standard MongoDB, `createReplicationHead` issues a `hello` command on a session, reads `session.clusterTime`, and passes it to the callback as the HEAD. This works because any subsequent change stream event is guaranteed to be at or after that cluster time.

Cosmos DB doesn't return `clusterTime` on the `hello` command. Instead, `createReplicationHead` writes a sentinel to `_powersync_checkpoints` (triggering a change stream event) and passes `null` as the HEAD. The caller — `createWriteCheckpoint` in `service-core` — detects the null and reads the current storage checkpoint LSN directly. This is valid because the storage LSN was committed before the sentinel was written, and the sentinel guarantees the streaming loop will advance past it, so the write checkpoint will eventually resolve.

<details>
<summary>Implementation details</summary>

- `MongoRouteAPIAdapter.createReplicationHead()` calls `detectCosmosDb()`, then either captures `session.clusterTime` (standard) or writes a sentinel via `findOneAndUpdate` with `$inc` and calls `callback(null)` (Cosmos DB).
- `createWriteCheckpoint()` in `checkpointing.ts` checks if HEAD is null. If so, it calls `syncBucketStorage.getCheckpoint()` to read the current storage LSN and uses that as the HEAD.
- `ReplicationHeadCallback` type was widened from `(head: string) => Promise<T>` to `(head: string | null) => Promise<T>` to support this.
- The v1 write checkpoint endpoint in `checkpointing.ts` throws on null HEAD (v1 doesn't support the polling fallback).

</details>

### Change stream configuration

On Cosmos DB:
- Uses cluster-level aggregate (`admin` db + `allChangesForCluster`) instead of database-level streams
- Skips `$changeStreamSplitLargeEvent` (not supported)
- Skips `showExpandedEvents` (not supported)
- Forces `fullDocument: 'updateLookup'` (post-images not supported)

<details>
<summary>Implementation details</summary>

All guards are in `rawChangeStreamBatches()` in `ChangeStream.ts`, which builds the `$changeStream` aggregation pipeline:

- **Cluster-level aggregate**: Cosmos DB does not support database-level change streams (`db.watch()`). The code uses `client.db('admin')` with `allChangesForCluster: true`, reusing the same path as multi-database replication on standard MongoDB. The `$match` filter on namespaces handles collection scoping.
- **`$changeStreamSplitLargeEvent`**: Pushed onto the pipeline only when `!this.isCosmosDb`. Without it, documents exceeding 16MB on Cosmos DB would fail — but this is an existing MongoDB limitation too, and Cosmos DB simply doesn't support the split workaround.
- **`showExpandedEvents`**: Set on `streamOptions` only when `!this.isCosmosDb`. This enables DDL events (create, rename, drop) on standard MongoDB 6.0+.
- **`fullDocument`**: Forced to `'updateLookup'` on Cosmos DB. Standard MongoDB can use `'required'` (which reads from `changeStreamPreAndPostImages`), but Cosmos DB doesn't support that collection option — even setting `enabled: false` throws. The `usePostImages` configuration is validated at startup and throws if enabled on Cosmos DB.

</details>

### .lte() dedup guard

When the streaming loop resumes from a stored LSN, standard MongoDB may use `startAtOperationTime` — a legacy path where the server can redeliver events at the boundary timestamp. The `.lte(startAfter)` guard skips events at or before the resume point to prevent duplicate processing. When `resumeAfter` is used instead (which includes Cosmos DB), the server guarantees no duplicates and the guard is unnecessary.

On Cosmos DB, the guard is actively harmful: `wallTime` has second precision (increment 0), so every event within the same wall-clock second as the last checkpoint would satisfy `.lte()` and be silently dropped — causing data loss after restart. The guard is skipped when `isCosmosDb` is true.

<details>
<summary>Implementation details</summary>

- The guard in `streamChangesInternal` is wrapped in `if (!this.isCosmosDb && startAfter != null && ...)`.
- A `try/catch` around the check handles events with neither `clusterTime` nor `wallTime` — the event is not skipped.
- The `resumeAfter` path is the default for Cosmos DB because resume tokens are always persisted with checkpoint LSNs.

</details>

### Keepalive guard

MongoDB's change stream API returns a **resume token** with every cursor batch — including empty batches from `getMore` when no events are available. This token is an opaque marker representing the cursor's current position in the stream. A new change stream opened with `resumeAfter: <token>` picks up from exactly that position.

When the change stream cursor returns empty batches (no data events), the streaming loop periodically persists the latest resume token to PowerSync's internal storage. This keepalive prevents the stored resume position from going stale — on standard MongoDB, a token older than the oplog retention window becomes unusable on restart (forcing a full re-snapshot), and on any backend a stale position means re-processing events unnecessarily.

On standard MongoDB, the keepalive path calls `parseResumeTokenTimestamp()` to extract the timestamp from the resume token's `_data` hex string, then constructs a `MongoLSN` with both the timestamp and the token for persistence. Cosmos DB resume tokens are opaque base64 blobs that can't be parsed by `parseResumeTokenTimestamp()`, which expects a specific hex byte layout. Instead, the keepalive path constructs the timestamp from `Date.now()` at second precision and pairs it with the raw resume token. The token is still persisted and usable for `resumeAfter` on restart — only the timestamp derivation differs.

<details>
<summary>Implementation details</summary>

- The keepalive branch in `streamChangesInternal` in `ChangeStream.ts` is gated by `if (this.isCosmosDb)`.
- Standard path: `MongoLSN.fromResumeToken(resumeToken)` — calls `parseResumeTokenTimestamp()` which reads bytes 1-8 of the hex `_data` field.
- Cosmos DB path: `new MongoLSN({ timestamp: Timestamp.fromBits(0, Math.floor(Date.now() / 1000)), resume_token: resumeToken })` — bypasses token parsing entirely.
- The `Date.now()` timestamp is structurally required by `MongoLSN` to construct a `comparable` string, but functionally unused on Cosmos DB — the resume token is what matters for stream resumption, and the Cosmos DB code paths don't use the timestamp for ordering or dedup.
- The keepalive is throttled by `keepaliveIntervalMs` (default 60s) and only fires when `waitForCheckpointLsn` is null (no pending checkpoint).

</details>

## Testing

> **Setup guide:** See [`modules/module-mongodb/test/COSMOS_DB_TESTING.md`](modules/module-mongodb/test/COSMOS_DB_TESTING.md) for environment variables, connection URI format, and test commands.

### When run against a MongoDB backend
59/59 existing tests pass. No regressions. 1 test skips, meant to be run against Cosmos DB (×3 storage versions).

### When run against a Cosmos DB backend
29/29 Cosmos DB tests pass (verified before the merge with upstream main). Tests include:
- Basic replication (insert, update, delete)
- Sentinel checkpoint resolution
- Write checkpoint flow
- Keepalive with Cosmos DB resume tokens
- Resume after restart
- Data events not dropped after restart (.lte() guard)

10 existing tests are skipped on Cosmos DB (unsupported features: post-images, large record split, drop/rename events, wildcard collections `test.skipIf(isCosmosDb)`).

### Post-merge status
After merging with upstream main (PR #591 raw change streams), standard MongoDB tests pass (59/59). Cosmos DB tests pass (30/30) after fixing two upstream Cosmos DB server changes — see [Cosmos DB upstream changes](#cosmos-db-upstream-changes) below.

## Files changed

<details>
<summary>14 files changed (click to expand)</summary>

| File | Change |
|------|--------|
| `modules/module-mongodb/src/replication/ChangeStream.ts` | Core Cosmos DB workarounds |
| `modules/module-mongodb/src/replication/MongoRelation.ts` | Sentinel mode for `createCheckpoint` |
| `modules/module-mongodb/src/replication/replication-utils.ts` | `isdbgrid` guard |
| `modules/module-mongodb/src/api/MongoRouteAPIAdapter.ts` | Cosmos DB detection + sentinel for `createReplicationHead` |
| `packages/service-core/src/util/checkpointing.ts` | Cosmos DB HEAD capture (direct storage read) |
| `packages/service-core/src/api/RouteAPI.ts` | `ReplicationHeadCallback` accepts `string \| null` |
| `packages/service-core/src/routes/endpoints/checkpointing.ts` | v1 endpoint null HEAD guard |
| `modules/module-postgres/test/src/util.ts` | Null guard for `lsn` after `ReplicationHeadCallback` type change |
| `modules/module-mongodb/test/src/cosmosdb_mode.test.ts` | Cosmos DB integration tests |
| `modules/module-mongodb/test/src/cosmosdb_helpers.test.ts` | Unit tests for helpers |
| `modules/module-mongodb/test/src/change_stream.test.ts` | skipIf guards for Cosmos DB |
| `modules/module-mongodb/test/src/change_stream_utils.ts` | keepaliveIntervalMs support |
| `modules/module-mongodb/test/COSMOS_DB_TESTING.md` | Test setup and run instructions |

</details>

## Cross-module impact

The `ReplicationHeadCallback` type in `packages/service-core/src/api/RouteAPI.ts` was changed from `(head: string) => Promise<T>` to `(head: string | null) => Promise<T>` to support the Cosmos DB sentinel flow. This required a null guard in:
- `packages/service-core/src/routes/endpoints/checkpointing.ts` (v1 endpoint)
- `packages/service-core/src/util/checkpointing.ts` (v2 write checkpoint)
- `modules/module-postgres/test/src/util.ts` (Postgres test utility)

The Postgres production code is unaffected — only the test utility needed a guard.

## Behavioral gaps from unsupported features

10 existing tests are skipped on Cosmos DB (`test.skipIf(isCosmosDb)`) because they exercise features Cosmos DB doesn't support. These gaps have not been verified independently against a live cluster — the risks below are inferred from the feature absence.

### Post-images (`changeStreamPreAndPostImages`) — 7 tests skipped

Cosmos DB doesn't support stored post-images, so `fullDocument: 'updateLookup'` is used instead. The difference: `updateLookup` re-queries the document at read time, while post-images store the document state at write time.

**Risk: Occasional missing update data.** If a document is updated and then quickly deleted (or updated again) before the `updateLookup` re-query executes, `fullDocument` is `null`. PowerSync skips the update — the client sees the previous state followed by the delete, missing the intermediate update. This would be a brief data staleness window, not permanent data loss — the data self-corrects once the delete (or next update) is processed.

### Large record splitting (`$changeStreamSplitLargeEvent`) — 1 test skipped

Cosmos DB doesn't support the pipeline stage that splits change events exceeding 16MB into fragments.

**Risk: Replication stall on oversized documents.** If a change event exceeds 16MB, the change stream would error. Replication restarts and hits the same error — an infinite restart loop until the document is reduced in size or the change ages out. No data loss for other documents, but replication would be blocked.

### DDL events (`drop`, `rename`, `showExpandedEvents`) — 2 tests skipped

Cosmos DB may not emit `drop` or `rename` events on the change stream.

**Risk: Stale phantom data.** If a collection is dropped, PowerSync would never learn about it. The dropped collection's data would persist in PowerSync's storage indefinitely, serving stale records to clients that no longer exist in the source. If a collection is renamed, PowerSync would continue tracking the old name — new writes under the new name are missed. Neither case would be self-correcting. Recovery would require restarting replication (which triggers a re-snapshot and reconciliation).

## Known limitations

- **wallTime precision**: See [wallTime as clusterTime substitute](#walltime-as-clustertime-substitute) above. Second-level precision means identical LSNs for events in the same second. Impact is negligible in practice.
- **maxAwaitTime not supported**: Cosmos DB returns empty batches immediately instead of long-polling. The streaming loop polls with the configured `maxAwaitTimeMS` which acts as a timeout on the raw `getMore`.
- **Cosmos DB test timeouts**: Tests use 50s poll deadline and 120s test timeout for remote clusters with variable propagation delay (10-30s spikes). Co-located deployments would be much faster.

## Dependencies

- PR #592 (URL-encoded credentials fix) is included in this branch

## Cosmos DB upstream changes

The Cosmos DB vCore dev cluster received a server-side update that broke two assumptions in this PR. Both were fixed:

1. **Detection field renamed**: The `hello` response changed from `internal.cosmos_versions` to `internal.documentdb_versions`. All three detection sites now check both field names for backwards compatibility.

2. **Resume token format changed**: Individual change event `_id` fields use a `_data` key that `resumeAfter` now rejects. The batch-level `postBatchResumeToken` uses a `_token` key that is accepted. On Cosmos DB, LSN construction now uses the batch-level `postBatchResumeToken` instead of `changeDocument._id`. Standard MongoDB is unaffected — both formats work there.


rkistner:
I started a doc on how exactly we use checkpoints and LSNs in PowerSync, partly to assist myself with reviewing this. We can eventually include it in the repo, but for now it's here: https://gist.github.com/rkistner/2fd45abec1ee6efabf778239afb238c5. Read that as background for some of my specific comments here.

I'm including review notes on the main sections in the PR description, as well as a proposal for restructuring the LSN format at the end.

# Review notes

## Cosmos DB detection

This looks fine for now.

> Throws on post-images configuration (unsupported by Cosmos DB)

We can revisit this later to perhaps just ignore the option - as I understand, CosmosDB behaves similar to if post-images is always enabled.

## wallTime as clusterTime substitute

See my proposal below to use `sentinal` instead of `wallTime` here.
I'm going through each of your main headings here:

## Sentinel checkpoint resolution / waitForCheckpointLsn

To recap for myself, we have this behavior:

For MongoDB:

1. Create/update a `_powersync_checkpoint` document, then get its `operationTime`.
2. Wait until we receive a `_powersync_checkpoint` document with a `resumeToken` timestamp >= the above `operationTime`.

New logic for CosmosDB:

1. Create/update a `_powersync_checkpoint` document with an auto-incrementing value as sentinel (actually the same we do for MongoDB).
2. Wait until we receive that exact value on the change stream.

In theory, we can do the same >= on the sentinel value in CosmosDB instead of strict equality check. But also the inverse: We could do an exact `operationTime ==` check with MongoDB. Except that I'm less confident that the `operationTime` in MongoDB would always match exactly between the value we got back from the new document, versus the value we get the the change stream. But in general, we have to assume that we'd never miss a value in the change stream. If we can miss values, it breaks all consistency properties.

My conclusion: The approach for CosmosDB here looks good. We could potentially even switch to the sentinel approach for standard MongoDB as well. There is no additional performance overhead, no realistic case it doesn't cover, and reduces reliance on `operationTime` / `clusterTime` (which are well-documented, but still feels like "internals").

## Snapshot LSN acquisition

The CosmosDB workaround looks reasonable here. It effectively reduces performance slightly, but this is not a frequently-called operation. In fact, we used the same approach for MongoDB until recently - we primarily optimized that for MongoDB to improve performance in tests.

For the LSN format itself, see my proposal later.

## Write checkpoint HEAD capture

This approach looks problematic:

> createWriteCheckpoint() in checkpointing.ts checks if HEAD is null. If so, it calls syncBucketStorage.getCheckpoint() to read the current storage LSN and uses that as the HEAD.

This is specifically an issue when replication lags behind: It would send the write checkpoint back to the client potentially long before replication has actually caught up. See my write-up covering write checkpoints above.

## Change stream configuration

Changes here make sense.

We should just test and document the limitations here, e.g. for lack of $changeStreamSplitLargeEvent support (as in test what actually happens when there is an event over the 16MB limit).

## .lte() dedup guard

Makes sense to disable this for CosmosDB.

We may also not need this for MongoDB anymore, since we're using raw change streams now (and the underlying issue was fixed in 6.20.0 of the MongoDB driver).

## Keepalive guard

> Instead, the keepalive path constructs the timestamp from Date.now() at second precision and pairs it with the raw resume token. The token is still persisted and usable for resumeAfter on restart — only the timestamp derivation differs.

That timestamp construction has potential to cause inconsistencies with regards to LSN ordering. See the proposal on LSN structuring below.

## Resume token format changed

> On Cosmos DB, LSN construction now uses the batch-level postBatchResumeToken instead of changeDocument._id.

This has the potential to skip events if replication is interrupted in the middle of a batch - the next try would skip all remaining changes in that batch.

One potential approach is to use the _previous_ batch's postBatchResumeToken. That could mean reprocessing some events if replication is interrupted, but that should be less problematic.

# Proposal for LSN structuring

I propose we restructure how we build "LSNs" completely for CosmosDB, taking the sentinel values further. Instead of using `${wallTime}|${resumeToken}`, instead use `${sentinel}|${resumeToken}`, where the `resumeToken` part is optional for write checkpoints.

- For resuming replication, we can use the `resumeToken` directly.
- For write checkpoints, we use the `sentinel` for ordering.
- For general ordering, ordering by `sentinel` first and then `resumeToken` should be fine, but we would need to double-check all the cases where we compare LSNs. Specifically, if the upstream `resumeToken` format is not sortable, we may need some extra workarounds here to only rely on `sentinal` for ordering / comparisons.
- For Snapshot LSN acquisition, we only really need the `resumeToken`, but could still use the `sentinel` for consistency.
- For the keepalive guard, we can leave the `sentinel` unchanged, and only update the `resumeToken`.

So for "Write checkpoint HEAD capture", we'd create a new sentinel value, and use only `${sentinel}` as the LSN. Note: we never resume replication from a write checkpoint - we only need to compare it to our current position in the replication stream, so we don't need the `resumeToken` here.

There are quite possibly edge cases in the implementation that I'm missing though .

stevensJourney:
After retesting against the latest DB instance. I verified that the `maxAwaitTime` issue has been fixed, I added an integration test which verifies that the call waits the relevant period before yielding an empty batch. It also seems like the `_id` field of Change stream events now can be used as a valid resume token. The implementation has been updated to take advantage of this. 

Most of the remaining updates are related to checkpointing. The proposed sentinel approach is now implemented for DocumentDB. We now support two different checkpoint/LSN mechanisms:

- clusterTime based `${timestamp}|${resumeToken}` the standard (current) MongoDB implementation
- sentinel based `${sentinel counter}|${resumeToken}` newly introduced implementation. This method could also be used by standard MongoDB connections, it's just not enabled at this stage.

The two methods are implemented by a checkpoint-strategy interface (CheckpointImplementation) with two implementations selected at runtime: the existing `TimestampCheckpointImplementation` for standard MongoDB, and a new `SentinelCheckpointImplementation` for DocumentDB. ChangeStream and the route adapter delegate every checkpoint decision to the strategy, instead of branching on `isCosmosDb` throughout the replication loop.

The DocumentDB LSN coordinate is a shared, monotonic counter (`_standalone_checkpoint.i`) advanced by writing to it and observed back through the change stream. Sentinel LSNs are serialized in the same 16-hex shape as a MongoDB timestamp (high 32 bits resemble epoch seconds, low 32 an increment), so they're directly string-comparable with timestamp LSNs. The counter is seeded in the seconds << 32 range, which guarantees a fresh sentinel LSN always sorts above any real-timestamp LSN issued in the past, and makes a re-created counter (after a consumer deletes the document) jump forward rather than reset. It only resembles a timestamp — the value is a synthetic counter and is never used as one (resume is purely via the opaque resume token).

All the changes are documented, in detail, in commited docs files:

All under `docs/documentdb/`:

- `documentdb-limitations.md` — user-facing limitations that differ from a standard MongoDB source (post-images, collection drop/rename, large-snapshot retention, large-document latency, source-database change detection).
- `documentdb-lsn-sentinel-checkpoints.md` — internal design notes: the checkpoint strategy interface, sentinel LSN encoding, change ordering, idle keepalive, standalone-counter persistence, and write-checkpoint observation.
- `documentdb-outstanding-items.md` — open design-level follow-ups (source-database change detection, large initial snapshot vs. change-feed retention).

---
AI Disclosure: The changes made here were assisted by Claude Opus 4.8. Changes were manually guided through many iterations of feedback. This work was reviewed by both Claude and Codex GPT5.5.

modules/module-mongodb/src/replication/checkpoints/CheckpointImplementation.ts
```diff
@@ -0,0 +1,234 @@
...
+   * (sentinel implementation: equal coordinate, opaque resume-token suffix tie —
+   * e.g. a restart replaying an event behind an already-persisted checkpoint, in
+   * which case the commit no-ops in storage / checkpointBlocked). Throws
+   * {@link ReplicationAssertionError} on a real ordering violation, stopping
+   * replication so a restart can recover.
+   *
+   * `lastCheckpointLsn` may be null (nothing committed yet), in which case there
+   * is nothing to compare against.
+   */
+  checkDescendingLsn(lsn: string, lastCheckpointLsn: string | null, doc: ProjectedChangeStreamDocument): void;
```
rkistner: This logic is removed in #670.

modules/module-mongodb/src/replication/MongoRelation.ts
```diff
@@ -208,25 +242,143 @@ async function createCheckpointInner(
...
+      // $setOnInsert cannot be combined with $inc on the same field, so this
+      // is a separate upsert; the loop then retries the increment. The
+      // $setOnInsert is a no-op if another process created the document
+      // concurrently.
+      await collection.updateOne(
+        {
+          _id: STANDALONE_CHECKPOINT_ID as any
+        },
+        {
+          $setOnInsert: { i: mongo.Long.fromBigInt(BigInt(Math.floor(Date.now() / 1000)) << 32n) }
```
rkistner:
```suggestion
          $setOnInsert: { i: BigInt(Math.floor(Date.now() / 1000)) << 32n }
```

The MongoDB driver supports BigInt everywhere for writes, and we use configuration so it also returns BigInt instead of Long everywhere.

modules/module-mongodb/src/replication/MongoRelation.ts
```diff
@@ -208,25 +242,143 @@ async function createCheckpointInner(
...
+    await session.endSession();
+  }
+}
+
+/**
+ * Normalize a numeric BSON value to bigint. The standalone counter is a 64-bit
+ * Long; the driver may return it as a Long, a promoted number (values <= 2^53),
+ * or a bigint (with useBigInt64).
+ */
+function longToBigInt(value: number | bigint | mongo.Long): bigint {
```
rkistner: We shouldn't need this, the driver returns the values as bigint already.

modules/module-mongodb/src/replication/checkpoints/SentinelCheckpointImplementation.ts
```diff
@@ -0,0 +1,291 @@
...
+    //    but embedding the value removes the dependency on cross-document
+    //    delivery order entirely.
+    const globalLsn = SentinelLSN.fromSerialized(
+      await createDocumentDbCheckpointLsn(this.context.client, this.context.db)
+    );
+    return createCheckpoint(this.context.client, this.context.db, this.context.checkpointStreamId, {
+      mode: 'sentinel',
+      globalSentinel: globalLsn.sentinel
+    });
+  }
```
rkistner, line 73: Can we change this to not require two writes for creating a batch checkpoint/barrier? I'm concerned about potential latency that this would introduce on every individual write that we replicate (not an issue for throughput on large batches, but can be an issue for latency on many individual writes).

And it might even be more than just two writes: The `createDocumentDbCheckpointLsn` here could trigger a separate `createBatchCheckpoint` here, resulting in 3 writes in for the barrier:
https://github.com/powersync-ja/powersync-service/blob/eefcee92be0084d0de5ba3196549f2b68ae98b02/modules/module-mongodb/src/replication/ChangeStream.ts#L775

The barrier should be completely private to the specific change stream, and does not need to persist / follow an incremental sequence after restarts. It does require a bit of a redesign though. I haven't thought through all the cases yet, but some initial thoughts:
1. Private batch/barrier checkpoints should not generate a "LSN" at all. Use a sequence completely separate from the LSN sequence, with no persistence. All we need is to write the document and check that we receive it back. In theory we could use the same approach with standard MongoDB without backwards-compatibility issues, since it's not persisted.
2. For the checkpoint LSN, we still need the global LSN as the combination of the global counter + resumeToken. If the resumeToken is incrementing, we don't need to increment the global counter for every new checkpoint - just updating the resumeToken part could be sufficient. But that's probably a big assumption that may not hold on MongoDB or DocumentDB.

So then there's another option: Can we use the `shared standalone checkpoint` as the barrier, and ignore the additional private doc? The only issue here is that we don't want to share the barriers between multiple replication processes replicating from the same source db. We can do this by tweaking the doc a bit.

```js
// Current
{ _id: '_standalone_checkpoint', i: <n> }
{ _id: <transient changestream id>, i: <n> }
// Proposal
{ _id: '_standalone_checkpoint', p: <transient changestream id>, i: <n> }
// Maybe also rename '_standalone_checkpoint' to something that makes more sense for this use case, if it doesn't affect backwards-compatibility?
```

The proposal above would allow preserving the global incrementing sequence, without needing a separate write for the private barriers.

modules/module-mongodb/test/DOCUMENTDB_TESTING.md
```diff
@@ -0,0 +1,107 @@
...
+There is also a **characterization test**, `does not report collection drop and rename events`,
+that documents the current limitation: it writes DDL plus a post-DDL marker, waits for the marker
+(proving the stream caught up), then asserts no `drop` / `rename` events were delivered. It passes
+today and **fails if a future DocumentDB engine starts delivering DDL events** — a signal to add
+real drop/rename replication support.
+
+## Known Issues
+
+- **Propagation delay**: DocumentDB has a variable delay between accepting a write and making it visible on the change stream cursor (internal propagation, not network latency). Against remote clusters this can take 10-30s during spikes. Tests use 50s poll deadlines and 120s test timeouts to handle this. Co-located deployments would be much faster.
+- **DocumentDB cluster availability**: The tests require a reachable DocumentDB cluster. If the cluster is down or unreachable (TLS timeout), all integration tests will fail with connection errors.
```
rkistner: This part feels kinda obvious, not really an "issue"?

modules/module-mongodb/test/DOCUMENTDB_TESTING.md
```diff
@@ -0,0 +1,107 @@
...
+
+There is also a **characterization test**, `does not report collection drop and rename events`,
+that documents the current limitation: it writes DDL plus a post-DDL marker, waits for the marker
+(proving the stream caught up), then asserts no `drop` / `rename` events were delivered. It passes
+today and **fails if a future DocumentDB engine starts delivering DDL events** — a signal to add
+real drop/rename replication support.
+
+## Known Issues
+
+- **Propagation delay**: DocumentDB has a variable delay between accepting a write and making it visible on the change stream cursor (internal propagation, not network latency). Against remote clusters this can take 10-30s during spikes. Tests use 50s poll deadlines and 120s test timeouts to handle this. Co-located deployments would be much faster.
```
rkistner: If this is internal propagation and not network latency, why would co-located deployments be much faster?

modules/module-mongodb/test/DOCUMENTDB_TESTING.md
```diff
@@ -0,0 +1,107 @@
...
+| `TEST_MONGO_STORAGE`      | No       | Set to `false` to skip MongoDB storage tests. Recommended when testing against DocumentDB to avoid using it as a storage backend.                                        |
+| `TEST_TIMEOUT_MULTIPLIER` | No       | Scale factor `testTimeout()` applies to a default timeout on DocumentDB when no explicit cloud override is given. Defaults to **6**. Increase for a slow/remote cluster. |
+
+### Connection URI format
+
+The `MONGO_TEST_DATA_URL` must include a database name in the path. DocumentDB URIs typically don't have one, so you need to add it before the query string:
+
+```
+# Original URI (no database):
+mongodb+srv://user:pass@cluster.mongocluster.cosmos.azure.com/?tls=true
```
rkistner: I believe `tls=true` should be the default for all `mongodb+srv` connections.

docs/documentdb/documentdb-limitations.md
```diff
@@ -0,0 +1,75 @@
...
+- A single large event blocks the change stream while it is being fetched, delaying every change behind it (including checkpoint advancement). Consistent checkpoints — and therefore write-checkpoint resolution — can lag by the fetch time.
+- For documents close to the 16 MiB limit, the fetch time may approach the change stream's socket/`maxTimeMS` budget, which would surface as a stream timeout and restart.
+
+If your workload includes large documents (especially frequently-updated ones), validate replication latency against your cluster.
+
+## Changing the source database is not detected
+
+On standard MongoDB, repointing a connection at a different source database invalidates the stored replication position and forces a resync. On DocumentDB the change stream is cluster-scoped, so the stored position stays valid when only the database name changes — replication silently continues against the new (typically empty) database instead of failing.
```
rkistner: > On standard MongoDB, repointing a connection at a different source database invalidates the stored replication position and forces a resync.

That is not quite true anymore - see the notes in https://github.com/powersync-ja/powersync-service/pull/609 (Flush section of the PR description).

docs/documentdb/documentdb-limitations.md
```diff
@@ -0,0 +1,75 @@
...
+
+If you change the source database for an existing connection, trigger a resync manually.
+
+## Throughput and latency
+
+DocumentDB's change stream historically delivered events more slowly than standard MongoDB (one operation per poll, with idle polls returning immediately). Microsoft has since added long-polling (`maxAwaitTime`) and event batching. Throughput and latency for high-write workloads should be validated against your target cluster, as these capabilities depend on the DocumentDB version in use.
```
rkistner: We probably don't need the historical notes here.

docs/documentdb/documentdb-outstanding-items.md
```diff
@@ -0,0 +1,56 @@
...
+>
+> - User-visible behaviour differences live in [documentdb-limitations.md](./documentdb-limitations.md).
+> - The checkpoint/LSN design lives in [documentdb-lsn-sentinel-checkpoints.md](./documentdb-lsn-sentinel-checkpoints.md).
+> - This file is for _open_ items: known correctness gaps, detection gaps, and doc corrections still pending.
+
+## Summary
+
+| #   | Item                                                                   | Kind          | Severity | Fixable?                                    |
+| --- | ---------------------------------------------------------------------- | ------------- | -------- | ------------------------------------------- |
+| 1   | Changing the source database is not detected                           | Detection gap | Medium   | Yes (future work)                           |
```
rkistner: As noted in the other comment, we don't have this for standard MongoDB anymore either.

modules/module-mongodb/src/replication/replication-utils.ts
```diff
@@ -7,16 +8,36 @@ export const CHECKPOINTS_COLLECTION = '_powersync_checkpoints';
...
 const REQUIRED_CHECKPOINT_PERMISSIONS = ['find', 'insert', 'update', 'remove', 'changeStream', 'createCollection'];
 
+/**
+ * Whether a `hello` response indicates Azure DocumentDB (formerly Azure Cosmos
+ * DB for MongoDB vCore). Older clusters report `cosmos_versions`; newer ones
+ * report `documentdb_versions` after Microsoft's rename.
+ */
+function isDocumentDbHello(hello: mongo.Document): boolean {
+  return hello.internal?.cosmos_versions != null || hello.internal?.documentdb_versions != null;
+}
```
rkistner, line 18: Do we (or do we need to) actually support the older clusters with `cosmos_versions`?

modules/module-mongodb/src/replication/MongoRelation.ts
stevensJourney: Thanks! I've fixed this

modules/module-mongodb/test/DOCUMENTDB_TESTING.md
stevensJourney: Indeed. I've removed this section :)

modules/module-mongodb/test/DOCUMENTDB_TESTING.md
stevensJourney: Fair point. I've updated the docs to remove `tls=true`

docs/documentdb/documentdb-limitations.md
stevensJourney: Good catch! I've updated the wording here and removed this as a DocumentDB limitation.

docs/documentdb/documentdb-limitations.md
stevensJourney: Removed

modules/module-mongodb/src/replication/replication-utils.ts
stevensJourney: I don't think we need to support older versions - especially since we currently rely on some updates in the latest version, supporting the older version is probably out of the question for now. I've simplified this check.

modules/module-mongodb/src/replication/checkpoints/SentinelCheckpointImplementation.ts
stevensJourney: Yeah, having two writes also seemed a little odd to me.. I like the shared record implementation above, went with that approach. I've updated the ID for sentinel based checkpoints to `_sentinel_checkpoint` now.

A batch checkpoint is a single write now: `{ $inc: { i: 1 }, $set: { stream_id: <checkpointStreamId> } }`. Basically your `p` field, just named `stream_id` for readability. Standalone bumps (write checkpoint heads, snapshot markers) clear it to `null`, so every stream still observes those as the global coordinate. Barrier resolution is by content: an event resolves a stream's barrier when `fullDocument.stream_id == ourStreamId && fullDocument.i >= marker`, so barriers stay private to the stream that created them and aren't shared between processes replicating the same source db. The global counter `i` is still a single shared monotonic sequence, so this drops the extra write entirely.

Right now sentinel checkpoints are only used for DocumentDB, so we don't technically have any backwards compatibility concerns yet - but a forward migration from existing standard MongoDB instances should generally be fine too. The one caveat is that switching schemes seeds the new record from the current timestamp (assuming clocks are accurate), so it's best done while replication is caught up - if you switch mid-backlog the first barrier sits out at "now" and you get one big stall-then-jump on the first checkpoint (no data loss or commit-ahead, just no checkpoint progress until it catches up).

A round-trip (clusterTime → sentinel → back to clusterTime) is also fine. The sentinel's high 32 bits are frozen at seed time and only tick up by a single increment per checkpoint, whereas a real `clusterTime` carries the current wall-clock seconds, so switching back lands far ahead of the sentinel value rather than behind it.

modules/module-mongodb/src/replication/MongoRelation.ts
stevensJourney: Thanks! I've removed this.

modules/module-mongodb/test/DOCUMENTDB_TESTING.md
stevensJourney: Unfortunately I don't have the full context for this. This section here was added before my [time](https://github.com/powersync-ja/powersync-service/commit/70b9c9c19c19096f20c009ddc81f8ea7fb7071e0). Happy to remove this if this seems incorrect.

rkistner: Yeah, I think just remove. Can add again if we do more testing.

modules/module-mongodb/src/replication/checkpoints/SentinelCheckpointImplementation.ts
```diff
@@ -0,0 +1,282 @@
...
+    // one is observed.
+    await this.createBatchCheckpoint();
+    return null;
+  }
+
+  async keepalive(_batch: storage.BucketStorageBatch, _resumeToken: mongo.ResumeToken): Promise<void> {
+    // Advance the shared sentinel, but do not persist a checkpoint here. The
+    // bump's own change event will flow through the stream and be committed by
+    // the standalone checkpoint handling, which advances the ordered LSN
+    // prefix and refreshes the resume token (using the event's own token).
```
rkistner: This comment is slightly outdated now - it's not handled by the "standalone checkpoint handling" anymore.

rkistner (commented):
I'm happy with the implementation now

modules/module-mongodb/src/replication/checkpoints/CheckpointImplementation.ts
rkistner: Since #670 is merged now, I think you can just remove the `checkDescendingLsn` call and implementation.

modules/module-mongodb/src/replication/checkpoints/SentinelCheckpointImplementation.ts
stevensJourney: Good catch. I've updated the comment

modules/module-mongodb/src/replication/checkpoints/CheckpointImplementation.ts
stevensJourney: Jip :) I've updated this branch with `main` and removed `checkDescendingLsn`.

modules/module-mongodb/test/DOCUMENTDB_TESTING.md
stevensJourney: Removed now

rkistner (dismissed):

docs/documentdb/documentdb-limitations.md
```diff
@@ -0,0 +1,65 @@
+# Azure DocumentDB — Limitations
+
+PowerSync can replicate from Azure DocumentDB (formerly Azure Cosmos DB for MongoDB vCore), but it is not fully MongoDB-compatible. This document lists the limitations that **affect users** — behaviour that differs from a standard MongoDB source in a way a deployment can observe.
+
+Internal implementation details (how checkpoints and LSNs work on DocumentDB) are documented separately in [documentdb-lsn-sentinel-checkpoints.md](./documentdb-lsn-sentinel-checkpoints.md).
+
+> **DocumentDB support is experimental.** Validate it against your own workload before relying on it in production.
```
benitav:
```suggestion
> **DocumentDB support is experimental.** APIs and behavior may change. We also can't yet guarantee continued support or long-term stability. This release is intended for early testing and to invite feedback. Your feedback will directly influence whether, and how, this integration evolve.
```

docs/documentdb/documentdb-limitations.md
```diff
@@ -0,0 +1,65 @@
+# Azure DocumentDB — Limitations
+
+PowerSync can replicate from Azure DocumentDB (formerly Azure Cosmos DB for MongoDB vCore), but it is not fully MongoDB-compatible. This document lists the limitations that **affect users** — behaviour that differs from a standard MongoDB source in a way a deployment can observe.
```
benitav: > but it is not fully MongoDB-compatible.

DocumentDB itself isn't 100% MongoDB compatible right? It might be good to rephrase this to make clear whether this incompatibility is PowerSync-specific or due to DocumentDB limitations (so likely not to change).

stevensJourney: Correct, there are certain differences between DocumentDB and MongoDB which we cater for internally, but some do (still) affect PowerSync users - this currently mentioned with `behaviour that differs from a standard MongoDB source in a way a deployment can observe`

I'm open to suggestions here :)

benitav: > PowerSync can replicate from Azure DocumentDB (formerly Azure Cosmos DB for MongoDB vCore). DocumentDB speaks the MongoDB wire protocol, so PowerSync connects to it through the same MongoDB connector used for standard MongoDB sources. DocumentDB is a separate engine, though, and doesn't implement every MongoDB feature. PowerSync handles most of these gaps internally. This page covers the ones that may still affect deployments.

maybe something like this?

stevensJourney: Thanks! That sounds good to me

modules/module-mongodb/src/replication/ChangeStream.ts
```diff
@@ -482,42 +531,54 @@ export class ChangeStream {
...
       watchDb = this.defaultDb;
     }
 
+    const maxAwaitTimeMS = options.maxAwaitTimeMS ?? this.maxAwaitTimeMS;
+
     return rawChangeStream(watchDb, pipeline, {
       batchSize: options.batchSize ?? this.snapshotChunkLength,
-      maxAwaitTimeMS: options.maxAwaitTimeMS ?? this.maxAwaitTimeMS,
+      maxAwaitTimeMS,
+      clientSideMaxAwaitTimeMS: this.isDocumentDb && maxAwaitTimeMS > 0,
```
rkistner, line 588: Can `maxAwaitTimeMS` ever be 0? What happens in that case?

modules/module-mongodb/src/replication/RawChangeStream.ts
```diff
@@ -212,31 +227,34 @@ async function* rawChangeStreamInner(
...
+        collection: nsCollection,
+        batchSize: batchSizer.next()
+      };
+      // Azure DocumentDB currently treats getMore maxTimeMS as a hard command
+      // timeout instead of just the idle maxAwaitTimeMS, which can fail slow
+      // batches with "Query exceeded command timeout of 200ms". In client-side
+      // mode, don't supply it and enforce the idle wait locally for empty batches below.
+      if (!options.clientSideMaxAwaitTimeMS) {
+        getMoreCommand.maxTimeMS = options.maxAwaitTimeMS;
+      }
```
rkistner: Can we always set the `maxTimeMS` here, and just additionally use the client-side delay? That should keep it forwards-compatible for when DocumentDB adds proper `maxTimeMS` support.

We use a default of 10s for `maxAwaitTimeMS`, which should be sufficient for slow batches to finish in most cases. For the individual cases where it's not enough, the existing retry logic should cover it. So as I understand, the main issue is just that DocumentDB currently returns immediately with empty batches, rather than waiting `maxTimeMS` for more data to arrive? In which case it should not hurt to still specify it.

modules/module-mongodb/src/replication/ChangeStream.ts
stevensJourney: It is currently set to zero when getting the snapshotLSN. I haven't seen any issues with this in testing, yet. Generally - in this workaround, we would not add any additional delay on the "server/client" side in this case.

modules/module-mongodb/src/replication/RawChangeStream.ts
stevensJourney: This is good feedback.

I started out by always passing the value through and that does work. I did notice in testing though that the default 10 second period then makes latency for updates, between clients, very bad. I've updated this to have a max "client side delay" for empty batches of 1 second. That way we can still wait 10 seconds for large batches, but only delay for 1 second between empty batches.

.github/workflows/documentdb-integration.yml
```diff
@@ -0,0 +1,64 @@
+name: Azure DocumentDB Integration Tests
```
stevensJourney, line 1: I temporarily enabled this workflow `on_push` in order to test it. The tests all passed [here](https://github.com/powersync-ja/powersync-service/actions/runs/28513908988/job/84520701211)

rkistner (approved):

PR #608, opened 2026-04-16
Support circular references in Sync Streams

simolus3:
For Sync Streams, building a graph of parameter lookups is implemented by going through expressions and recursively handling dependencies. E.g. if a stream has outputs of a table `a` and we encounter a filter expression `a.foo = b.bar`, we'd try to resolve `b` as a parameter lookup recursively before looking at additional expressions. This generally works well, because it both resolves partition keys and parameter outputs while sorting the flat list of expressions topologically, giving us a description for a querier.

Because of this recursion, we can't support circular references between result sets. Initially, I believed that circular references are inherently unsupported (we need to look up parameters in sequence, after all). Then I looked at an actual instance of this in the wild, and realized that the circular reference isn't really all that circular:

```SQL
-- This currently crashes the compiler
SELECT invoices.* FROM invoices, stores WHERE invoices.store_name = stores.name AND invoices.region = stores.region AND stores.owner = auth.user_id()
```

Currently, we start by analyzing `invoices`, track `stores` as a dependency due to `invoices.store_name = stores.name`. When parsing `stores`, we discover `invoices.region = stores.region` and crash due to a circular reference. The fix is very simple: Ignore expressions referencing a result set that is currently being analyzed, so we'd then:

1. Start by analyzing `invoices`, track `stores` as a dependency due to `invoices.store_name = stores.name`.
2. Resolve `stores`, find `stores.owner = auth.user_id()` as the only expression and create a pending parameter lookup partitioned by `owner` without any outputs.
3. Resolve the `invoices.store_name = stores.name` expression by adding `stores.name` as an output to the lookup and `invoices.store_name` as a bucket parameter.
4. Encounter the `invoices.region = stores.region` expression in the context of `invoices`, realize we've resolved `stores` already and just do the same add output + bucket parameter step with the resolved parameter lookup.

The same strategy works for larger reference chains as well.

### DIFF
--- a/.changeset/pretty-ties-judge.md
+++ b/.changeset/pretty-ties-judge.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Support most circular references in join conditions for Sync Streams (the compiler would crash on those before).
--- a/.changeset/pretty-ties-judge.md
+++ b/.changeset/pretty-ties-judge.md
@@ -2,4 +2,4 @@
 '@powersync/service-sync-rules': patch
 ---
 
-Support most circular references in join conditions for Sync Streams (the compiler would crash on those before).
+Support multiple references between the same tables as join conditions for Sync Streams (the compiler would previously crash due to an "internal circular reference error").
--- a/packages/sync-rules/test/src/compiler/advanced.test.ts
+++ b/packages/sync-rules/test/src/compiler/advanced.test.ts
@@ -334,5 +334,13 @@ streams:
         `)
       ).toMatchSnapshot();
     });
+
+    test('table-valued', () => {
+      expect(
+        compileSingleStreamAndSerialize(`SELECT a.*
+          FROM a, b, json_each(a.tags) AS j
+          WHERE a.k = b.k AND j.value = b.v`)
+      ).toMatchSnapshot();
+    });
   });
 });
--- a/packages/sync-rules/test/src/compiler/advanced.test.ts
+++ b/packages/sync-rules/test/src/compiler/advanced.test.ts
@@ -313,4 +313,26 @@ streams:
 
     expect(serializeSyncPlan(plan)).toMatchSnapshot();
   });
+
+  describe('circular references', () => {
+    test('at first level', () => {
+      expect(
+        compileSingleStreamAndSerialize(
+          'SELECT a.* FROM a, b WHERE a.c1 = b.c1 AND a.c2 = b.c2 AND b.u = auth.user_id()'
+        )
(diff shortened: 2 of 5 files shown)


.changeset/pretty-ties-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Support most circular references in join conditions for Sync Streams (the compiler would crash on those before).
```
rkistner, line 5: As noted in the PR description, most of these cases are not actually circular references: In the implementation these may have been processed as circular references, but conceptually in the query it's just multiple references in the same set of tables. Can we update the changeset to reflect this?

packages/sync-rules/test/src/compiler/advanced.test.ts
```diff
@@ -313,4 +313,26 @@ streams:
...
+
+    test('nested', () => {
+      // This introduces two bucket parameters (for a.c1 and a.c4). We create a lookup on
+      //   - Table d (lookup index 0), output columns c3 and c4, no partitions.
+      //   - Table c (lookup index 1), output column c2 and partition by c3.
+      //   - Table b (lookup index 2), output column c1 and partition by c2.
+      expect(
+        compileSingleStreamAndSerialize(`SELECT a.* FROM a, b, c, d WHERE
+          a.c1 = b.c1 AND b.c2 = c.c2 AND c.c3 = d.c3 AND d.c4 = a.c4
+        `)
```
rkistner, line 334: The test verifies that this can now compile and that the output doesn't change, but it's quite difficult to verify that the generated plan would actually do what we expect.

Could you add an "integration test" for this (or a similarly-shaped query) that tests that the resulting set of buckets (`queryDynamicBucketDescriptions`) is what we'd expect?

rkistner (commented):
Another case still failing (picked up by Codex):

```sql
SELECT a.*
    FROM a, b, json_each(a.tags) AS j
    WHERE a.k = b.k AND j.value = b.v
```

I'm not sure if it is actually feasible or useful to support this case, but maybe we can have a more descriptive error message?

packages/sync-rules/test/src/compiler/advanced.test.ts
simolus3: I've added a test matching generated buckets, parameters and queriers 👍

simolus3:
> I'm not sure if it is actually feasible or useful to support this case, but maybe we can have a more descriptive error message?

We should be able to support that as well, I've fixed that and added a test 👍

rkistner (approved):

PR #619, opened 2026-04-29
[MongoDB Storage V3] Replication streams

rkistner:
This introduces a split between "sync config definition" and "replication stream status" in v3 storage. Previously there were a 1:1 mapping between the two, so it used the same SyncRulesDocument in the sync_rules collection, but with incremental reprocessing that's changing.

Overall, the goal is to have a single replication stream that can process multiple sync config definitions. We may also have multiple replication streams like we do currently, for example if they're using different storage versions, or different compatibility options, in which case a single stream can't handle both.

This does not yet implement the capability to handle multiple sync config definitions in the same stream, but it refactors the model to support it in the future.

This keeps the "sync_rules" collection as the one responsible for the status and locks of the stream - this keeps the loops to check which streams are active simple. The name is now outdated, but that would be disruptive to change.

The new `sync_config` collection is treated as "immutable" documents only keeping the config itself. Splitting this collection out serves two purposes:
1. It keeps the sync_rules collection smaller, which reduces overhead when it gets many updates (can be multiple per second).
2. It helps keep past versions around if we need to inspect them, without risking ever-growing `sync_rules` documents.

The `sync_rules` collection keeps track of the status of every active sync_config handled by it, as a nested array.

A lot of the code changes here are just from the V1 and V3 storage implementations now diverging further, which requires "duplicating" more code.

## Other fixes

Updating sync config is now a transactional process, including taking the lock. This helps avoid potential race conditions.

## AI usage

This PR was AI-assisted: A large portion of the code was written or refactored using Codex (gpt-5.5 medium/high), but manually directed, and manually reviewed and tested.


stevensJourney (dismissed):

rkistner:
@stevensJourney Codex code review picked up a couple of issues - I fixed and added more tests.

modules/module-mongodb-storage/src/storage/implementation/v1/MongoBucketBatchV1.ts
```diff
@@ -1,15 +1,21 @@
...
 import { SourceRecordStoreV1 } from './SourceRecordStoreV1.js';
 import { VersionedPowerSyncMongoV1 } from './VersionedPowerSyncMongoV1.js';
+import { SyncRuleDocumentV1 } from './models.js';
 
 export class MongoBucketBatchV1 extends MongoBucketBatch {
   declare public readonly db: VersionedPowerSyncMongoV1;
 
   private readonly store: SourceRecordStore;
+  private needsActivation = true;
+  private lastWaitingLogThottled = 0;
```
Sleepful:
```suggestion
  private lastWaitingLogThrottled = 0;
```

:)

stevensJourney (dismissed):

stevensJourney (approved):
Looks solid to me :)

PR #621, opened 2026-04-30
Implement limit for parameter results

simolus3:
Currently, limits for parameters are checked after we've fetched parameter results from bucket storage. So despite the fact that we have a 1k default limit, we might load millions of parameter results for pathological Sync Streams before we emit an error. This typically leads to an OOM on the instance before we're able to show a warning.

As a solution, this implements a limit when loading results and prints an early breakdown of what went wrong when too many parameters are returned. There's an existing check counting parameter results on buckets (before deduplication). That check claims to count parameter results, but it doesn't really do that:

- For sync streams, not every parameter result is a bucket since one can combine parameter queries with e.g. table-valued functions to generate many buckets without that many parameter lookups.
  - This is still a problem for sync rules: Since all queriers are merged into one, it's possible to get a "too many parameters" error even with a single parameter result if the other buckets are contributed by static buckets.

Since we're now able to track parameters properly, we can remove that check. It still makes sense to print a breakdown for users when a bucket limit is exceeded, so I've moved that check to run if the amount of de-duplicated buckets exceeds the allowed limit instead.

To avoid having to load everything upfront before checking results, we need to forward a remaining limit to bucket storage. To still be able to provide a breakdown of used parameters, this instruments calls to `getParameterSets` in `BucketParameterState` to add include information about the stream responsible for resolving parameters. This still isn't great, there are potential race conditions since sync streams will invoke it concurrently without waiting for a previous result. But even in the worst case, this creates an upper bound on the amount of parameter rows we'd fetch.

While this PR is an improvement, our error reporting for too many parameters and too many buckets is really rough when Sync Streams are involved. I don't think it's possible to make this substantially better with the `Querier` interface because the information of what's actually causing a parameter fetch (ideally, we'd like to know about the stream _subscription_ causing this instead of just the stream) isn't available where we need it. The entire `BucketParameterState` logic can probably be improved a lot for Sync Streams since we can cache lookups across streams and even connections by inspecting sync plans. Doing that would be a good opportunity to add first-class support for these errors and to track them with much more context.

### DIFF
--- a/modules/module-mongodb-storage/src/storage/implementation/v1/MongoSyncBucketStorageV1.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/v1/MongoSyncBucketStorageV1.ts
@@ -6,6 +6,7 @@ import {
   GetCheckpointChangesOptions,
   InternalOpId,
   internalToExternalOpId,
+  ParameterSetLimitExceededError,
   ProtocolOpId,
   storage,
   utils
@@ -96,9 +97,10 @@ export class MongoSyncBucketStorageV1 extends MongoSyncBucketStorage {
 
   protected getParameterSetsImpl(
     checkpoint: MongoSyncBucketStorageCheckpoint,
-    lookups: ScopedParameterLookup[]
+    lookups: ScopedParameterLookup[],
+    limit: number | undefined
   ): Promise<SqliteJsonRow[]> {
-    return getParameterSetsV1(this.versionContext, checkpoint, lookups);
+    return getParameterSetsV1(this.versionContext, checkpoint, lookups, limit);
   }
 
   protected getBucketDataBatchImpl(
@@ -195,47 +197,53 @@ export class MongoSyncBucketStorageV1 extends MongoSyncBucketStorage {
 export async function getParameterSetsV1(
   ctx: MongoSyncBucketStorageContext<VersionedPowerSyncMongoV1>,
   checkpoint: MongoSyncBucketStorageCheckpoint,
-  lookups: ScopedParameterLookup[]
+  lookups: ScopedParameterLookup[],
+  limit: number | undefined
 ): Promise<SqliteJsonRow[]> {
   return ctx.db.client.withSession({ snapshot: true }, async (session) => {
     setSessionSnapshotTime(session, checkpoint.snapshotTime);
     const lookupFilter = lookups.map((lookup) => {
       return storage.serializeLookup(lookup);
     });
-    const rows = await ctx.db.parameterIndexV1
-      .aggregate(
-        [
-          {
-            $match: {
-              'key.g': ctx.group_id,
-              lookup: { $in: lookupFilter },
-              _id: { $lte: checkpoint.checkpoint }
-            }
-          },
-          {
-            $sort: {
-              _id: -1
-            }
-          },
-          {
-            $group: {
-              _id: { key: '$key', lookup: '$lookup' },
-              bucket_parameters: {
-                $first: '$bucket_parameters'
-              }
-            }
-          }
-        ],
-        {
-          session,
-          readConcern: 'snapshot',
-          maxTimeMS: lib_mongo.db.MONGO_OPERATION_TIMEOUT_MS
+    const pipeline: mongo.Document[] = [
+      {
+        $match: {
+          'key.g': ctx.group_id,
+          lookup: { $in: lookupFilter },
+          _id: { $lte: checkpoint.checkpoint }
         }
-      )
+      },
+      {
+        $sort: {
+          _id: -1
+        }
+      },
+      {
+        $group: {
+          _id: { key: '$key', lookup: '$lookup' },
+          bucket_parameters: {
+            $first: '$bucket_parameters'
+          }
+        }
+      }
+    ];
+    if (limit != null) {
+      pipeline.push({ $limit: limit + 1 });
+    }
+    const rows = await ctx.db.parameterIndexV1
+      .aggregate(pipeline, {
+        session,
+        readConcern: 'snapshot',
+        maxTimeMS: lib_mongo.db.MONGO_OPERATION_TIMEOUT_MS
+      })
       .toArray()
       .catch((e) => {
         throw lib_mongo.mapQueryError(e, 'while evaluating parameter queries');
       });
+    if (limit != null && rows.length > limit) {
+      throw new ParameterSetLimitExceededError(limit);
+    }
+
     const groupedParameters = rows.map((row) => {
       return row.bucket_parameters;
     });
--- a/modules/module-mongodb-storage/src/storage/implementation/v1/MongoSyncBucketStorageV1.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/v1/MongoSyncBucketStorageV1.ts
@@ -98,7 +98,7 @@ export class MongoSyncBucketStorageV1 extends MongoSyncBucketStorage {
   protected getParameterSetsImpl(
     checkpoint: MongoSyncBucketStorageCheckpoint,
     lookups: ScopedParameterLookup[],
-    limit: number | undefined
+    limit: number
   ): Promise<SqliteJsonRow[]> {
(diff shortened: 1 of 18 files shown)


modules/module-mongodb-storage/src/storage/implementation/v3/MongoSyncBucketStorageV3.ts
```diff
@@ -235,6 +239,9 @@ export async function getParameterSetsV3(
               _id: -1
             }
           },
+          // This limit still allows returning too many rows because this filter might be put into a $unionWith, but
+          // at least the amount of rows we'd return isn't unbounded.
+          { $limit: limit + 1 },
```
rkistner: The limit should be applied _after_ the `$group` stage, not before.

A high number of results before the $group stage indicates a need for running a parameter compact job, but it shouldn't count towards the limit. And having the limit applied there could change the output without triggering the `ParameterSetLimitExceededError` below.

Then there is a mismatch in the limit applied in the `$limit` stage and here - the check is applied _after_ a `flatMap` call:
1. The check matches the number parameters seen in the final results - that's good.
2. But in theory there could be many cases of empty `row.bucket_parameters`. In that case the `$limit` could be applied in the pipeline, without triggering the check, also silently altering the output.

An additional `$match: { bucket_parameters: { $ne: [] }  }` stage before the `$limit` could avoid that issue.

We could also do an `$unwind` before applying the limit, but the effect there shouldn't be as drastic.

modules/module-mongodb-storage/src/storage/implementation/v1/MongoSyncBucketStorageV1.ts
```diff
@@ -217,6 +220,9 @@ export async function getParameterSetsV1(
               _id: -1
             }
           },
+          {
+            $limit: limit + 1
+          },
```
rkistner: Same as mentioned for V3 storage - the limit should be applied after the `$group`, not before, and also need to filter out empty `bucket_parameters` arrays before the limit.

modules/module-postgres-storage/src/storage/PostgresSyncRulesStorage.ts
```diff
@@ -424,9 +426,14 @@ export class PostgresSyncRulesStorage
...
         source_key,
         id DESC
+      LIMIT
+        ${{ type: 'int4', value: limit + 1 }}
     `
       .decoded(pick(models.BucketParameters, ['bucket_parameters']))
       .rows();
+    if (rows.length > limit) {
+      throw new ParameterSetLimitExceededError(limit);
+    }
```
rkistner: See related the comments on the MongoDB side. Here the check does match the limit in the query, so it is "safe" in the sense that the limit can't silently alter results. But it could apply the limit alittle too drastically for cases with many empty `bucket_parameters` arrays.

Note that the cast `as sync_rules.SqliteJsonRow` below is wrong - it should be `as sync_rules.SqliteJsonRow[]`.

packages/service-core/src/sync/BucketChecksumState.ts
```diff
@@ -730,6 +794,33 @@ function formatParameterQueryBreakdown(parameterQueryResultsByDefinition: Map<st
...
+  let message = '\nInvoked parameter queries by definition:';
+  results.sort((a, b) => b.resultsOrLimit - a.resultsOrLimit);
+  const largestResults = results.splice(0, 9);
+
+  for (const entry of largestResults) {
+    message += `\n  ${entry.definition}: ${entry.resultsOrLimit} results.`;
+  }
+
+  if (largestResults.length < log.length) {
+    message += `\n ... and ${largestResults.length - log.length} more invocations`;
```
rkistner: This will log a negative number

packages/service-core/src/sync/BucketChecksumState.ts
```diff
@@ -499,20 +503,60 @@ export class BucketParameterState {
...
 
   async getCheckpointUpdate(checkpoint: storage.StorageCheckpointUpdate): Promise<CheckpointUpdate> {
+    // We have three limits to check:
+    //
+    //   1. Parameter query results: To avoid overloading the service, all Sync Streams / bucket definitions for the
+    //      current connection have a maximum amount of parameters they're allowed to fetch from bucket storage.
+    //      An `ParameterSetLimitExceededError` error is thrown when this limit is exceeded.
+    //   2. If we were able to fetch all parameter results, we apply the same limit to the amount of buckets returned.
+    //      This is mostly relevant for Sync Streams, where the amount of parameter queries is not strictly related to
+    //      the amount of buckets returned.
```
rkistner: Technically it's a different limit - `maxBuckets` versus `maxParameterQueryResults` - even though the default is the same for both (1000).

There are cases even in sync rules where there could be many duplicates. A bad example:

```sql
parameters: select sender_id as user_id from chats
```

Since `sender_id` is not an unique field, it would return many duplicate values, even though it only results in a small number of buckets. In this case it's a bad query to use as a parameter query, but there could be other cases where a small number of duplicates is expected. We can consider increasing the default `maxParameterQueryResults` in the future for that reason.

simolus3: Isn't the `maxBuckets` limit described by step 3 in this comment?

Limit check 1 has been added in this PR, `maxParameterQueryResults` has previously been implemented by checking for the amount of buckets directly returned by the dynamic querier (which the comment describes as step 2).

The point I was trying to make is that the amount of parameter results is not always the amount of buckets (even before deduplicatation) anymore, because you can combine multiple parameters. E.g. if you had a regular subquery with `N` results and a `json_each` on request data expanding to `M` rows, we'd now consider `N` for step 1 and `N * M` for step 2 (previously the only `maxParameterQueryResults` check).

rkistner: I think I'm seeing what you're saying now. The part that's confusing to me as written is this:
> we apply the same limit to the amount of buckets returned.

I interpreted "limit to the amount of buckets returned" as the total number of buckets returned to the client (step 3). In step 2 it is still using `maxParameterQueryResults`, and it's not the raw parameter query results anymore, but it's also not quite "the amount of buckets returned" either.

I think your error message "Too many buckets derived from parameter queries" is a good description here.

modules/module-mongodb-storage/src/storage/implementation/v3/MongoSyncBucketStorageV3.ts
simolus3: I like the `$unwind` idea, it sounds like the most accurate way of expressing in the query pipeline what we're actually interested in (avoiding the flatmap in JS).

For my understanding, when would there be multiple entries in `bucket_parameters` for the same lookup key? The only scenario I could some up with is a `json_each` in a parameter subquery, e.g. `SELECT * FROM tasks WHERE id IN (SELECT json_each.value FROM task_assignment a, json_each(a.tasks) WHERE a.user = auth.user_id())`?

rkistner: I had to look into the code, and actually couldn't find _any_ case that produces more than 1 entry in `bucket_parameters`. Turns out your example above generates separate entries, but with the same lookup (expected) and key (an issue). Those then get de-duplicated, which discards all results except one, which is a bug.

Added failing test case:
https://github.com/powersync-ja/powersync-service/compare/json-each-parameter-index-bug

The same issue does not appear to affect legacy sync rules, since it's the combination of json_each + parameter indexes that triggers this, which is not a combination that legacy sync rules supports.

To fix this we'd need to actually use a single record with multiple `bucket_parameters` entries, ~or use unique keys for each~. _Update: can't use unique keys for each record, since the key is tied to the replicated row_

simolus3: Thanks, I can take a look at that based from your test (storing a single records with multiple entries in its array sounds like what we were supposed to do here?).

Either way, I think the queries in this PR should be correct now.

rkistner (approved):

PR #624, opened 2026-05-06
Rework `resolveTable(s)`

rkistner:
This is in preparation for incremental reprocessing.

This builds on #619. While the two PRs are mostly focusing on different aspects, there is some overlap that would cause conflicts if I split them into independent PRs.

### Multiple tables

Previously, `resolveTable` would always return a single `SourceTable`. This changes it to `resolveTables` that may return more than one.

The specific use case here is for adding new sync streams (new BucketDataSource and/or ParameterIndexLookupCreator). If we have other sources referencing the same source table, we already have a `SourceTable` entry with a completed snapshot. But for the new source, we need a new snapshot. To avoid that new snapshot affecting the existing sources, we create a new `SourceTable` for that. Only storage v3 supports this - v1 storage does not support adding or removing sync streams on existing replication streams.

For reference, everything under `source_records`, `bucket_data` and `parameter_index` collections are scoped to specific `SourceTable`s - either directly or indirectly.

On each `SourceTable` we specifically keep track of the individual data sources and parameter index definitions. As mentioned above, we never add new definitions to an existing table, but we do remove definitions from existing `SourceTables`.

More details on the storage structure are available in `docs/storage-v3.md`.

We now also supporting returning zero `SourceTables` for tables not referenced in sync queries. This is roughly equivalent to the case of `table.syncAny == false`, but avoids keeping track of the table at all.

### Source module updates

Each source replication implementation is updated to handle the multiple source tables. For the most part it just affects the caching logic, and routes writes separately for each `SourceTable`.

### Filtered definitions per table

Each `SourceTable` now has this:

```
bucketDataSources: BucketDataSource[];
parameterLookupSources: ParameterIndexLookupCreator[];
```

For storage V3, those are persisted with the tables. This ensures specific sources are only used with the specific `SourceTable`s they were assigned to.

For storage V1, each `SourceTable` applies to all matching definitions. This change still computes the specific list of sources for the table upfront (when `resolveTables` is called), which can speed up replication when replicating dozens or hundreds of tables.

### SourceTableInterface restructuring

This introduces a clean split between `TablePattern`, `SourceTableRef` (new term for `SourceTableInterface`) and `SourceTable`.

Since we no longer have a 1:1 mapping between the underlying source table (as encapsulated by `SourceTableRef`) and the persisted `SourceTable`, using those interchangably in calls introduced some issues. These are now split, with `SourceTable` no longer matching the `SourceTableRef` interface. Instead, we can use `SourceTable.ref` when we need it.

More details on this split are documented in `docs/resolve-tables-flow.md`.

### Other restructuring

1. We now avoid creating an entire SourceTable instance where a SourceTableRef is sufficient, such as in the API adapters.
2. In tests we avoid mocking SourceTable, instead using the actual `resolveTables` call. Only a small number of tests were still using the old `makeTestTable` function.
3. This fixes some cases where `ColumnDescriptor` was constructed incorrectly - `typeId` vs `typeOid`, and bigint instead of number. This did not appear to cause any issues before, but it started causing issues after some of these refactorings.

### AI Usage

This PR was AI-assisted: A large portion of the code was written using Codex, but manually directed, and manually reviewed and tested.

### DIFF
--- a/.changeset/little-years-flow.md
+++ b/.changeset/little-years-flow.md
@@ -1,0 +1,13 @@
+---
+'@powersync/service-module-postgres-storage': minor
+'@powersync/service-module-mongodb-storage': minor
+'@powersync/service-core-tests': minor
+'@powersync/service-module-postgres': minor
+'@powersync/service-module-mongodb': minor
+'@powersync/service-core': minor
+'@powersync/service-module-mssql': minor
+'@powersync/service-module-mysql': minor
+'@powersync/service-sync-rules': minor
+---
+
+[Internal] rework resolveTables to handle multiple SourceTables.
--- a/docs/resolve-tables-flow.md
+++ b/docs/resolve-tables-flow.md
@@ -19,9 +19,15 @@ A `TablePattern` is a sync-rule selector:
 
 Think of it as: "which tables should we care about?"
 
+### SourceTableRef
+
+A `SourceTableRef` describes the table of a source row as replicated.
+
+We can do directy matching of `TablePattern.matches(SourceTableRef)`.
+
 ### SourceTable
 
-A `SourceTable` is a concrete tracked runtime table:
+A `SourceTable` is a concrete tracked runtime table with state:
 
 - It represents one resolved physical source table identity.
 - It carries replication identity metadata (replica-id columns).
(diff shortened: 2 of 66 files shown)


stevensJourney (dismissed):

stevensJourney (approved):
This looks like a nice addition. The added docs were very nice to read and helped parse the changes :)

PR #625, opened 2026-05-06
Fix replicating parameter rows with multiple outputs

simolus3:
Typically, parameter indexes generate at most one set of bucket parameters per output row:

```sql
-- Prepares a parameter index from projects.owner -> projects.id.
-- A row with a null value in owner or id generates zero lookups, otherwise it generates one lookup result.
SELECT tasks.* FROM tasks INNER JOIN projects on tasks.project = projects.id AND projects.owner = auth.user_id()
```

Thanks to `IN` and table-valued functions, it's also possible to have parameter indexes generate multiple lookup results per row:

```sql
-- Assuming projects.id is non-null, this generates one lookup for each value in projects.owners (interpreted as a JSON array)
SELECT tasks.* FROM tasks INNER JOIN projects on tasks.project = projects.id AND auth.user_id() IN projects.owners
```

Both of these cases work correctly. However, there's a third possible topology that's currently broken: A parameter row can generate a lookup with multiple output values:

```sql
-- Assuming a non-null owner value, generates a single lookup (with the owner) with multiple output parameters (one per task id).
SELECT tasks.* FROM tasks INNER JOIN projects WHERE tasks.id IN projects.task_ids AND projects.owner = auth.user_id()
```

(multiple `IN` and `json_each` calls can also be combined to generate multiple lookups with multiple output parameters each)

The third case currently generates multiple lookups with the same key, which bucket storage implementations interpret as multiple revisions from different database states. Only the last lookup would be used in this case, which is incorret! This fixes evaluating parameters to generate a single lookup with multiple outputs in that case.

### DIFF
--- a/packages/sync-rules/src/sync_plan/evaluator/parameter_index_lookup_creator.ts
+++ b/packages/sync-rules/src/sync_plan/evaluator/parameter_index_lookup_creator.ts
@@ -3,7 +3,7 @@ import { ParameterIndexLookupCreator } from '../../BucketSource.js';
 import { ParameterLookupScope } from '../../HydrationState.js';
 import { SourceTableInterface } from '../../SourceTableInterface.js';
 import { TablePattern } from '../../TablePattern.js';
-import { SqliteJsonValue, SqliteRow, UnscopedEvaluatedParametersResult } from '../../types.js';
+import { SqliteJsonValue, SqliteParameterValue, SqliteRow, UnscopedEvaluatedParametersResult } from '../../types.js';
 import { ScalarExpressionEvaluator } from '../engine/scalar_expression_engine.js';
 import * as plan from '../plan.js';
 import { StreamEvaluationContext } from './index.js';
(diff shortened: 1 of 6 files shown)


packages/sync-rules/src/sync_plan/evaluator/parameter_index_lookup_creator.ts
```diff
@@ -53,9 +53,44 @@ export class PreparedParameterIndexLookupCreator implements ParameterIndexLookup
...
+        for (const existing of lookups) {
+          if (parametersEqual(existing.partitionValues, entries)) {
+            return existing;
+          }
+        }
+
+        const created: PendingLookup = { partitionValues: entries, bucketParameters: [] };
+        lookups.push(created);
+        return created;
+      }
```
rkistner, line 92: In theory this is `O(n^2)`, which can become slow for a large number of lookups.

In practice, I don't see how a single row would produce a very large number of lookups, especially since unbounded values from json_each / IN operators would not generate additional lookups anymore, so I don't think this is an issue.

rkistner (approved):

PR #626, opened 2026-05-07
Improve consistency in logs, error messages and comments

rkistner:
### Log update

This uses a consistent prefix using the replication stream identifier in all replication logs.

Specific changes:
1. Add the prefix to checksum pre-calculation and in some replication job lifecycle logs.
2. Avoid custom identifiers like `Sync rules 4` in logs.
3. Always use the full "slot name" in the identifier, matching what is available in the diagnostics API.

Some examples before:
```
Sync rules from configuration unchanged
Sync rules 4 has been locked for replication with lock ID 5be58fc72dedaa84.
[powersync_4] Initial replication already done
...
[powersync_5] Snapshot done. Need to replicate from  ...
Populating persistent checksum cache...
Calculating checksums for batch of 1 buckets, estimated count of 16
Updating 1 bucket states
info: Updated checksums for batch of 1 buckets in 2ms
info: Populated persistent checksum cache in 0.0s
...
Stopping replication job for sync rule iteration: 4
```

After:

```
Sync config unchanged
[powersync_507f191e810c19729de860ea_4_8077] Locked replication stream for processing
[powersync_507f191e810c19729de860ea_4_8077] Initial replication already done
...
[powersync_507f191e810c19729de860ea_5_32e2] Snapshot done. Need to replicate from ...
[powersync_507f191e810c19729de860ea_5_32e2] Populating persistent checksum cache...
[powersync_507f191e810c19729de860ea_5_32e2] Calculating checksums for batch of 1 buckets, estimated count of 16
[powersync_507f191e810c19729de860ea_5_32e2] Updating 1 bucket states
[powersync_507f191e810c19729de860ea_5_32e2] Updated checksums for batch of 1 buckets in 3ms
[powersync_507f191e810c19729de860ea_5_32e2] Populated persistent checksum cache in 0.0s
...
[powersync_507f191e810c19729de860ea_4_8077] Stopping replication job
```

### Broad terminology update

1. "sync rules" in the context of replication state becomes "replication stream".
    a. "sync rules version" or "sync rules id" becomes "replication stream id".
2. "sync rules" in the context of the config / queries becomes "sync config".

This is updated in error messages, logs and comments.

External APIs, TypeScript APIs and variables names are not touched in this refactor.

## AI usage

This PR was implemented manually, then checked with Codex (gpt-5.5 medium) for issues.


README.md
```diff
@@ -32,7 +32,7 @@ The service can be started using the public Docker image. See the image [notes](
 
 - [packages/sync-rules](./packages/sync-rules/README.md)
 
-  - Library containing logic for PowerSync sync rules
+  - Library containing logic for PowerSync sync config (sync streams or sync rules)
```
cahofmeyr:
```suggestion
  - Library containing logic for PowerSync sync config (Sync Streams or Sync Rules)
```

packages/service-errors/src/codes.ts
```diff
@@ -5,10 +5,10 @@
  * for each.
  */
 export enum ErrorCode {
-  // # PSYNC_Rxxxx: Sync rules issues
+  // # PSYNC_Rxxxx: Sync config issues
```
stevensJourney, line 8: Looks like we might need to update the docs version of this also https://docs.powersync.com/debugging/error-codes#psync_rxxxx-sync-rules-issues

stevensJourney (approved):
The new output looks much better!

PR #629, opened 2026-05-11
Upgrade to pnpm 11

simolus3:
This upgrades to pnpm version 11 (see [release notes](https://pnpm.io/blog/releases/11.0)) and a [migration guide](https://pnpm.io/migration) for details.

A breaking change that affects us here is that `pnpm install` seems to error if we have a dependency with an install script for which we haven't made an explicit decision. This affects `protobufjs` here, I've disabled that package's build script.

An interesting change is that `pnpm publish` is now implemented in pnpm instead of using `npm publish` as a subcommand. I've triggered a dev release to make sure things are still working.

### DIFF
--- a/.github/workflows/packages_release.yaml
+++ b/.github/workflows/packages_release.yaml
@@ -87,7 +87,7 @@ jobs:
           # If no changesets are available the status check will fail
           # We should not continue if there are no changesets    
           pnpm changeset status
-          pnpm changeset version --no-git-tag --snapshot dev
+          pnpm changeset version --snapshot dev
           pnpm changeset publish --tag dev
 
   release-docker-image:
--- a/package.json
+++ b/package.json
@@ -50,5 +50,5 @@
     "vitest": "catalog:",
     "ws": "^8.2.3"
   },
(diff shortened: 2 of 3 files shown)


.github/workflows/packages_release.yaml
```diff
@@ -87,7 +87,7 @@ jobs:
           # If no changesets are available the status check will fail
           # We should not continue if there are no changesets    
           pnpm changeset status
-          pnpm changeset version --no-git-tag --snapshot dev
+          pnpm changeset version --snapshot dev
```
simolus3, line 90: It looks like the `--no-git-tag` option has been removed in a recent changesets version, so we're affected by this since https://github.com/powersync-ja/powersync-service/pull/623.

`changeset version` doesn't create git tags (`changeset publish` does). According to [this](https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#publish), we're supposed to run `git push --follow-tags` after publishing, implying that changeset doesn't do that by default? We don't seem to do that though, and the documentation has recommended that for a while. So I'm not sure if we need to fix the publishing workflow to push tags at this point.

rkistner (approved):

.github/workflows/packages_release.yaml
simolus3: I guess we'll see on the next release :D

PR #630, opened 2026-05-11
Include requested lookup for parameter results

simolus3:
When querying for parameters, we provide a `ScopedParameterLookup[]` array and currently get a `SqliteJsonRow[]` of results in response. The response is a union of parameter rows found for all lookups in the input array, we don't know which lookup resuled in which rows.

Returning all results in a single array blocks two improvements:

1. An implementation of Sync Streams might want to pass multiple `ScopedParameterLookup` values for different parameter indexes into a single `getParameterSets` call for efficiency. This is not currently possible since we wouldn't be able to inspect which output was generated by which parameter.
2. In complex cases where a parameter has multiple outputs which are passed into a different parameter, we need to track how rows from the first parameter influence the second to implement `JOIN` semantics correctly.

(neither of those two things is currently implemented, but since this is a non-trivial change on its own it helps to have it in a separate PR).

## AI usage disclaimer

Most test changes in this PR were migrated using Claude after I've updated a few tests by hand.

### DIFF
--- a/.changeset/gold-hats-sit.md
+++ b/.changeset/gold-hats-sit.md
@@ -1,0 +1,8 @@
+---
+'@powersync/service-module-postgres-storage': minor
+'@powersync/service-module-mongodb-storage': minor
+'@powersync/service-sync-rules': minor
+'@powersync/service-core': patch
+---
+
+When querying parameter rows, track which lookups resulted in which rows.
--- a/modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
@@ -22,7 +22,7 @@ import {
   utils,
   WatchWriteCheckpointOptions
 } from '@powersync/service-core';
-import { HydratedSyncRules, ScopedParameterLookup, SqliteJsonRow } from '@powersync/service-sync-rules';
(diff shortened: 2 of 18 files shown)


rkistner (commented):
The pipeline change to `$unionWith` in `MongoSyncBucketStorageV1` is a potentially risky change in terms of performance and limits - will have to test that part. I know V3 is using essentially the same type of pipeline, but that hasn't gone through the same amount of testing yet.

For example, MongoDB has a [limit of 1000 stages](https://www.mongodb.com/docs/manual/core/aggregation-pipeline-limits/#number-of-stages-restrictions) in a pipeline. It's not clear how that applies to `$unionWith` - whether that applies to the number of top-level stages, or the total number of nested stages. Either way, we'll need to test and potentially split into separate operations if the number of lookups grow too big.

simolus3:
> The pipeline change to `$unionWith` in `MongoSyncBucketStorageV1` is a potentially risky change in terms of performance and limits - will have to test that part

In practice the array would be very small (probably a single element in almost all cases), but nested parameter lookups with Sync Streams can make it go wrong. It probably makes sense to implement a check for that (if we request more than 1000 lookups at once, the only way for that to succeed is if most lookups didn't return a result set. So we might as well forbid it outright).

Anyway, the safe approach is probably to start the pipeline like this:

1. Keep the existing `$match` on `group_id`, `$in: lookupFilter` and `< checkpoint.checkpoint`.
2. Then afterwards, use `$indexOfArray` and pay the cost of a linear search since the array is likely to be small.

rkistner (approved):

PR #633, opened 2026-05-12
Correlate parameters if they originate from the same logical result set

simolus3:
Sorry for the noise, I closed https://github.com/powersync-ja/powersync-service/pull/632 by mistake with a careless rebase.

Starting from https://github.com/powersync-ja/powersync-service/pull/608 (which hasn't been released yet), it's possible for Sync Streams to use compound keys as parameters, e.g. in:

```sql
SELECT products.* FROM products, stores
  WHERE stores.name = products.store_name
    AND stores.region = products.region
    AND stores.id = subscription.parameter('store');
```

The way this is replicated is fairly obvious: When we encounter a `stores` row, we populate a parameter index from `id` to `name` and `region`. When building a querier, we use these two outputs as parameters. Unfortunately, that is not quite correct because we then build the cartesian product of all parameters as buckets. So, if we had two rows in `stores` with `{name: 'foo', region: 'us'}` and `{name: 'bar', region: 'eu'}`, we'd generate four buckets: `[foo,us], [foo, eu], [bar, us], [bar, eu]`. Oops.

To implement join semantics correctly, we also need to track which logical row in a joined result set a parameter value was derived from. This is implemented here by tracking:

1. The result set, as an opaque unique symbol instance.
2. The row in a result set, as an incrementing number.

When we build the cartesian product for parameter instantiations, we keep track of which result sets have already appeared in partial instantiations. When we discover a new value from a covered result set with a different row, we skip it.

This fixes an issue introduced in https://github.com/powersync-ja/powersync-service/pull/608 (which hasn't been released yet, hence no changeset entry).

## AI usage disclaimer

Ralf originally found this issue with Codex security scanning. This builds ontop of 37f42ea2125ee0591fcba2534912d461f778e731, which contains a fix and tests generated by Codex. The implementation was rewritten manually for clarity and to fix this across multiple lookup stages too.

### DIFF
--- a/packages/service-core-tests/src/tests/register-data-storage-parameter-tests.ts
+++ b/packages/service-core-tests/src/tests/register-data-storage-parameter-tests.ts
@@ -1067,4 +1067,87 @@ streams:
     expect(foundLookupA).toBeTruthy();
     expect(foundLookupB).toBeTruthy();
   });
+
+  test('sync streams preserve duplicate downstream lookups with different provenance', async () => {
+    await using factory = await generateStorageFactory();
+    const syncRules = await factory.updateSyncRules(
+      updateSyncRulesFromYaml(
+        `
+config:
+  edition: 3
+streams:
+  stream:
+    auto_subscribe: true
+    query: |
+      SELECT a.*
+      FROM a, b, c
+      WHERE a.x = b.x
+        AND a.z = c.z
+        AND b.y = c.y
+        AND c.u = auth.user_id()
+    `,
+        {
+          storageVersion
+        }
+      )
+    );
+    const bucketStorage = factory.getInstance(syncRules);
+    const sync_rules = syncRules.parsed(test_utils.PARSE_OPTIONS).hydratedSyncRules();
+
+    await using writer = await bucketStorage.createWriter(test_utils.BATCH_OPTIONS);
+    const tableB = await test_utils.resolveTestTable(writer, 'b', ['id'], config);
+    const tableC = await test_utils.resolveTestTable(writer, 'c', ['id'], config);
+
+    await writer.markAllSnapshotDone('1/1');
+    await writer.save({
+      sourceTable: tableB,
+      tag: storage.SaveOperationTag.INSERT,
+      after: {
+        id: 'b1',
+        y: 'shared-y',
+        x: 'x-from-shared-y'
+      },
+      afterReplicaId: test_utils.rid('b1')
+    });
+    await writer.save({
+      sourceTable: tableC,
+      tag: storage.SaveOperationTag.INSERT,
+      after: {
+        id: 'c1',
+        u: 'user1',
+        y: 'shared-y',
+        z: 'z1'
+      },
+      afterReplicaId: test_utils.rid('c1')
+    });
+    await writer.save({
+      sourceTable: tableC,
+      tag: storage.SaveOperationTag.INSERT,
+      after: {
+        id: 'c2',
+        u: 'user1',
+        y: 'shared-y',
+        z: 'z2'
+      },
+      afterReplicaId: test_utils.rid('c2')
+    });
+    await writer.commit('1/1');
+
+    const checkpoint = await bucketStorage.getCheckpoint();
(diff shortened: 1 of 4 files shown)


rkistner (commented):
Codex picked up a further issue, relating to a mismatch between how ScopedParameterLookup references are expected to be handled in the evaluator vs storage layer. The current implementation works with MongoDB Storage V3, but has a case that fails on MongoDB V1 and Postgres storage. This specifically happens when the same lookup values is used with multiple different provenances. MongoDB V3 storage preserves the original ScopedParameterLookup references, which avoids the issue. But MongoDB V1 and Postgres storage de-duplicate the queries by lookup _value_, which can then lose the provenance associations.

I don't think MongoDB V1 and Postgres storage are wrong here - it makes sense to de-duplicate these lookups by value, and we should perhaps implement the same for MongoDB V3 storage. But I think we need to change how the evaluator does the association.

Basically, instead of `Map<ScopedParameterLookup, provenance>` where we rely on reference equality for ScopedParameterLookup, use something like `Map<lookup value, provenance[]>`.

Changing that the storage implementation _could_ also work, but then we need to document the expected behavior clearly.

This is an "end to end" test demonstrating the issue, in `register-data-storage-parameter-tests.ts`:

```ts
  test('sync streams preserve duplicate downstream lookups with different provenance', async () => {
    await using factory = await generateStorageFactory();
    const syncRules = await factory.updateSyncRules(
      updateSyncRulesFromYaml(
        `
config:
  edition: 3
streams:
  stream:
    auto_subscribe: true
    query: |
      SELECT a.*
      FROM a, b, c
      WHERE a.x = b.x
        AND a.z = c.z
        AND b.y = c.y
        AND c.u = auth.user_id()
    `,
        {
          storageVersion
        }
      )
    );
    const bucketStorage = factory.getInstance(syncRules);
    const sync_rules = syncRules.parsed(test_utils.PARSE_OPTIONS).hydratedSyncRules();

    await using writer = await bucketStorage.createWriter(test_utils.BATCH_OPTIONS);
    const tableB = await test_utils.resolveTestTable(writer, 'b', ['id'], config);
    const tableC = await test_utils.resolveTestTable(writer, 'c', ['id'], config);

    await writer.markAllSnapshotDone('1/1');
    await writer.save({
      sourceTable: tableB,
      tag: storage.SaveOperationTag.INSERT,
      after: {
        id: 'b1',
        y: 'shared-y',
        x: 'x-from-shared-y'
      },
      afterReplicaId: test_utils.rid('b1')
    });
    await writer.save({
      sourceTable: tableC,
      tag: storage.SaveOperationTag.INSERT,
      after: {
        id: 'c1',
        u: 'user1',
        y: 'shared-y',
        z: 'z1'
      },
      afterReplicaId: test_utils.rid('c1')
    });
    await writer.save({
      sourceTable: tableC,
      tag: storage.SaveOperationTag.INSERT,
      after: {
        id: 'c2',
        u: 'user1',
        y: 'shared-y',
        z: 'z2'
      },
      afterReplicaId: test_utils.rid('c2')
    });
    await writer.commit('1/1');

    const checkpoint = await bucketStorage.getCheckpoint();
    const parameters = new RequestParameters(new JwtPayload({ sub: 'user1' }), {});
    const querier = sync_rules.getBucketParameterQuerier({
      ...test_utils.querierOptions(parameters)
    }).querier;

    const buckets = await querier.queryDynamicBucketDescriptions({
      async getParameterSets(lookups) {
        return checkpoint.getParameterSets(lookups, 1000);
      }
    });

    expect(buckets.map((bucket) => bucket.bucket).sort()).toStrictEqual([
      expect.stringMatching(/stream.*\["x-from-shared-y","z1"\]$/),
      expect.stringMatching(/stream.*\["x-from-shared-y","z2"\]$/)
    ]);
  });
```

In MongoDB V1 storage, the second bucket is missing.

simolus3:
> Basically, instead of `Map<ScopedParameterLookup, provenance>` where we rely on reference equality for ScopedParameterLookup, use something like `Map<lookup value, provenance[]>`.

Done 👍

rkistner (approved):

PR #637, opened 2026-05-14
Fix NOT IN json type mismatch

joshuabrink:
This pull request addresses issues with evaluating `IN` and `NOT IN` operators against scalar JSON arrays in SQLite, ensuring proper type affinity for booleans and large integers. The changes improve how values are compared, especially for booleans and integers larger than 2^53, and adds tests to prevent regressions.

### AI Usage
A large portion of the tests were written using Claude, but the results were manually verified.

### DIFF
--- a/packages/sync-rules/src/sql_functions.ts
+++ b/packages/sync-rules/src/sql_functions.ts
@@ -652,7 +652,8 @@ export function evaluateOperator(op: string, a: SqliteValue, b: SqliteValue): Sq
         return null;
       }
       const bParsed = checkJsonArray(b, 'IN is only supported on JSON arrays');
-      return sqliteBool(bParsed.includes(a));
+      // Use compare() for SQLite affinity since Array.includes strict equality mismatches bigint vs number.
+      return sqliteBool(bParsed.some((v) => compare(v, a) === 0));
     }
     case '&&': {
       // a && b evaluates to true iff they're both arrays and have a non-empty intersection.
@@ -675,17 +676,18 @@ export function evaluateOperator(op: string, a: SqliteValue, b: SqliteValue): Sq
   }
 }
 
-export function checkJsonArray(value: SqliteValue, errorMessage: string): any[] {
+export function checkJsonArray(value: SqliteValue, errorMessage: string): SqliteValue[] {
   if (typeof value != 'string') {
     throw new Error(errorMessage);
   }
 
-  const parsed = JSON.parse(value);
+  // JSONBig preserves precision for integers above 2^53
+  const parsed = JSONBig.parse(value);
   if (!Array.isArray(parsed)) {
     throw new Error(value);
   }
 
-  return parsed;
+  return parsed.map((v: any) => jsonValueToSqlite(true, v));
 }
 
 export function getOperatorReturnType(op: string, left: ExpressionType, right: ExpressionType) {
--- a/packages/sync-rules/src/sql_functions.ts
+++ b/packages/sync-rules/src/sql_functions.ts
@@ -676,7 +676,7 @@ export function evaluateOperator(op: string, a: SqliteValue, b: SqliteValue): Sq
   }
 }
 
-export function checkJsonArray(value: SqliteValue, errorMessage: string): SqliteValue[] {
+export function checkJsonArray(value: SqliteValue, errorMessage: string): any[] {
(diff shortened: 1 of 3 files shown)


packages/sync-rules/src/sql_functions.ts
```diff
@@ -680,12 +681,13 @@ export function checkJsonArray(value: SqliteValue, errorMessage: string): any[]
...
 
-  const parsed = JSON.parse(value);
+  // JSONBig preserves precision for integers above 2^53
+  const parsed = JSONBig.parse(value);
   if (!Array.isArray(parsed)) {
     throw new Error(value);
   }
 
-  return parsed;
+  return parsed.map((v: any) => jsonValueToSqlite(true, v));
```
rkistner: Not a regression in this PR, but I saw you experimented with this: For better typing, use `SqliteJsonValue[]` for the return type. That's like `SqliteValue`, but excludes `Uint8Array`, which will resolve the type issue you had.

packages/sync-rules/test/src/sync_plan/evaluator/not_in_type_affinity.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  stream:
+    auto_subscribe: true
+    query: SELECT * FROM docs WHERE flag IN '[true]'
+`);
+
+    expect(desc.evaluateRow({ sourceTable: DOCS, record: { id: 'a', flag: 1n } })).toHaveLength(1);
+    expect(desc.evaluateRow({ sourceTable: DOCS, record: { id: 'b', flag: 0n } })).toHaveLength(0);
+  });
+
+  syncTest('large bigint column NOT IN keeps precision via JSONBig parse', ({ sync }) => {
```
simolus3, line 38: The inverse of this test (`WHERE big IN '[9999999999999999]'` with inverted expectations) fails.

The reason is that positive `IN` filters are compiled to `SELECT * FROM docs WHERE big IN (SELECT value FROM json_each('[9999999999999999]'))`. These get evaluated by the `json_each` implementation in `TableValuedFunctions.ts`, which doesn't use JSONBig (it should!, see the comment in line 27).

IMO, we should try to fix both cases here. The main concern is that we can't change how existing deployments of Sync Streams behave, and these changes slightly alter how `IN` operators work. Maybe this change is so minor that it's acceptable to alter replication (I'll defer to @rkistner to make that call), if not we'd have to add another `CompatibilityOption` for this.

rkistner: From current usage on the cloud version:
1. `NOT IN` usage is _very_ rare, and not a compatibility risk.
2. `IN` filters are quite common, for example `IN auth.parameter('project_ids')`.

If I'm only thinking about `IN` filters, I can't imagine any case where the fixed behavior would be worse than the current behavior.

But I'm less sure about generally changing the behavior of `json_each`. I think we should implement a fix, but the impact might be significant enough that we need a new compatibility option.

If we don't change that right now, that might mean a small difference between how `IN` and `NOT IN` works. @simolus3 Do you think that would be an issue?

simolus3: > Do you think that would be an issue?

This PR fixes another difference between `IN` and `NOT IN`, so it's clearly an improvement already and we were fine with this difference before. So I agree that leaving things as-as and looking at `json_each` fixes with another compatibility option later is fine.

Maybe it makes sense to add that as a skipped test for now.

packages/sync-rules/src/sql_functions.ts
joshuabrink: Ah makes sense, thank you.

packages/sync-rules/test/src/sync_plan/evaluator/not_in_type_affinity.test.ts
joshuabrink: Fair enough leaving it as-is then, thanks for the eyes. Makes sense to keep this PR more scoped given the impact.

I've added a skipped test covering `WHERE big IN '[9999999999999999]'` with a TODO pointing at `TableValuedFunctions.ts` for when follow up lands.

rkistner (approved):

simolus3 (approved):

PR #642, opened 2026-05-19
Normalize host strings before IP validation

joshuabrink:
Normalizes the host string inside `validateIpHostname` so that direct-IP literals coming from URL parsers are recognized as IPs (instead of falling through to the DNS path).

Two upstream parsers produce host strings that didn't match the previous `ip.isValid(hostname)` check:
- `URL.hostname` wraps IPv6 literals in brackets `new URL('https://[::1]/').hostname === '[::1]'`.
- `mongodb-connection-string-url` returns `uri.hosts` entries as `host:port`, `[::1]`, or `[::1]:port`.

Adds a small `hostWithoutPort` helper that strips a `:port` suffix and unwraps bracketed IPv6, called from `validateIpHostname` before `ip.isValid`.

### AI Usage
Tests and codebase exploration done with Claude. Verified issues and fix correctness.

### DIFF
--- a/.changeset/fix-host-normalization.md
+++ b/.changeset/fix-host-normalization.md
@@ -1,0 +1,5 @@
+---
+'@powersync/lib-services-framework': patch
+---
+
+Normalize host strings in `validateIpHostname` before IP validation. Adds `hostWithoutPort`, which strips port suffixes and unwraps bracketed IPv6 literals, so values produced by `URL.hostname` and the `uri.hosts` entries returned by `mongodb-connection-string-url` are recognized as IPs.
--- a/.changeset/fix-host-normalization.md
+++ b/.changeset/fix-host-normalization.md
@@ -1,5 +1,7 @@
 ---
 '@powersync/lib-services-framework': patch
+'@powersync/lib-service-mongodb': patch
+'@powersync/service-core': patch
 ---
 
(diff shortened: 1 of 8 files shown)


rkistner (commented):
There are some more edge cases this doesn't cover, for example `mongodb://127.0.0.1:/powersync_test`, and there may be others.

I think we need two changes to make this robust:

Change 1: In each function, make it clear where we're dealing with:
1. A socket address, typically from a URI, which can include examples such as `[::1]:1234`, but not bare IPv6 addresses are not allowed. This is what we'd get from `new URL().host` , `new URL().hostname` (excludes port), and `mongodb-connection-string-url` `hosts`.
2. A normalized hostname - no ports, no wrapping in brackets. This is what `urijs.parse().host` gives us.

Never mix the two - be explicit about which form is required as a function argument.

Change 2: Avoid custom parsing. Instead, use `net.SocketAddress.parse()` to go from socket address to a normalized hostname.

As part of this we'd need to audit every place that a "hostname" is set. For example, we have a couple of patterns like `const hostname = options.hostname ?? uri.host`. That's fine if `uri` is from `uri-js` (both are normalized hostnames), but not if it's from `URL` or `mongodb-connection-string-url`.

rkistner (approved):

PR #644, opened 2026-05-21
Fix Sync Streams iif arity validation

sravan27:
## Summary
- validate Sync Streams `iif()` calls with the same three-argument arity used by the evaluator
- add compiler regressions for two-arg and four-arg `iif()` calls
- add a patch changeset for `@powersync/service-sync-rules`

## Why
`iif` is implemented and documented as `iif(x, y, z)`, but the compiler previously accepted any call with at least two arguments. That allowed invalid sync definitions such as `iif(1, 2)` or `iif(1, 2, 3, 4)` through validation even though the runtime implementation only consumes three arguments.

## Verification
- `corepack pnpm --filter @powersync/service-sync-rules exec vitest run test/src/compiler/sqlite.test.ts`
- `corepack pnpm --filter @powersync/service-sync-rules build:tsc`

### DIFF
--- a/.changeset/fix-sync-streams-iif-arity.md
+++ b/.changeset/fix-sync-streams-iif-arity.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Validate `iif()` calls with the same three-argument arity used by the sync stream evaluator.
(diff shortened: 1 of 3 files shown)


rkistner (approved):
This is technically a breaking change - any case currently using `iif` with a different number of arguments would stop working. But those cases also mean a bug in the query, so I'm happy with accepting that risk.

simolus3 (approved):

PR #645, opened 2026-05-21
Fix Sync Streams signed string casts

sravan27:
## Summary

Fix Sync Streams expression evaluation for SQLite-style casts from signed numeric strings. The JavaScript evaluator previously parsed only unsigned numeric prefixes, so values like `CAST('-12' AS integer)` and `CAST('-.5' AS numeric)` evaluated as `0` instead of matching SQLite.

## Why this matters

Sync rules can cast request parameters or JSON/text values before filtering rows. A signed value in a parameter or source row could therefore select the wrong rows in the JavaScript evaluator compared with SQLite semantics.

## Changes

- Allow optional `+` / `-` prefixes when parsing integer casts from strings.
- Allow signed decimal and exponent forms for numeric casts, including `.5` / `-.5`.
- Add direct cast tests and scalar expression engine coverage.
- Add a changeset for `@powersync/service-sync-rules`.

## Verification

- `corepack pnpm --dir packages/sync-rules exec vitest test/src/sql_functions.test.ts --run`
- `corepack pnpm --dir packages/sync-rules exec vitest test/src/sync_plan/engine/engine.test.ts --run -t "javascript.*signed string casts|javascript"`
- `corepack pnpm --dir packages/sync-rules run build:tsc`

Note: running the full scalar engine test file against local Node v23.7.0 fails in the existing SQLite-engine tests because Node's experimental `node:sqlite` returns object rows instead of the array rows expected by the suite. The JavaScript evaluator path and pure cast tests pass locally.

### DIFF
--- a/.changeset/fix-sync-streams-signed-cast.md
+++ b/.changeset/fix-sync-streams-signed-cast.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Fix SQLite-compatible casts from signed numeric strings in Sync Streams expressions.
--- a/packages/sync-rules/src/cast.ts
+++ b/packages/sync-rules/src/cast.ts
@@ -101,23 +101,25 @@ export function cast(value: SqliteValue, to: string) {
 }
 
 function parseNumeric(text: string): bigint | number {
-  const match = /^\s*(\d+)(\.\d*)?(e[+\-]?\d+)?/i.exec(text);
+  const match = /^\s*([+\-]?)(?:(\d+)(\.\d*)?|\.(\d+))(e[+\-]?\d+)?/i.exec(text);
   if (!match) {
     return 0n;
   }
 
-  if (match[2] != null || match[3] != null) {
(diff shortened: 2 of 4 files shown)


rkistner:
There is a test failure: `packages/sync-rules/test/src/sync_plan/engine/engine.test.ts:125` gives `120` for the JS engine (real sqlite type), while the sqlite engine gives `120n` (integer sqlite type). I think this difference is fine - we don't need to match sqlite's behavior 100%. But we do need to fix the test to handle both cases.

Note that the Node version we use is defined in `.nvmrc` - currently v24.15.0. You should be able to run the tests locally using that.

sravan27:
Pushed the test expectation tweak from review: the JS engine keeps `120`, while the SQLite engine accepts `120n` under Node 24.15.0.\n\nFocused validation under Node 24.15.0:\n- `engine.test.ts`: 37/37 passed\n- `sql_functions.test.ts`: 25/25 passed

rkistner (approved):

PR #646, opened 2026-05-21
Fix Sync Streams division by zero semantics

sravan27:
## Summary

This fixes a Sync Streams JavaScript evaluator divergence from SQLite around division by zero.

SQLite returns `NULL` for divide-by-zero expressions:

- `5 / 0`
- `5 / 0.0`
- `5 / '0'`
- `0 / 0`

The current JavaScript evaluator can instead return `Infinity` for real division by zero or throw for bigint division by zero. In filter expressions this can create evaluator errors or non-SQLite truthiness instead of SQLite's `NULL` result.

## Changes

- Return `null` from the shared math operator implementation when the division RHS casts to zero.
- Add regression coverage for bigint, number, string-cast, negative, and zero-over-zero cases.
- Add a patch changeset for `@powersync/service-sync-rules`.

## Verification

```text
sqlite3 :memory: "select typeof(5/0), 5/0, typeof(5/0.0), 5/0.0, typeof(5/'0'), 5/'0';"
# null||null||null|

corepack pnpm --dir packages/sync-rules exec vitest test/src/sql_operators.test.ts --run
# 15 passed

corepack pnpm --dir packages/sync-rules exec vitest test/src/sql_functions.test.ts --run --testTimeout=10000
# 25 passed

corepack pnpm --dir packages/sync-rules run build:tsc
# passed

git diff --check
# passed
```

### DIFF
--- a/.changeset/quiet-lemons-divide.md
+++ b/.changeset/quiet-lemons-divide.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Fix Sync Streams division by zero semantics
--- a/packages/sync-rules/src/sql_functions.ts
+++ b/packages/sync-rules/src/sql_functions.ts
@@ -744,6 +744,10 @@ function doMath(op: string, a: SqliteValue, b: SqliteValue) {
   let na = cast(a, 'numeric') as number | bigint;
   let nb = cast(b, 'numeric') as number | bigint;
 
+  if (op == '/' && nb == 0) {
+    return null;
+  }
+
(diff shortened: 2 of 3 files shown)


rkistner (dismissed):

rkistner:
Minor thing: Could you please run prettier on the changed files? (You can run `pnpm format` to format all files)

sravan27:
Done, ran Prettier on the changed changeset file and force-pushed the formatted commit. The previous logic/test changes are otherwise unchanged.

simolus3 (approved):

PR #647, opened 2026-05-21
Fix Sync Streams json_each object and scalar evaluation

sravan27:
This fixes a Sync Streams JavaScript evaluator divergence from SQLite around `json_each()`.

SQLite's `json_each()` emits rows for JSON arrays, JSON objects, and scalar JSON values:

```sql
SELECT key, value FROM json_each('{"a":1,"b":true}');
SELECT key, value FROM json_each('42');
```

The current JavaScript table-valued-function evaluator only accepts arrays, throws for objects, and cannot expose the `key` column. That means valid Sync Streams rules using object-shaped request/JWT/connection JSON can fail or diverge when evaluated by the JavaScript engine.

This PR:

- Keeps existing array behavior.
- Adds SQLite-compatible rows for JSON objects using object keys.
- Adds SQLite-compatible scalar handling with `key = null`.
- Exposes both `key` and `value` columns from the JavaScript evaluator.
- Adds focused engine coverage for object and scalar `json_each()` evaluation.
- Adds a patch changeset for `@powersync/service-sync-rules`.

Verification:

```bash
corepack pnpm --filter @powersync/service-sync-rules exec vitest --run test/src/sync_plan/engine/engine.test.ts -t "javascript.*json_each"
corepack pnpm --filter @powersync/service-sync-rules run build:tsc
git diff --check
```

Note: the local `node:sqlite` side of this engine test still hits the same local Node binding issue seen on earlier Sync Streams PR verification (`column index out of range` while binding table-valued JSON inputs), so I scoped the runnable local test command to the JavaScript evaluator path this patch changes.

### DIFF
--- a/packages/sync-rules/src/TableValuedFunctions.ts
+++ b/packages/sync-rules/src/TableValuedFunctions.ts
@@ -1,5 +1,5 @@
 import { CompatibilityContext, CompatibilityOption } from './compatibility.js';
-import { SqliteJsonValue, SqliteRow, SqliteValue } from './types.js';
+import { SqliteRow, SqliteValue } from './types.js';
 import { jsonValueToSqlite } from './utils.js';
 
 export interface TableValuedFunction {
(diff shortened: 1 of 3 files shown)


sravan27:
Fixed the lint blocker from CI: the changeset file had a Prettier quote-style mismatch. I formatted it and force-pushed the branch; implementation/test changes are otherwise unchanged.

packages/sync-rules/src/TableValuedFunctions.ts
```diff
@@ -22,25 +22,32 @@ function jsonEachImplementation(fixedJsonBehavior: boolean): TableValuedFunction
...
+        value = JSON.parse(valueString);
       } catch (e) {
         throw new Error('Expected JSON string');
       }
-      if (!Array.isArray(values)) {
-        throw new Error(`Expected an array, got ${valueString}`);
+
+      if (Array.isArray(value)) {
+        return value.map((v, key) => ({
+          key,
```
rkistner:
```suggestion
          key: BigInt(key),
```

For our JS <-> SQLite type mapping, we treat JS `number` as SQLite `real` values, and JS `bigint` as SQLite `integer` values. So casting to BigInt here is required to avoid treating these indices as `real`.

rkistner (approved):

Issue #321, opened 2025-08-09
Token generator doesn't understand `!env` vars


KazimirPodolski:
1. Have some env vars in your `powersync.conf`:
    ```yaml
    ...
    client_auth:
      jwks:
        keys:
          - alg: HS256
            kty: oct
            k: !env PS_JWT_SECRET
    ...
    ```
1. Run token generator for this config:
    ```
    (node:1755082) [TAG_RESOLVE_FAILED] YAMLWarning: Unresolved tag: !env at line 59, column 12:
    
            k: !env PS_JWT_SECRET
               ^^^^
    ```

The token **was** generated tho, but I have a suspicion it's not correct...

joshua-journey-apps:
I made a [fork](https://github.com/joshua-journey-apps/powersync-service) that supports loading env vars directly when generating a token.
You can try it like this:

```shell
node dist/bin.js generate-token \
  --config ~/path/to/powersync.yaml \
  --sub test-user \
  --env ~/path/to/.env
```
This will populate all the !env values in your config before token generation.

PR #650, opened 2026-05-25
Resolve !env tags in test-client

joshuabrink:
This addresses #321 

The `test-client` can now resolve `!env` tags in `powersync.yaml` configuration files by using the service’s existing `YamlEnvTag` custom tag.

Environment variables are read from exported shell variables by default, or they can be loaded explicitly from a file using the new `--env` flag.

### DIFF
--- a/.changeset/test-client-env-tag.md
+++ b/.changeset/test-client-env-tag.md
@@ -1,0 +1,6 @@
+---
+'@powersync/service-core': patch
+'test-client': patch
+---
+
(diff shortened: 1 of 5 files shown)


stevensJourney (approved):
The changes here look good to me.

I get the feeling it could be nice to move the functionality of this `test-client` to the PowerSync CLI at some point. The current DX of having to clone and build this repo to use this tool is not the greatest. FWIW, the PowerSync CLI already supports token generation given a self hosted configuration (the commands and structure are different though).

PR #651, opened 2026-05-26
[Incremental Reprocessing] Multiplexed sync configs

rkistner:
This refactors `HydratedSyncRules` to support multiple `SyncConfig` instances. When multiple are passed in, we de-duplicate definitions. This currently requires the storage layer to assign unique ids/keys to each definition, which `HydratedSyncRules` then uses for de-duplication. That process will be implemented in a future PR - nothing uses multiple sync configs yet.

## Why?

This will allow a replication stream/job to continue using `HydratedSyncConfig` as-is to process multiple `SyncConfig`s at a time, provided that the storage implementation constructs it correctly.

## Further refactoring included

 * Rename `HydratedSyncRules` -> `HydratedSyncConfig`.
 * Include `key` in `BucketDataScope` and `ParameterLookupScope`.
 * Split out `ParameterLookupDefinitionId` as a subset of `ParameterLookupScope`, excluding `source` and the new `key`, to avoid the storage details leaking into the sync config parsing.

## AI usage

Used Codex gpt-5.5 for planning and implementation, manually guided and reviewed.


rkistner:
@codex review

packages/sync-rules/src/HydrationState.ts
```diff
@@ -70,14 +100,23 @@ export function versionedHydrationState(version: number): HydrationState {
...
         source
       };
     },
 
     getParameterIndexLookupScope(source: ParameterIndexLookupCreator): ParameterLookupScope {
       // No transformations applied here
-      return source.defaultLookupScope;
+      return {
+        key: parameterLookupScopeKey(source.sourceId),
+        ...source.sourceId,
```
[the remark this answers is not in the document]
rkistner: This `versionedHydrationState` is and will never be used for incremental reprocessing, so this is not relevant.

packages/sync-rules/src/HydratedSyncConfig.ts
```diff
@@ -21,80 +22,107 @@ import {
...
+      (source) => hydrationState.getBucketSourceScope(source).key
+    );
+    this.bucketParameterLookupSources = uniqueBy(
+      definitions.flatMap((definition) => definition.bucketParameterLookupSources),
+      (source) => hydrationState.getParameterIndexLookupScope(source).key
+    );
+    this.bucketSourceDefinitions = uniqueBy(
+      definitions.flatMap((definition) => definition.bucketSources),
+      (source) => source.name
+    );
```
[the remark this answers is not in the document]
rkistner: This will be covered in a future PR.

packages/sync-rules/src/HydratedSyncConfig.ts
```diff
@@ -21,80 +22,107 @@ import {
...
+      this.bucketParameterLookupSources
     ).evaluateParameterRow;
 
-    if (params.eventDescriptors) {
-      this.eventDescriptors = params.eventDescriptors;
-    }
-    if (params.compatibility) {
-      this.compatibility = params.compatibility;
-    }
+    this.eventDescriptors = definitions.flatMap((definition) => definition.eventDescriptors);
```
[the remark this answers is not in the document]
rkistner: This will be covered in a future PR.

packages/service-core/src/storage/PersistedSyncRulesContent.ts
```diff
@@ -170,5 +170,5 @@ export interface PersistedSyncRules {
    */
   readonly hydrationState: HydrationState;
 
-  hydratedSyncRules(): HydratedSyncRules;
+  hydratedSyncRules(): HydratedSyncConfig;
```
simolus3: Should we also rename this to `hydratedSyncConfig` while we're at it?

packages/sync-rules/src/HydratedSyncConfig.ts
```diff
@@ -21,80 +22,107 @@ import {
...
+ * The persisted state specifically affects bucket names, as well as V3+ storage structure.
  */
-export class HydratedSyncRules {
+export class HydratedSyncConfig {
   bucketSources: HydratedBucketSource[] = [];
   eventDescriptors: SqlEventDescriptor[] = [];
   compatibility: CompatibilityContext = CompatibilityContext.FULL_BACKWARDS_COMPATIBILITY;
 
-  readonly definition: SyncConfig;
+  readonly definitions: SyncConfig[];
```
simolus3: This is more of a question, but I wonder if we should make this private or give it another name that makes it clear that we might not use everything from all definitions?

I tried removing it by rewriting e.g `tableSyncsData` with `return this.bucketDataSources.some((s) => s.tableSyncsData(table))`, but there's no obvious solution for `tableTriggersEvent`.

packages/sync-rules/src/HydrationState.ts
```diff
@@ -1,13 +1,33 @@
...
+
 export interface BucketDataScope {
+  /**
+   * Stable key identifying the bucket data source.
+   *
+   * Generated by the storage implementation, and may be persisted.
+   *
+   * Must be unique per replication stream. May have duplicates over different streams, depending on storage requirements.
+   */
+  key: BucketDefinitionId;
```
simolus3: For my understanding, two bucket data scopes would have the same key if and only if they have the same bucket prefix too, right (because it's the same physical bucket even if defined in different sync config deployments)?

Would if make sense to treat `bucketPrefix` as a unique key then instead of adding another one?

packages/sync-rules/src/HydrationState.ts
```diff
@@ -1,13 +1,33 @@
...
 
 export interface ParameterLookupScope {
+  /**
+   * Stable key identifying the parameter index.
+   *
+   * Generated by the storage implementation, and may be persisted.
+   *
+   * Must be unique per replication stream. May have duplicates over different streams, depending on storage requirements.
+   */
+  key: ParameterIndexId;
```
simolus3: Same here, is `lookupName + queryId` another unique key?

simolus3 (commented):
These changes make sense to me after our offline discussion 👍 I have a few minor questions.

packages/sync-rules/src/HydratedSyncConfig.ts
rkistner: I'm making it private so that it's at least only used internally. Can eventually see if we can remove it completely.

packages/sync-rules/src/HydrationState.ts
rkistner: You might be right - we could probably use that as the unique key - I'll see if it works in practice.

It will still be different from the storage key, but that shouldn't matter where this is used.

packages/sync-rules/src/HydratedSyncConfig.ts
```diff
@@ -21,80 +22,113 @@ import {
...
+      throw new Error('HydratedSyncRules requires at least one SyncConfig definition');
+    }
+
+    this.sourceDefinitions = [...definitions];
+    this.compatibility = assertSharedCompatibility(this.sourceDefinitions);
+
+    this.bucketDataSources = uniqueBy(
+      definitions.flatMap((definition) => definition.bucketDataSources),
+      (source) => hydrationState.getBucketSourceScope(source).key
+    );
```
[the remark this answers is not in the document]
rkistner: This will be handled in #653.

packages/sync-rules/src/HydrationState.ts
rkistner: I removed the `key` fields now and added comments on `bucketPrefix`, clarifying that it's safe to use for this use case. Same for `ParameterLookupScope` below.

simolus3 (approved):

PR #653, opened 2026-05-27
[WIP] Support multiple SyncConfigs per replication stream

rkistner:
This wires up incremental reprocessing, by using multiple SyncConfigs per replication stream.

## Refactoring

The old `PersistedSyncRules` + `PersistedSyncRulesContent` is split into:
1. `PersistedReplicationStream`: This represents the overall replication stream - mostly for job management.
2. `PersistedSyncConfigContent` - the content for individual sync configs inside a replication stream. This is mostly an immutable representation of the config.
3. `PersistedSyncConfigStatus` - this contains the status such as whether the snapshot is complete.
4. `ParsedSyncConfigSet`: This represents a parsed version of one or more SyncConfigs in the same replication stream. Does not have to cover all SyncConfigs in the stream, depending on usage.
 
This also refactors some fields:
1. `PersistedSyncRulesContent.id` or `group_id` is renamed to `replicationStreamId`, to distinguish from `syncConfigId`.
2. `slot_name` is renamed to `replicationStreamName`, to indicate the more general use case outside of Postgres.

## Storage V3 implementation changes

1. Deploying new sync config may now add the config to an existing replication stream, if compatible (uses sync streams, same storage version, same compatibility config).
2. Updating SourceTable status (snapshot done/required) now updates the status of specific sync configs referencing it.

_TODO: More details here_


packages/sync-rules/src/sync_plan/plan_equality.ts
```diff
@@ -0,0 +1,246 @@
...
+  TableProcessorTableValuedFunction,
+  TableProcessorTableValuedFunctionOutput
+} from './plan.js';
+
+export function streamBucketDataSourcesEqual(a: StreamBucketDataSource, b: StreamBucketDataSource): boolean {
+  if (a === b) {
+    return true;
+  }
+
+  return a.hashCode == b.hashCode && arrayEquals(a.sources, b.sources, streamDataSourcesEqual);
```
simolus3: Is not including `uniqueName` here deliberate? It's implicitly included in `hashCode`.

rkistner: I'm still busy investigating whether it's feasible to allow renames without reprocessing, but I feel it could work conceptually.

Where is `uniqueName` currently used in `hashCode`, or are you referring to my stubs in this PR?

simolus3: > I'm still busy investigating whether it's feasible to allow renames without reprocessing, but I feel it could work conceptually.

I agree that would be nice 👍 

> Where is `uniqueName` currently used in `hashCode`, or are you referring to my stubs in this PR?

I assumed that `uniqueName` would influence `StreamBucketDataSource.hashCode` (in which case this function would implicitly compare `uniqueName` as well), but that's [actually not true](https://github.com/powersync-ja/powersync-service/blob/5ac534505b161966a8bcbb1257098ffaf871fc24/packages/sync-rules/src/compiler/ir_to_sync_plan.ts#L69). So this seems correct.

rkistner: Despite the technical feasibility, the other side of this is the `bucketPrefix`: Right now we have the stream name in the `bucketPrefix` (via `uniqueName`), which helps with debugging. But if we allow renaming the stream name, that becomes harder. We could:
1. Use a generated name independent from the stream name, but that makes debugging more difficult. We could adapt tooling to report the stream name itself instead of just the bucket name, but it affects many places including diagnostics app, service logs, ps_buckets on the client, etc.
2. Use the original `uniqueName`, and leave it unchanged when renaming a stream. This could be confusing, potentially leading users to believe the update did not apply.
3. Update the name to use the new stream name / `uniqueName`. We could avoid reprocessing for that, but users would still need to re-download the data, losing a lot of the gains from this.

So given all of that, I might just go with requiring the `uniqueName` to stay the same, at least for the initial version.

rkistner:
Split into #671 + #670 to assist with reviewing.

PR #654, opened 2026-05-27
Support SQLite as a scalar expression engine

simolus3:
Apart from a few PowerSync-specific additions, the SQL dialect implemented in Sync Streams is supposed to be SQLite. Currently, we implement operators and functions in JavaScript, and the idea is that we're supposed to match semantics from SQLite. Recently, @sravan27 found a few cases where we're not quite getting this right:

1. `json_each` should work on objects and scalar values, we only support arrays (https://github.com/powersync-ja/powersync-service/pull/647).
2. SQLite returns `NULL` for a division by zero, our implementation throws or returns Infinity (https://github.com/powersync-ja/powersync-service/pull/646).
3. We don't handle `+` / `-` prefixes on numbers when casting from text (https://github.com/powersync-ja/powersync-service/pull/645).
4. There are a few cases where we don't handle `NULL` values correctly (https://github.com/powersync-ja/powersync-service/pull/643).
5. Noteworthy mention: `iif` was broken in the compiler but only there (https://github.com/powersync-ja/powersync-service/pull/644). This PR would not have fixed that one.

Fixing these is possible, but a breaking change in most cases (requiring a compatibility option). And even with options, we can't know how many remaining differences there are (those would require additional options).

As a solution, this adds the `unstable_sqlite_expression_engine` compatibility option. When enabled, functions and operators are evaluated by a scalar `SELECT` statement (without a `FROM` clause) on an in-memory SQLite database. This means we'll get the exact behavior from SQLite going forward.

This also includes a small refactoring on internal sync config representations to not require an SQL engine upfront. Instead, we:

1. First parse YAML (or restore a serialized) plan into an in-memory structure, this should not require access to a SQLite database.
6. To then evaluate rows, parameters or queriers, we hydrate the config. This step requires a SQL engine now.

## Known issues

`node:sqlite` is not yet stable, so we can't build stable options with it. SQLite is also known to change how it evaluates expressions from time to time (the text -> float conversion update in 3.52 is a good example for this).

I think this is good enough to ship it as an experimental option. To stabilize it, we should perhaps look into adopting a WebAssembly build of SQLite. That would give us a fixed SQLite version and allows us to install a custom VFS without time support (so we wouldn't have to override builtin date and time functions).

## Product visibility

We'll have to update the docs page on available options to mention this.

### DIFF
--- a/packages/sync-rules/test/src/sync_plan/evaluator/evaluator.test.ts
+++ b/packages/sync-rules/test/src/sync_plan/evaluator/evaluator.test.ts
@@ -1223,6 +1223,25 @@ streams:
   expect(dynamicBuckets.map((b) => b.bucket)).toStrictEqual(['stream|0["issue"]']);
 });
 
+syncTest('throws when using SQLite without providing it', ({ sync }) => {
+  const desc = sync.prepareWithoutHydration(`
+config:
+  edition: 3
+  unstable_sqlite_expression_engine: true
+
+streams:
+  stream:
+      auto_subscribe: true
+      query: SELECT comments.* FROM comments, issues
+        WHERE issues.id = comments.issue
+        AND issues.owner = auth.user_id()
+`);
+
+  expect(() => desc.hydrate({ hydrationState: DEFAULT_HYDRATION_STATE, sqlite: null })).toThrow(
+    'sqlite_expression_engine is unsupported on this target.'
+  );
+});
+
 function evaluateBucketIds(source: HydratedSyncConfig, sourceTable: SourceTableRef, record: SqliteRow) {
   return source.evaluateRow({ sourceTable, record }).map((r) => r.bucket);
 }
--- a/packages/sync-rules/test/src/sync_plan/evaluator/evaluator.test.ts
+++ b/packages/sync-rules/test/src/sync_plan/evaluator/evaluator.test.ts
@@ -1,7 +1,9 @@
+import * as sqlite from 'node:sqlite';
 import { describe, expect } from 'vitest';
 import {
   DEFAULT_HYDRATION_STATE,
   HydratedSyncConfig,
+  nodeSqlite,
   ParameterLookupRows,
   PrecompiledSyncConfig,
   ScopedParameterLookup,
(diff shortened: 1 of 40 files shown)


packages/sync-rules/test/src/sync_plan/evaluator/evaluator.test.ts
```diff
@@ -1170,6 +1172,57 @@ streams:
   });
 });
 
+syncTest('SQLite smoke test', async ({ sync }) => {
```
rkistner, line 1175: `syncTest` currently has a hardcoded `new CompatibilityContext({ edition: 3 })`, so does not actually test the sqlite engine here.

rkistner (commented):
I'm happy in principle with the new experimental option. Sorry for all the merge conflicts...

In terms of compatibility changes with future SQLite versions: I'm generally happy to accept the compatibility guarantees from SQLite itself. They do change behavior in some cases, but it's fairly rare. That said, it would still be good to have some control over _when_ we upgrade SQLite, rather than getting updates automatically with new NodeJS versions.

Codex picked up two issues - one in the SQLite engine, one in the tests. While neither is a regression in this PR specifically, they are surfaced with this PR. I manually verified these.

### Issue 1:

`scalarStatementToSql` does not preserve order of operations in all cases. Example (add to `packages/sync-rules/test/src/sync_plan/engine/engine.test.ts`):

```ts
  test('preserves right-nested binary expression precedence', () => {
    const stmt = prepare({
      outputs: [
        {
          type: 'binary',
          left: { type: 'data', source: 1 },
          operator: '-',
          right: {
            type: 'binary',
            left: { type: 'data', source: 2 },
            operator: '-',
            right: { type: 'data', source: 3 }
          }
        }
      ]
    });

    expect(stmt.evaluate([10n, 3n, 1n])).toStrictEqual([[8n]]);
  });
```

The above test passes with the JS engine, but fails with the SQLite engine.

### Issue 2

Issue with tests not actually testing SQLite - see inline comment.

simolus3:
I've fixed the merge conflicts and precedence issue 👍

rkistner (approved):

PR #459, opened 2026-01-13
[Postgres] Avoid storing copy of replicated rows with REPLICA IDENTITY FULL

hamzzy:
## Summary
Implements table-level `storeCurrentData` configuration for PostgreSQL REPLICA IDENTITY FULL optimization, eliminating the need to store raw row data when complete rows are always available through replication.

## Problem
Previously, PowerSync used a global `storeCurrentData` setting that applied uniformly across all tables in a database. PostgreSQL tables configured with `REPLICA IDENTITY FULL` always send complete row data in replication events, making the `current_data` collection redundant for these tables and wasting storage.

## Solution
Made `storeCurrentData` a table-level property that's automatically determined during table resolution:

- **PostgreSQL**: Detects `REPLICA IDENTITY FULL` and sets `storeCurrentData = false`
- **Other sources**: Defaults to `storeCurrentData = true` (existing behavior)

### Core API Changes
- Extended `SourceEntityDescriptor` with optional `replicationIdentity` field
- Added `storeCurrentData` property to `SourceTable` class
- Property is preserved through table cloning

### PostgreSQL Module
- `WalStream` now extracts and passes `replicationIdentity` from replication info
- Properly propagates this to `resolveTable`

### PostgreSQL Storage
- `PostgresBucketBatch` now checks `sourceTable.storeCurrentData` per-record instead of using global setting
- Filters operations to only query `current_data` for tables that need it

### MongoDB Storage
- Refactored to compute `storeCurrentData` based on `replicationIdentity`
- Stores setting in `source_tables` collection for persistence
- Updates setting when table configuration changes

### Testing
- **267 test cases** for MongoDB storage scenarios
- **367 test cases** for PostgreSQL REPLICA IDENTITY configurations  
- **120 unit tests** for SourceTable behavior
- Covers: FULL, DEFAULT, INDEX, and NOTHING replica identity types

## Performance Impact
For PostgreSQL tables with `REPLICA IDENTITY FULL`:
- **~50% storage reduction** (no `current_data` entries)
- **25-30% throughput increase** (fewer database operations)
- **Zero configuration** required (automatic detection)

## Backward Compatibility
✅ Fully backward compatible:
- Defaults to `storeCurrentData = true` for all non-FULL tables
- No migration required
- Safe gradual rollout for existing deployments
- Old deployments continue to work without changes

## Database Support
- **PostgreSQL**: ✅ Automatic optimization
- **MySQL**: No change (always stores current_data)
- **MSSQL**: No change (always stores current_data)  
[...]

PR #655, opened 2026-05-29
 [Postgres] Skip redundant current_data copies for REPLICA IDENTITY FULL

### LINKED ISSUE #398: [Postgres] Avoid storing copy of replicated rows with REPLICA IDENTITY FULL
When a table is configured with REPLICA IDENTITY FULL, a copy of the entire row is sent with every operation, avoiding any issues around TOAST values. This means we don't need to persist a copy of the raw data in the _current_data_ collection - we can remove it for these tables.

Currently the logic for whether or not to store this is global based on the source database type - see logic here for example:
https://github.com/powersync-ja/powersync-service/blob/b457d92e491b11e1bed20b2fb801994b15b7f97d/modules/module-mongodb-storage/src/storage/implementation/MongoBucketBatch.ts#L200

We should make this more granular to support table-level config, to reduce storage requirements and potentially increase throughput.

joshuabrink:
Re-implements #459 (which resolves #398)

Tables with `REPLICA IDENTITY FULL` always send the complete row on every change, making PowerSync's per-row `current_data` copy redundant. This turns `store_current_data` into a per-table decision, skipping the copy (and the TOAST-merge read) for FULL tables while keeping it for every other replica identity.

### What changed

The source reports a source-agnostic `sendsCompleteRows` on the entity descriptor (true only for FULL in Postgres), from which service-core derives the per-table flag: complete rows → no copy, otherwise → keep a copy. The flag is **derived in memory on every resolve and never persisted**, and a source that doesn't report completeness defaults to keeping a copy, so no row data is ever dropped.

Because it's re-derived on every `pgoutput` Relation message, a mid-stream `ALTER TABLE ... REPLICA IDENTITY` is picked up immediately. The same ALTER also changes the reported identity columns (DEFAULT → PK, FULL → all columns), so the table re-resolves to a new `SourceTable` and re-snapshots transparently.

### Compatibility

Backward compatible, no migration: with nothing persisted, existing FULL tables pick up the optimization on their next resolve. Identity columns are unchanged across an upgrade, so no re-snapshot is required on upgrade, that only happens when the replica identity itself changes.

### Testing

E2E tests in `module-postgres` cover both ALTER directions (including TOAST recovery); `module-mongodb-storage` covers the storage-layer flag for v1 and v3.

### AI Usage

This PR was AI-assisted (Claude Code), manually directed, reviewed, and tested.


rkistner (commented):
Before I did too far into the details - is there a way to do this that would avoid persisting the `store_current_data` flag on the `SourceTable`?

If that's not feasible, we should look into a more generic way of storing db-specific metadata on the `SourceTable` - I know we have another similar requirement for persisting MSSQL change capture instances - cc @Rentacookie.

joshuabrink:
In-memory works and is quite a bit cleaner.

It seemed like we would need to persist a signal for the unreported completeness case (to maintain a table's `store_current_data` state despite no signal from the catalog). But on Postgres replica identity is always reported, on startup and on every resolve. And when the source doesn't report, we default to copying, so no row data is dropped.

modules/module-postgres/test/src/replica_identity_full.test.ts
```diff
@@ -0,0 +1,347 @@
...
+    const [{ id: id1 }] = pgwireRows(
+      await pool.query(`INSERT INTO test_mid_stream (description, value) VALUES ('before', 1) RETURNING id`)
+    );
+    let data = await context.getBucketData('global[]');
+    expect(data).toMatchObject([putOp('test_mid_stream', { id: id1, description: 'before', value: 1n })]);
+
+    // pgoutput sends the new Relation message lazily, on the next DML rather than at ALTER time.
+    await pool.query(`ALTER TABLE test_mid_stream REPLICA IDENTITY FULL`);
+
+    // The next DML carries it; handleRelation must re-derive storeCurrentData without re-snapshotting.
```
rkistner: The "without re-snapshotting" part is not quite true here. Even though the `storeCurrentData` change can be handled without re-snapshotting, the replica identity change also causes `replicaIdColumns` to change, which does trigger a re-snapshot.

This is not a regression or issue, but the test comments should be fixed.

This also affects the comments below - `// snapshotComplete stays true - no re-snapshot` is not accurate. This does trigger a re-snapshot, but it's handled transparently so not visible to the test as implemented here. You should see that `initial.id` is not the same as `after.id`.

modules/module-postgres/test/src/replica_identity_full.test.ts
```diff
@@ -0,0 +1,347 @@
...
+    // Update only `name`: `description`'s TOAST value is omitted and DEFAULT carries only the key in
+    // the old tuple, so with no stored copy the row is incomplete -> a targeted resnapshot repairs it.
+    await pool.query(`UPDATE test_revert SET name = 'test2' WHERE id = '${id}'`);
+
+    data = await context.getBucketData('global[]');
+    expect(data.findLast((op) => op.object_id === id)).toMatchObject(
+      putOp('test_revert', { id, name: 'test2', description: largeDescription })
+    );
+
+    // Back to storeCurrentData=true, without a re-snapshot of the whole table.
```
rkistner: Same here - it does trigger a re-snapshot.

modules/module-postgres/test/src/replica_identity_full.test.ts
```diff
@@ -0,0 +1,347 @@
...
+    );
+    let data = await context.getBucketData('global[]');
+    expect(data.at(-1)).toMatchObject(putOp('test_revert', { id, name: 'test1', description: largeDescription }));
+
+    // pgoutput sends the new Relation message lazily, on the next DML rather than at ALTER time.
+    await pool.query(`ALTER TABLE test_revert REPLICA IDENTITY DEFAULT`);
+
+    // Update only `name`: `description`'s TOAST value is omitted and DEFAULT carries only the key in
+    // the old tuple, so with no stored copy the row is incomplete -> a targeted resnapshot repairs it.
+    await pool.query(`UPDATE test_revert SET name = 'test2' WHERE id = '${id}'`);
```
rkistner: I don't think a targeted resnapshot is performed here:
1. As explained above, the replica identity change triggers a full resnapshot.
2. Even if it didn't do a full resnapshot, we'd still have the full description since the last update was performed with `REPLICA IDENTITY FULL`, in which case no targeted resnapshot would be needed.

rkistner (commented):
This is great, and appears simpler now that the value is not persisted anymore.

I just highlighted some inaccurate comments in tests. The PR description also has some outdated parts and inaccuracies:
1. `resolveStoreCurrentData` and the persisted value is not present anymore.
2. Same with `the column defaults to NULL`
3. Same as in the tests, `picked up without a re-snapshot` is not quite true.

"No re-snapshot required on upgrade" is true, but probably trickier to test.

You could technically test a case where REPLICA IDENTITY changing with `replicaIdColumns` changing (e.g. by having the primary key include all columns), but I'm not sure if that's valuable enough.

joshuabrink:
Totally makes sense, thanks for the pointers. I should of given the tests another look through. But okay, updated the comments, and added some source table id checks for prosperity.

rkistner (approved):

PR #659, opened 2026-06-03
[MongoDB] Fix checkpoint stalling

rkistner:
This fixes an issue surfaced in #573 (v1.20.3). Even though the issue was not directly _caused_ by that PR, it removed a fallback path that would have bypassed the issue.

What happens:
1. In the replication process, `createCheckpoint()` is called.
2. A connection error causes the write to fail after the server received it.
3. The driver retries the command automatically.
4. The server sees the retry as a no-op since it already applied the original one, and does not apply it again.
5. The client session's operationTime still increases after the retry.
6. Now, the operationTime reports the time of the second operation, while the change stream event contains the clusterTime of the first operation.
7. This causes the `lsn >= waitForCheckpointLsn` condition to not be hit when the change stream event is received.
8. Since #573, nothing else creates an update that passes the filter. The replication process indefinitely waits for a change stream event that it never receives.
9. This causes the replication process to never create a new sync checkpoint.

## Issue symptoms

The issue shows up as replication lag building up, without an actual backlog of data to process. Specifically:

1. An initial slow batch with long `processing` time.
2. After that, replication lag builds up indefinitely.
3. No specific large data volumes that cause replication lag. For example, there may be many small batches, where the majority of time is just spent waiting for new changes.
4. A restart resolves the issue, without processing a backlog.

## The fix

A good long term fix would be to use a sentinel-based approach, rather than relying on MongoDB clusterTime behavior and potentially making bad assumptions around that - see #605. However, that's a big and risky change.

The fix here is to bypass the internal driver's retry mechanism for writes. By adding our own retry logic, the retry will actually produce a new write with a new change stream event, which avoids this issue.

## AI usage

Used Codex gpt-5.5 to debug the issue and assist in writing a test for reproducing it.

The fix here was implemented manually.

### DIFF
--- a/modules/module-mongodb/src/replication/ChangeStream.ts
+++ b/modules/module-mongodb/src/replication/ChangeStream.ts
@@ -713,11 +713,15 @@ export class ChangeStream {
                 // change stream events, collapse standalone checkpoints into the normal batch
                 // checkpoint flow to avoid commit churn under sustained load.
                 const hasBufferedChanges = eventIndex < events.length - 1;
-                if (waitForCheckpointLsn != null || hasBufferedChanges) {
-                  if (waitForCheckpointLsn == null) {
-                    waitForCheckpointLsn = await createCheckpoint(this.client, this.defaultDb, this.checkpointStreamId);
-                  }
+                if (hasBufferedChanges && waitForCheckpointLsn == null) {
+                  // Buffered changes - create a new batch checkpoint to rate limit commits
+                  waitForCheckpointLsn = await createCheckpoint(this.client, this.defaultDb, this.checkpointStreamId);
                   continue;
+                } else if (waitForCheckpointLsn != null) {
+                  // Skip this checkpoint - wait for the batch checkpoint.
+                  continue;
+                } else {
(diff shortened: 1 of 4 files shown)


modules/module-mongodb/src/replication/ChangeStream.ts
```diff
@@ -713,11 +713,15 @@ export class ChangeStream {
...
+                if (hasBufferedChanges && waitForCheckpointLsn == null) {
+                  // Buffered changes - create a new batch checkpoint to rate limit commits
+                  waitForCheckpointLsn = await createCheckpoint(this.client, this.defaultDb, this.checkpointStreamId);
                   continue;
+                } else if (waitForCheckpointLsn != null) {
+                  // Skip this checkpoint - wait for the batch checkpoint.
+                  continue;
+                } else {
+                  // No buffered changes, and no batch checkpoint pending - commit immediately.
                 }
```
rkistner, line 725: This change just makes the logic easier to follow - it does not change behavior.

stevensJourney (approved):
The changes here look good to me

Issue #565, opened 2026-03-14
Edition 3: aliasing synced table in JOIN query causes 0 rows to sync


jmalmo:
## Bug description

In Sync Streams edition 3, aliasing the **synced (source) table** in an explicit `INNER JOIN` query causes zero rows to sync on the client, with no validation or runtime errors.

### Sync config (failing)

```yaml
config:
  edition: 3

streams:
  user_chat:
    priority: 2
    auto_subscribe: true
    queries:
      - SELECT * FROM chat_conversations WHERE user_id = auth.user_id()
      - >
        SELECT cm.* FROM chat_messages cm
        INNER JOIN chat_conversations ON cm.conversation_id = chat_conversations.id
        WHERE chat_conversations.user_id = auth.user_id()
```

The second query syncs 0 rows. Removing the alias `cm` (using `chat_messages` directly) fixes it.

### Test matrix

| Variant | Query pattern | Source table aliased | Joined table aliased | Result |
|---------|--------------|---------------------|---------------------|--------|
| A | `SELECT cm.* FROM chat_messages cm INNER JOIN chat_conversations cc ON ...` | yes | yes | **0 rows** |
| B | `SELECT chat_messages.* FROM chat_messages INNER JOIN chat_conversations ON ...` | no | no | All rows |
| C | `SELECT * FROM chat_messages WHERE conversation_id IN (SELECT ...)` | n/a | n/a | All rows |
| D | `SELECT cm.* FROM chat_messages cm INNER JOIN chat_conversations ON ...` | yes | no | **0 rows** |
| E | `SELECT chat_messages.* FROM chat_messages INNER JOIN chat_conversations cc ON ...` | no | yes | All rows |

The bug occurs specifically when the synced (FROM) table is aliased in an explicit JOIN. Aliasing the joined table is fine. Subquery-based queries (variant C) are unaffected.

### Minimal DDL and seed data

```sql
CREATE TABLE chat_conversations (
  id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  user_id uuid NOT NULL,
  kind text NOT NULL,
  created_at timestamptz DEFAULT now()
);

CREATE TABLE chat_messages (
  id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  conversation_id uuid NOT NULL REFERENCES chat_conversations(id),
  user_id uuid,
  role text NOT NULL,
  content text NOT NULL,
  created_at timestamptz DEFAULT now()
);

[...]

PR #662, opened 2026-06-06
fix(sync-rules): surface JOIN clauses in Sync Streams as a loud error (#565)

sravan27:
Addresses #565.

## What this PR changes

This PR emits a non-fatal Sync Streams compiler warning when a stream joins multiple sources and the primary table selected by the stream has an unquoted alias.

That matters because an alias on the primary table changes the name the rows sync under. For example:

```sql
SELECT cm.*
FROM chat_messages cm
INNER JOIN chat_conversations
  ON cm.conversation_id = chat_conversations.id
WHERE chat_conversations.user_id = auth.user_id()
```

can compile, but rows sync under `cm` rather than `chat_messages`. If the alias was only added to simplify the join expression, clients querying `chat_messages` can unexpectedly see zero rows.

The warning explains that behavior and tells users to either:

- drop the alias on the source table when the rename was unintentional, or
- quote the alias, for example `AS "cm"`, when they intentionally want rows to sync under the alias name.

## Where the detection lives

The diagnostic runs in the new compiler after result columns have been processed, so the primary table is determined from the selected source row rather than from `FROM` clause order.

It warns when:

- the stream has more than one source in `FROM`, and
- the primary source row has an alias, and
- that alias was not explicitly double-quoted in the source SQL.

It is intentionally suppressed for quoted aliases such as `FROM user_data AS "users"`, which serve as an escape hatch for deliberate table renames.

## Tests

Regression coverage in `packages/sync-rules/test/src/compiler/errors.test.ts` covers:

- aliased primary table with a join warns that rows sync under the alias;
- quoted alias suppresses the warning;
- aliased joined tables do not suppress the warning when the primary table is still aliased;
- aliased primary table without a join does not warn;
- existing unsupported join diagnostics still include the new warning where appropriate.

Also touched intentional-alias fixtures in evaluator/table-valued tests so deliberate aliasing uses the quoted-alias escape hatch.

Local validation from the latest pushed head:

- `pnpm --dir packages/sync-rules build`
- `pnpm --dir packages/sync-rules test`
- Prettier check on touched files

## Why this style of fix

This follows the same Sync Streams correctness pattern as the earlier sprint fixes: surface surprising SQLite/Sync Streams semantic traps before users silently get wrong rows. It does not add broader JOIN support; it makes the existing alias behavior explicit and actionable for the #565 case.

### DIFF
--- a/.changeset/sync-rules-join-aliased-warning.md
+++ b/.changeset/sync-rules-join-aliased-warning.md
@@ -1,0 +1,8 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Surface joining onto an aliased primary table as a non-fatal warning in the Sync Streams compiler. The stream
+compiles successfully, but filter expressions that reference the joined table cannot be resolved through the
+alias and the stream may silently sync zero rows. Drop the alias on the primary table, or quote it
+(`FROM user_data AS "users", ...`) to acknowledge the constraint and suppress the warning. Addresses #565.
--- a/.changeset/sync-rules-join-aliased-warning.md
+++ b/.changeset/sync-rules-join-aliased-warning.md
@@ -2,7 +2,6 @@
 '@powersync/service-sync-rules': patch
 ---
 
-Surface joining onto an aliased primary table as a non-fatal warning in the Sync Streams compiler. The stream
-compiles successfully, but filter expressions that reference the joined table cannot be resolved through the
-alias and the stream may silently sync zero rows. Drop the alias on the primary table, or quote it
-(`FROM user_data AS "users", ...`) to acknowledge the constraint and suppress the warning. Addresses #565.
+Emit a warning for Sync Streams with joins where the primary table has an alias. Adding an alias syncs the table under the changed name, which may be unintentional for joins if aliases are only added to simplify filters.
+
+Closes #565.
--- a/packages/sync-rules/src/compiler/parser.ts
+++ b/packages/sync-rules/src/compiler/parser.ts
@@ -254,7 +254,6 @@ export class StreamQueryParser {
       return false;
     }
 
-    this.diagnoseAliasedPrimaryWithJoin(node.from);
     node.from?.forEach((f) => this.processFrom(f));
     if (node.where) {
       this.addAndTermToWhereClause(node.where);
@@ -262,6 +261,9 @@ export class StreamQueryParser {
 
     if (!options.forSubquery && node.columns) {
       this.processResultColumns(node, node.columns);
+      // Run after processResultColumns so the primary table (determined by the selected columns, not the
+      // order of the FROM clause) is known. See #565.
+      this.diagnoseAliasedPrimaryWithJoin(node.from);
     }
 
     this.warnUnsupported(node.groupBy, 'GROUP BY');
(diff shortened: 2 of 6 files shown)


packages/sync-rules/src/streams/from_sql.ts
```diff
@@ -469,6 +469,26 @@ class SyncStreamCompiler {
...
+          '(SELECT id FROM chat_conversations WHERE user_id = auth.user_id())`. ' +
+          'Tracking: https://github.com/powersync-ja/powersync-service/issues/565',
+        this.sql,
+        stmt._location
+      );
+    }
+
     if (stmt.from == null || stmt.from.length != 1 || stmt.from[0].type != 'table') {
       throw new SqlRuleError('Must SELECT from a single table', this.sql, stmt);
     }
```
simolus3, line 474: This changes an older implementation of Sync Streams which doesn't support joins at all. So it's unrelated to all test variants mentioned in https://github.com/powersync-ja/powersync-service/issues/565.

The solution would be to emit a (non-fatal!) warning in the new sync streams compiler (`src/compiler/`) if there's a join and the primary table has an alias. Ideally there should also be an escape hatch for cases where this is intentional, e.g. by allowing aliases with explicit quotes (`SELECT ... FROM user_data AS "users", $joins`).

simolus3 (changes requested):

sravan27:
recheck

sravan27:
Reworked per your guidance — thanks for the precise pointer.

Force-pushed: the loud-error in the deprecated `streams/from_sql.ts` is gone. Replaced with a non-fatal warning in the new compiler (`src/compiler/parser.ts`, `diagnoseAliasedPrimaryWithJoin`).

Decision matrix:

| Primary alias | Joined-table alias | Behavior |
| --- | --- | --- |
| no | (any) | no warning |
| yes (bare) | every join aliased | no warning |
| yes (bare) | at least one join unaliased | **warning** |
| yes (`AS "name"`) | (any) | no warning (escape hatch) |

The narrowed check is intentional — when every join target is also aliased, the author has already disambiguated scopes (`FROM users AS u JOIN orgs AS uom ON ...`) and the silent-row-loss footgun does not apply. So all the joins-feedback snapshots in `advanced.test.ts` still pass clean (verified: 143/143 in `test/src/compiler/` green).

Tests updated:
- `errors.test.ts` — three pre-existing `FROM users u JOIN orgs` tests now also expect the warning (the join target is unaliased there). Plus three new dedicated tests: silent-row-loss diagnostic, quoted-alias escape hatch, no-join-no-warn.
- `advanced.test.ts` snapshots — untouched, all still pass.

Changeset rewritten to match.

.changeset/sync-rules-join-aliased-warning.md
```diff
@@ -0,0 +1,8 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Surface joining onto an aliased primary table as a non-fatal warning in the Sync Streams compiler. The stream
+compiles successfully, but filter expressions that reference the joined table cannot be resolved through the
+alias and the stream may silently sync zero rows. Drop the alias on the primary table, or quote it
+(`FROM user_data AS "users", ...`) to acknowledge the constraint and suppress the warning. Addresses #565.
```
simolus3: > filter expressions that reference the joined table cannot be resolved through the
alias and the stream may silently sync zero rows

This is not true. The issue is that users might add an alias because it's convenient for joins, and then forget about the fact that this causes rows from the source table to be synced under that alias too (causing them to see zero rows on clients because they're querying the wrong table!).


```suggestion
Emit a warning for Sync Streams with joins where the primary table has an alias. Adding an alias syncs the table under the changed name, which may be unintential for joins if aliases are only added to simplify filters.

Closes #565.
```

packages/sync-rules/src/compiler/parser.ts
```diff
@@ -357,6 +358,70 @@ export class StreamQueryParser {
...
+   * unaliased. See #565: a query like
+   *
+   *   SELECT cm.* FROM chat_messages cm
+   *     INNER JOIN chat_conversations ON cm.conversation_id = chat_conversations.id
+   *     WHERE chat_conversations.user_id = auth.user_id()
+   *
+   * parses, but filter compilation cannot route the `chat_conversations.user_id` reference back to the
+   * joined table when the primary table wears an alias and the joined table is referenced by its bare name.
+   * Downstream the stream may silently degenerate and sync zero rows.
+   *
```
simolus3: Also not correct (if you have a test case observing this behavior, that would be another issue). Can you mention instead that the `cm` alias may have been introduced only to make the `ON` clause easier to write and users might be suprised to see this sync the table as `cm` instead of `chat_messages`?

packages/sync-rules/src/compiler/parser.ts
```diff
@@ -357,6 +358,70 @@ export class StreamQueryParser {
...
+   * AS uom ON ...`) and the silent-row-loss footgun does not apply.
+   *
+   * Escape hatch: if the primary alias is double-quoted in source SQL (e.g. `FROM user_data AS "users"`) we
+   * treat that as a deliberate signal that the author has accepted the constraint and suppress the warning.
+   */
+  private diagnoseAliasedPrimaryWithJoin(fromList: From[] | nil) {
+    if (!fromList || fromList.length < 2) {
+      return;
+    }
+    const primary = fromList[0];
```
simolus3: We can't assume that the first entry is the primary table, this is determined by the columns in the select statement. If this check runs after `processResultColumns`, we know about the primary table.

packages/sync-rules/src/compiler/parser.ts
```diff
@@ -357,6 +358,70 @@ export class StreamQueryParser {
...
+      if (entry.type == 'call' || entry.type == 'statement') {
+        return !entry.alias;
+      }
+      return false;
+    });
+    if (!hasUnaliasedJoin) {
+      return;
+    }
+    this.errors.report(
+      'Joining onto an aliased primary table is not currently supported and may silently sync zero rows: ' +
```
simolus3: This error message also needs to be updated to point out that while this is supported, the name of the synced table might be unexpected due to the alias.

sravan27:
Thanks for the detailed review @simolus3 — pushed updates addressing all of it:

- The check now resolves the primary table from `primaryResultSet` (determined by the selected columns) and runs **after** `processResultColumns`, instead of assuming the first `FROM` entry is the primary table.
- Reworked the warning to match your point: it now explains that the alias **renames the synced table** (rows sync under the alias; clients querying the real table name see zero rows), dropping the inaccurate "filter expressions … cannot be resolved through the alias" wording. Changeset updated to your suggested text.
- Kept the quoted-alias escape hatch, and the warning stays suppressed when every joined source is itself aliased (e.g. `json_each(...) AS x`), so idiomatic table-valued joins don't trip it.

Full `sync-rules` suite is green (900 passed / 7 skipped).

packages/sync-rules/src/compiler/parser.ts
```diff
@@ -357,6 +360,78 @@ export class StreamQueryParser {
...
+      }
+      return false;
+    });
+    if (!hasUnaliasedJoinTarget) {
+      return;
+    }
+    const alias = primary.source.explicitName;
+    const tableName = primary.tablePattern.name;
+    this.errors.report(
+      `The primary table is synced under its alias '${alias}' instead of '${tableName}'. ` +
```
simolus3: primary table is mostly an internal name, can we change this to `source row`?
```suggestion
      `The source row is synced under its alias '${alias}' instead of '${tableName}'. ` +
```

packages/sync-rules/src/compiler/parser.ts
```diff
@@ -357,6 +360,78 @@ export class StreamQueryParser {
...
+        return entry.name !== primary.source.origin && !entry.name.alias;
+      }
+      if (entry.type == 'call' || entry.type == 'statement') {
+        return !entry.alias;
+      }
+      return false;
+    });
+    if (!hasUnaliasedJoinTarget) {
+      return;
+    }
```
simolus3: I'm not sure we should have this check? Something like `SELECT u.* FROM users, orgs o WHERE p.user_id = u.id` still seems like something worth flagging to me? Allowing explicit aliases is enough of an escape hatch IMO.

packages/sync-rules/src/compiler/parser.ts
```diff
@@ -261,6 +261,9 @@ export class StreamQueryParser {
 
     if (!options.forSubquery && node.columns) {
       this.processResultColumns(node, node.columns);
+      // Run after processResultColumns so the primary table (determined by the selected columns, not the
```
simolus3: nit: Can we pass the `primaryResultSet` as a non-nullable parameter to `diagnoseAliasedPrimaryWithJoin` to make this explicit?

Also, instead of calling the function here, perhaps call it at the end of `processResultColumns` in an else branch of the final `if` checking for `this.primaryResultSet == null`.

sravan27:
Quick housekeeping note: I force-pushed this branch only to rewrite the latest commit author from the unmatched personal email to my signed GitHub noreply identity so CLA could pass. The code changes are the same as the prior review-addressing update; CLA is now green.

packages/sync-rules/src/compiler/parser.ts
```diff
@@ -357,6 +357,60 @@ export class StreamQueryParser {
     }
   }
 
+  /**
+   * Warn when a Sync Stream joins multiple tables, the primary table (the one the stream selects from) carries
+   * an alias, and at least one joined source is referenced by its bare (unaliased) name. See #565: a query like
```
simolus3:
```suggestion
   * Warn when a Sync Stream joins multiple tables and the primary table (the one the stream selects from) carries
   * an alias. See {@link github.com/powersync-ja/powersync-service/issues/565}: a query like
```

packages/sync-rules/src/compiler/parser.ts
```diff
@@ -357,6 +357,60 @@ export class StreamQueryParser {
...
+   *     INNER JOIN chat_conversations ON cm.conversation_id = chat_conversations.id
+   *     WHERE chat_conversations.user_id = auth.user_id()
+   *
+   * compiles successfully, but the alias renames the synced table: rows sync under `cm` rather than
+   * `chat_messages`. When joining, an alias is frequently introduced only to make the join's `ON` clause
+   * easier to write, in which case the rename is unintentional and clients querying `chat_messages` will
+   * unexpectedly see zero rows.
+   *
+   * The primary table is determined by the selected columns, not by the order of the `FROM` clause, so this
+   * runs after a non-null primary result set has been established.
```
simolus3: This feels irrelevant for a documentation comment

```suggestion
```

simolus3 (commented):
Two final nits, the implementation looks good to me now.

Could you also change the PR description to reflect the updated detection? Basically copying the changeset entry should be enough.

sravan27:
Addressed the final doc-comment nits from the latest review in d6174c8f.

Changes:
- updated the diagnostic comment to say the warning applies when the primary table has an alias in a multi-source Sync Stream;
- removed the internal implementation-detail sentence from the comment.

Validation:
- `pnpm --dir packages/sync-rules build`
- `pnpm --dir packages/sync-rules test` (32 files, 901 passed / 7 skipped)
- `pnpm prettier --check packages/sync-rules/src/compiler/parser.ts`
- `git diff --check`

simolus3 (approved):
Thanks 👍

PR #664, opened 2026-06-06
fix(sync-rules): length() should count Unicode code points to match SQLite

sravan27:
Same silent-data-loss class as the merged Sync Streams fixes #644 / #645 / #646 / #647, the open #565 JOIN loud-error PR #662, and the ASCII upper/lower PR #663.

### The bug

`length()` on text returned `value.length` — JavaScript's `String.prototype.length`, which counts **UTF-16 code units**. SQLite's `length(X)` returns the number of **characters** (Unicode code points).

For any string containing a non-BMP code point (emoji like `😀`, CJK Extension B–G, ancient scripts) the two counts diverge: each non-BMP code point is one UTF-16 surrogate pair = 2 code units in JS but 1 character in SQLite.

| Source value | Server `length()` (JS) | Client `length()` (SQLite) | Bucket key match? |
|---|---|---|---|
| `"hello"` | `5` | `5` | ✅ |
| `"straße"` | `6` | `6` | ✅ (ß is BMP) |
| `"😀"` | `2` | `1` | ❌ silent miss |
| `"a😀b"` | `4` | `3` | ❌ silent miss |
| `"𐀀"` (Linear B) | `2` | `1` | ❌ silent miss |

A bucket-parameter or filter using `length(...)` silently routed any row containing such characters to a different bucket than the client query asks for.

### Fix

A tiny `countCodePoints(text)` helper that iterates the string with `for…of` (which walks code points, not code units) and `length` uses it for the text case. The blob branch already matches SQLite (byte length).

### Tests

New `syncTest('length() counts Unicode code points, not UTF-16 code units (matches SQLite)')` in `sqlite_semantics.test.ts` covering:

- `"hello"` → `5n` (ASCII control)
- `"straße"` → `6n` (BMP non-ASCII, unchanged)
- `"😀"` → `1n` (was `2n`)
- `"a😀b"` → `3n` (was `4n`)
- `"𐀀"` (U+10000, Linear B) → `1n` (was `2n`)

Full `sync_rules.test.ts`: 40/40.

### DIFF
--- a/.changeset/sync-rules-length-code-points.md
+++ b/.changeset/sync-rules-length-code-points.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Make `length()` count Unicode code points (matching SQLite) rather than JavaScript UTF-16 code units. Previously a string containing a non-BMP code point (emoji like `😀`, CJK Extension B–G, ancient scripts, etc.) had its length silently doubled on the server, producing a different bucket key than the SQLite client computes for the same expression.
--- a/packages/sync-rules/src/sql_functions.ts
+++ b/packages/sync-rules/src/sql_functions.ts
@@ -157,6 +157,25 @@ const hex: DocumentedSqlFunction = {
   detail: 'Convert a blob to hex text'
 };
 
+/**
+ * SQLite's `length(X)` on text returns the number of *characters* in X
+ * (Unicode code points), not bytes and not UTF-16 code units. JavaScript's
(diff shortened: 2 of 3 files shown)


sravan27:
recheck

sravan27:
Acknowledged — closing reasoning carries over from #663 (Unicode-aware JS over ASCII-only SQLite is the preferred direction for the engine). Will park the SQLite-matching framing and look at the inverse: client-side overrides in powersync-js so the wa-sqlite client matches the Service's Unicode behavior. Thanks for the time.

PR #665, opened 2026-06-06
fix(sync-rules): substring() should index by code points to match SQLite

sravan27:
Same silent-data-loss class as the merged #644 / #645 / #646 / #647 and the open #565 PR #662, #663 (ASCII upper/lower), and #664 (length code points).

### The bug

`substring()` used `String.prototype.substring` and `String.prototype.length`, which count **UTF-16 code units**. SQLite's `substr()` indexes by **characters** (Unicode code points).

For any string containing a non-BMP code point — `😀`, CJK Extension B–G, ancient scripts — the two views disagree:

| Call | Server (JS) | SQLite client | Match? |
|---|---|---|---|
| `substring('hello', 1, 2)` | `'he'` | `'he'` | ✅ |
| `substring('a😀bc', 1, 2)` | `'a\uD83D'` (unpaired surrogate!) | `'a😀'` | ❌ broken |
| `substring('😀😀😀', 1, 2)` | `'😀'` (only first emoji's pair) | `'😀😀'` | ❌ |
| `substring('𐀀ab', 1, 2)` | `'𐀀'` (just the surrogate pair of one emoji) | `'𐀀a'` | ❌ |

The two-code-unit case happens to coincide for a single non-BMP code point, but the general case is broken — the JS path produces either an unpaired-surrogate string or splits characters across boundary indices.

### Fix

One change: walk the string into a code-point array (`const codePoints = [...text]`) once, and slice/length-check on that array instead of on the original string. All index arithmetic now operates on code points; `join('')` produces well-formed output.

### Tests

New `syncTest('substring() indexes by code points, not UTF-16 code units (matches SQLite)')` in `sqlite_semantics.test.ts`:

- `'hello'` → `'he'` (ASCII control)
- `'a😀bc'` → `'a😀'` (was `'a\uD83D'`)
- `'😀😀😀'` → `'😀😀'` (was a single emoji's surrogate pair)
- `'𐀀ab'` (U+10000 Linear B) → `'𐀀a'` (was just `𐀀` surrogate pair)

`sync_rules.test.ts`: 40/40 still green.

### DIFF
--- a/.changeset/sync-rules-substr-code-points.md
+++ b/.changeset/sync-rules-substr-code-points.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Make `substring()` index by Unicode code points (matching SQLite) rather than UTF-16 code units. Previously, slicing through a non-BMP code point (emoji, CJK Extension, ancient scripts) returned a broken unpaired surrogate or split the character in half — server output silently disagreed with SQLite client output.
--- a/packages/sync-rules/src/sql_functions.ts
+++ b/packages/sync-rules/src/sql_functions.ts
@@ -101,18 +101,29 @@ const substring: DocumentedSqlFunction = {
       // Different from undefined in this case, to match SQLite behavior
       return null;
     }
+    // SQLite's substr() indexes by *characters* (Unicode code points), not
+    // UTF-16 code units. JavaScript's `String.prototype.substring` and
+    // `String.prototype.length` count code units, so any non-BMP code point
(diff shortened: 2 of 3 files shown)


sravan27:
recheck

sravan27:
Acknowledged — closing reasoning carries over from #663 (Unicode-aware JS over ASCII-only SQLite is the preferred direction for the engine). Will park the SQLite-matching framing and look at the inverse: client-side overrides in powersync-js so the wa-sqlite client matches the Service's Unicode behavior. Thanks for the time.

PR #669, opened 2026-06-09
[Incremental Reprocessing] Sync plan equality checks

rkistner:
This implements equality checks for serialized sync plans. Specifically, this can compare:
1. `SerializedBucketDataSource` (requires some denormalization to lookup the `SerializedDataSource` definitions)
2. `SerializedParameterIndexLookupCreator`

This is to be used for incremental reprocessing, to preserve bucket data sources and parameter indexes if unchanged.

Since this operates on the serialized plans, only sync streams with `config.edition: 3` are supported.

Currently this requires `uniqueName` to be unchanged, to allow still using `uniqueName` in the bucket prefix. This means that:
1. A sync stream rename needs to reprocess the entire sync stream.
2. Changing order of queries in a sync stream _may_ require reprocessing:
    a. If the queries share the same `BucketDataSource` (same "bucket parameters"), order between them may change freely.
    b. If the same stream contains multiple `BucketDataSource`s (multiple queries with different "bucket parameters"), changing the order of those may trigger reprocessing.

For usage of these APIs, see #653. This PR is split out from that one to make it more manageable.

This also changes some interfaces to be public, so that we can access the types for serialized plans on the storage layer. The serialized format is considered stable now, so it should not introduce issues to make these public.

## Alternatives considered

1. Comparing parsed sync plans. This is feasible, but requires more logic. It also requires parsing sync plans just to compare them. See experimental implementation [here](https://github.com/powersync-ja/powersync-service/pull/653/changes/c7e8635af58897e3d54bd0d7a8820928e8f8b05b..9741d0924821640249a4aea35ef7d901eef93aa5)
2. Implementing an equality check directly on `PreparedStreamBucketDataSource` requires comparing `ScalarExpressionEvaluators` and other internals. We could use for example `ExpressionToSqlite.toSqlite` here, but overall this requires even more specific checks than the above. The one specific advantage this approach would have is that it could work for sync rules and legacy sync streams as well.
3. Using the existing equality checks the compiler uses for de-duplication. These currently only work when parsing directly from source, and would require significant restructuring to make these work with serialized plans, so not considered feasible.

## AI Usage

The implementation was mostly performed by Codex gpt-5.5. Manually evaluated the various approaches and tweaked the implementation.

### DIFF
--- a/packages/sync-rules/src/sync_plan/plan_equality_serialized.ts
+++ b/packages/sync-rules/src/sync_plan/plan_equality_serialized.ts
@@ -1,0 +1,47 @@
+import { Equality } from '../compiler/equality.js';
+import {
+  SerializedBucketDataSource,
+  SerializedDataSource,
+  SerializedParameterIndexLookupCreator
+} from './serialize.js';
+
+export interface SerializedBucketDataSourceWithDataSources {
+  bucket: SerializedBucketDataSource;
+  dataSources: readonly SerializedDataSource[];
+}
+
+export const serializedStreamDataSourceEquality = jsonEquality<SerializedDataSource>();
+
+export const serializedStreamParameterIndexLookupCreatorEquality =
+  jsonEquality<SerializedParameterIndexLookupCreator>();
+
+export const serializedStreamBucketDataSourceEquality: Equality<SerializedBucketDataSourceWithDataSources> = {
+  hash(hasher, value) {
+    hasher.addString(JSON.stringify(bucketIdentity(value)));
+  },
+  equals(a, b) {
+    return a === b || JSON.stringify(bucketIdentity(a)) == JSON.stringify(bucketIdentity(b));
+  }
+};
+
+function bucketIdentity(value: SerializedBucketDataSourceWithDataSources) {
+  const { bucket, dataSources } = value;
+
+  return {
+    hash: bucket.hash,
+    uniqueName: bucket.uniqueName,
+    // Sort so that the order of sources does not affect equality, as long as the same data sources are included.
+    sources: bucket.sources.map((index) => JSON.stringify(dataSources[index])).sort()
+  };
+}
+
+function jsonEquality<T>(): Equality<T> {
+  return {
+    hash(hasher, value) {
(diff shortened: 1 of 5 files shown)


packages/sync-rules/src/sync_plan/plan_equality_serialized.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export interface SerializedBucketDataSourceWithDataSources {
+  bucket: SerializedBucketDataSource;
+  dataSources: readonly SerializedDataSource[];
+}
+
+export const serializedStreamDataSourceEquality = jsonEquality<SerializedDataSource>();
+
+export const serializedStreamParameterIndexLookupCreatorEquality =
+  jsonEquality<SerializedParameterIndexLookupCreator>();
```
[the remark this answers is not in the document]
rkistner: That is fine for now.

simolus3: It's kind of obvious, but maybe add a comment to mention that data / parameter sources are self-contained structures which is why they can be compared through JSON.

We can also point out the `uniqueName` / `queryId` limitation there to potentially revisit it in the future.

packages/sync-rules/test/src/sync_plan/evaluator/bucket_data_source_equality.test.ts
```diff
@@ -0,0 +1,229 @@
...
+  SerializedBucketDataSourceWithDataSources,
+  serializedStreamBucketDataSourceEquality,
+  serializeSyncPlan,
+  SyncConfig
+} from '../../../../src/index.js';
+import { compileToSyncPlanWithoutErrors } from '../../compiler/utils.js';
+import { syncTest } from './utils.js';
+
+describe('prepared bucket data source equality', () => {
+  syncTest('matches equivalent prepared bucket sources from different plans', ({ sync }) => {
```
simolus3: nit: None of these have to be `syncTest`s, we can use a regular vitest `test` function

simolus3 (dismissed):

simolus3 (approved):

PR #670, opened 2026-06-09
[Incremental Reprocessing] Support for multiple sync configs on MongoDB storage

rkistner:
## High-level Changes

1. Append sync configs to existing replication streams if compatible, instead of always creating new streams. This enables incremental reprocessing!
2. Update many V3 operations to support actually using multiple sync configs in the same stream.
3. Remove the `batch.lastCheckpointLsn` API. It was only used by MongoDB to cover an edge case that is not relevant anymore.
4. V3 storage: Add `resume_lsn` as a stream-level identifier. The sync-config-specific `last_checkpoint_lsn` is now only used for write checkpoints. This simplifies the logic to get the LSN/resume token to resume from. In the future, we can update this on every flush, but don't want to increase the scope of this PR further.
5. V3 storage: Add `last_persisted_op` on the stream level, updated on every flush. This removes the need for the in-memory `this.persisted_op`, and fixes some edge cases. This also replaces the sync-config-specific `keepalive_op`, and removes the need to separately track `lastSnapshotOpId` in `MongoSnapshotter`.
6. V3 bucket prefixes: Updated to `${uniqueName}.${replicationStreamId}.${definitionId}`, e.g. `my_stream|0.1a.9` (uniqueName includes `|<index>` for sync streams).
7. Restructure sync config parsing to avoid re-parsing, so that we can get stable object references to use in the BucketDefinitionMappings.
8. Fix many edge cases in V3 `resolveTables`. Extract large parts to separate `source-table-utils`, to keep the actual `resolveTables` method manageable.

Note that some of the above are storage-level changes for storage version 3, and will break existing deployments using that version. That is intentional - storage version 3 is still considered unstable, and breaking changes are expected.

## Remaining issues / limitations

1. Custom write checkpoint events are not reprocessed automatically - see details below.
2. Existing data is not cleaned up after adding a sync config to an existing replication stream. Implemented in #675.
3. Dropping tables is still a blocking operation for the entire stream, unlike table snapshots. This is not a regression with using incremental reprocessing, just a shortcoming that is not yet addressed, that we can handle better with the new structure.

## Usage

Full usage examples and documentation will follow before making this a stable feature.

In the meantime, deploy sync streams with `storage_version: 3`:

```yaml
config:
  edition: 3
  storage_version: 3
```

## Example

Logs after editing a single sync query:

```
info: Incremental reprocessing sync config update:
Reused definitions:
  - type=bucket_data, id=1, name=globalb1|0
  - type=bucket_data, id=2, name=globalb2|0
  - type=bucket_data, id=3, name=globalb4|0
New definitions:
  - type=bucket_data, id=6, name=globalb13|0, tables=todos
Definitions to drop after switching:
  - type=bucket_data, id=5, name=globalb13|0

```

## Custom Write Checkpoints

Custom write checkpoints are currently scoped per replication stream, not per sync config. These do not trigger re-snapshots if the event query changes.

During processing of new sync config, if the event query changed, write checkpoints would be created using _both_ queries. If the produced values change, the result may be ambiguous.

Recommendation for deploying substantial changes to write checkpoint event queries:
1. Deploy all other changes, and wait for reprocessing to complete.
2. Deploy the write checkpoint event query changes. This should require no re-processing and complete immediately.
3. Trigger an update on all existing write checkpoint source rows to trigger re-replication.

## AI Usage

Used Codex gpt-5.5 and Claude Fable 5 signficantly to iterate on implementing and reviewing these changes. Manually reviewed and tweaked each part.


modules/module-mongodb-storage/src/storage/implementation/v3/source-table-utils.ts
```diff
@@ -0,0 +1,382 @@
...
+    )
+  );
+}
+
+function matchingSourceTableIdentity(doc: SourceTableDocumentV3, identity: SourceTableIdentity) {
+  return (
+    doc.schema_name == identity.schema &&
+    doc.table_name == identity.name &&
+    (identity.objectId == null || doc.relation_id == identity.objectId) &&
+    sameReplicaIdColumns(doc.replica_id_columns, identity.replicaIdColumns)
```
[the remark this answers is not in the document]
rkistner: Intentional change to the storage format.

simolus3 (commented):
The `BucketDefinitionMapping` / `updateSyncRulesV3` changes and the sync config parsing refactorings look good to me. My only question is whether it could make sense to move some of that (e.g. `BucketDefinitionMapping` itself and also the `MongoBucketStorage.isCompatible` check) into something shared between implementations? It might help to be more explicit about what we expect to share in the future and what is specific to each storage implementation.

I had a look at the rest too, but I don't think I'm familiar enough with the bucket batch mechanism to be an effective reviewer for changes 3-5.

rkistner:
> My only question is whether it could make sense to move some of that (e.g. `BucketDefinitionMapping` itself and also the `MongoBucketStorage.isCompatible` check) into something shared between implementations? It might help to be more explicit about what we expect to share in the future and what is specific to each storage implementation.

Good call. I moved these to `storage/implementation` in service-core, to indicate that it's re-usable code, but not part of the public storage API at this point.

modules/module-mongodb-storage/src/storage/implementation/MongoPersistedReplicationStream.ts
```diff
@@ -1,18 +1,18 @@
 import { mongo } from '@powersync/lib-service-mongodb';
 import { ServiceAssertionError } from '@powersync/lib-services-framework';
-import { storage } from '@powersync/service-core';
+import { SingleSyncConfigBucketDefinitionMapping, storage } from '@powersync/service-core';
+import { ReplicationStreamStorageIds } from '@powersync/service-core/src/storage/implementation/ReplicationStreamStorageIds.js';
```
stevensJourney: Nit. we should probably export this from the `service-core` package and update the import statement.

modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
```diff
@@ -21,25 +21,25 @@ import {
   utils,
   WatchWriteCheckpointOptions
 } from '@powersync/service-core';
+import { ReplicationStreamStorageIds } from '@powersync/service-core/src/storage/implementation/ReplicationStreamStorageIds.js';
```
stevensJourney: This import should also be updated

modules/module-mongodb-storage/src/storage/implementation/MongoPersistedReplicationStream.ts
rkistner: Just realized the same thing - it was an auto-import during refactoring. It works in tests, but will fail in the Docker build. I now configured an `exports` field in each package to make this break hard, which should prevent this from happening again in the future.

modules/module-mongodb-storage/src/storage/implementation/v3/MongoBucketBatchV3.ts
```diff
@@ -45,6 +51,79 @@ export class MongoBucketBatchV3 extends MongoBucketBatch {
...
+    }
+    return [...ids.values()];
+  }
+
+  private snapshotBlockingSourceTablesFilter(): Record<string, unknown> {
+    const clauses = this.syncConfigIds.flatMap((syncConfigId) => {
+      const filter = this.mapping.snapshotBlockingSourceTablesFilter(syncConfigId.toHexString()) as {
+        $or?: Record<string, unknown>[];
+      };
+      return filter.$or ?? [];
```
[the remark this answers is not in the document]
rkistner: Snapshots are not currently considered important for events, but the current approach for events is not great. Will address in a follow-up PR.

libs/lib-services/package.json
```diff
@@ -5,6 +5,11 @@
   "main": "dist/index.js",
   "type": "module",
   "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "default": "./dist/index.js"
+    }
+  },
```
simolus3: I'm definitely a fan of adding `exports` 👍

Given that we only expect these packages to be processed by modern toolchains, we should probably also remove the old `main` and `types` entries from `package.json`. Node.js docs describe it like this:

> For new packages targeting the currently supported versions of Node.js, the ["exports"](https://nodejs.org/api/packages.html#exports) field is recommended. [...] If both ["exports"](https://nodejs.org/api/packages.html#exports) and ["main"](https://nodejs.org/api/packages.html#main) are defined, the ["exports"](https://nodejs.org/api/packages.html#exports) field takes precedence [...]

So we might as well be consistent about not supporting these old exports:

```suggestion
  "type": "module",
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "default": "./dist/index.js"
    }
  },
```

simolus3 (dismissed):

stevensJourney (dismissed):
The changes here look good from my side.

> Custom write checkpoints are currently scoped per replication stream, not per sync config. These do not trigger re-snapshots if the event query changes. During processing of a new sync config, if the event query changed, write checkpoints would be created using both queries. If the produced values change, the result may be ambiguous.

This does seem like a fairly risky situation, and one that developers could hit accidentally if they miss the recommended deployment sequence.

I wonder whether we should eventually scope custom write checkpoints by `sync_config_id` as well as the replication stream id. That would make the active and processing configs write/read separate checkpoint records, avoiding accidental pollution of the active config while a new config is being processed. It would still leave the issue that historical custom checkpoints are not automatically rebuilt for the new config, so the existing recommendation to update all checkpoint source rows would still matter.

Longer term, I wonder if checkpoint event source tables should participate in snapshot/reprocessing too, so a new sync config can rebuild its custom checkpoint state from source rows instead of relying on manual row updates. That feels like follow-up work rather than something this PR needs to solve.

libs/lib-services/package.json
rkistner: I now simplified further to this for most packages:
```
  "type": "module",
  "exports": "./dist/index.js",
```

The types are inferred automatically if only the extension is different, so no need to explicitly specify.

The only exceptions are packages that provide some explicit exports such as `./schema.json`, which I preserved.

Along with this I now simplified the tsconfigs across packages:
* Avoid setting explicit `declarationDir` or `tsBuildInfoFile` - the default is fine.
* Avoid setting explicit config if `tsconfig.base.json` already specifies it.

I also changed the `@powersync/service-types` package to have `"type": "module"`, consistent with other packages, which required fixing some imports to use the `.js` extension.

rkistner:
> Longer term, I wonder if checkpoint event source tables should participate in snapshot/reprocessing too, so a new sync config can rebuild its custom checkpoint state from source rows instead of relying on manual row updates. That feels like follow-up work rather than something this PR needs to solve.

Yes, I think that would be best, and would match the behavior without incremental reprocessing. That would require keeping track of the queries per sync config, detecting whether there are changes requiring a re-snapshot. So it's a big change, and I'm not going to bloat this PR further, but I consider that a blocker before marking the V3 storage as stable.

modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
```diff
@@ -165,6 +182,47 @@ export class MongoBucketStorage extends storage.BucketStorageFactory {
...
+      );
+      if (active != null) {
+        const existingConfigDocs = await this.loadSyncConfigDefinitions(versioned, active, session);
+
+        if (
+          this.options.supportsMultipleSyncConfigs &&
+          isCompatible(
+            existingConfigDocs.map((d) => d.serialized_plan ?? null),
+            options.config,
+            logger
```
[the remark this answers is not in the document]
rkistner: Acknowledged, will address in a follow-up PR.

simolus3 (approved):
tsc/exports changes look good to me

PR #671, opened 2026-06-09
[Incremental Reprocessing] Storage model refactor

rkistner:
This is a refactor to prepare for #670. The intent is to keep this primarily to a purely refactoring, while #670 contains the actual implementation changes.

The old `PersistedSyncRules` + `PersistedSyncRulesContent` is split into:
1. `PersistedReplicationStream`: This represents the overall replication stream - mostly for job management.
2. `PersistedSyncConfigContent` - the content for individual sync configs inside a replication stream. This is mostly an immutable representation of the config.
3. `PersistedSyncConfigStatus` - this contains the status such as whether the snapshot is complete.
4. `ParsedSyncConfigSet`: This represents a parsed version of one or more SyncConfigs in the same replication stream. Does not have to cover all SyncConfigs in the stream, depending on usage.
 
This also refactors some fields:
1. `PersistedSyncRulesContent.id` or `group_id` is renamed to `replicationStreamId`, to distinguish from `syncConfigId`.
2. `slot_name` is renamed to `replicationStreamName`, to indicate the more general use case outside of Postgres.

This also splits out some classes into separate files.

This does not attempt to completely remove "sync rules" in API and variable names yet - mostly just in the APIs touched by this PR for now. A future PR can continue that work.

Minor behavioral change: The "reprocess" API, when there is already sync config processing, now produces a proper `PSYNC_S4106` error code, instead of a generic error.

## AI Usage

Most refactorings were a combination of IDE refactorings and Codex gpt-5.5. Manually specified the new class and API structure.


stevensJourney (approved):
After going through all the changes from this and skimming over #670 - this looks good to me. I could not think of any potential issues or concerns from my side.

simolus3 (approved):
I mostly skimmed over the module changes, but the refactoring and `service-core` changes make sense to me.

PR #673, opened 2026-06-11
feat: S3 object storage offloading for V3 bucket data

Sleepful:
## Summary

Offload `BucketDataDocumentV3.ops[]` arrays to object storage (S3), keeping only a metadata shell in MongoDB. The service reads S3 objects and streams ops to clients using the existing wire protocol — no protocol changes. Object storage is optional at configuration level; when not configured, all ops remain inline in MongoDB as today.

## Design Decisions

- ~~No inline threshold~~ → `inline_threshold_bytes` (default 256). Chunks below this go inline.
- ~~S3 path format `bucket-data/<group>/<def>/<bucket>/<minOp>-<maxOp>`~~ → paths now include `.bson.zstd` suffix and carry Content-Type `application/bson` + Content-Encoding `zstd` metadata on the S3 object.
- Zstd whole-object compression — centralized in `BucketDataObjectStorage` wrapper (one place for BSON serialization + zstd compress on store, decompress + deserialize on retrieve).
- ~~`compressed_size * 3` heuristic~~ → `doc.size` — the write path already stores the actual decompressed data size on the metadata shell; use it directly.
- **`has_clear_op` top-level flag** on `BucketDataDocumentV3`, set during write and compaction. Lets the checksum pipeline detect CLEAR ops without filtering on embedded `$ops`.
- **S3 straddle fixup in checksums**: when a checkpoint falls mid-document for an S3-backed doc, the MongoDB aggregation can't filter on `$ops` (they're on S3). Post-aggregation JS fetches ops from S3, filters by range, and recomputes checksum/count/clear_op.
- **Orphan cleanup via `pending_s3_deletes`**: pending deletes are persisted inside the compaction transaction. A crash after commit but before S3 delete leaves entries that the next compaction retries.
- **Hard errors on S3 fetch failure**: a corrupt or missing S3 object throws immediately — no silent `doc.ops = []`.
- **MinIO toggle**: set `MINIO_ENDPOINT` to run all S3 tests against a real S3-compatible endpoint. Internal-state tests auto-skip against real S3.
- **No explicit credentials in config** — deferred. Real AWS uses the default credential chain (IAM roles, env vars, `~/.aws/credentials`). Explicit `access_key_id`/`secret_access_key` fields are only needed for self-hosted S3-compatible endpoints in production, which isn't a target yet.

## Manual Verification

S3ObjectStorage is not exercised in CI. To manually validate against MinIO:

```bash
docker rm -f minio 2>/dev/null
docker run -d --name minio -p 9000:9000 minio/minio server /data --console-address ":9001" &&
sleep 2 &&
docker run --rm --network host --entrypoint sh minio/mc -c "
  mc alias set local http://localhost:9000 minioadmin minioadmin &&
  mc mb local/powersync-s3-test --ignore-existing
" &&
MINIO_ENDPOINT=http://localhost:9000 pnpm --filter='./modules/module-mongodb-storage' test
```

The `MINIO_ENDPOINT` toggle switches all S3 tests from MemoryObjectStorage to the real `S3ObjectStorage` client. Default credentials (`minioadmin` / `minioadmin`) are used — `forcePathStyle` is already set when an endpoint is present.


modules/module-mongodb-storage/src/storage/implementation/v3/MongoCompactorV3.ts
```diff
@@ -613,6 +829,18 @@ export class MongoCompactorV3 extends MongoCompactor {
       await session.endSession();
     }
 
+    // After commit: delete old S3 objects (best-effort)
```
rkistner: Not a blocker for initial testing, but it could be problematic if we leave orphaned documents in the bucket indefinitely (either from the delete request failing, or from say a process crash/restart between the commit and the delete).

Is there some way we can ensure these are cleaned up eventually? Maybe persisting a "delete queue" in mongodb, or running a periodic cleanup job (maybe part of the compact job)?

modules/module-mongodb-storage/src/storage/implementation/v3/MongoSyncBucketStorageV3.ts
```diff
@@ -474,17 +474,48 @@ export class MongoSyncBucketStorageV3 extends MongoSyncBucketStorage {
...
+              }
+            })
+          );
+        }
+      }
+
+      // Track sizes: for S3 docs multiply compressed_size by 3 as a rough
+      // decompressed estimate to keep chunk byte tracking bounded. Without a
+      // multiplier, metadata shells (~200 bytes) would let thousands of
+      // S3-backed docs pack into a single chunk before splitting.
```
rkistner: We already have the size on the mongodb document - could we use that instead of the estimate?

modules/module-mongodb-storage/src/storage/implementation/v3/MongoSyncBucketStorageV3.ts
```diff
@@ -474,17 +474,48 @@ export class MongoSyncBucketStorageV3 extends MongoSyncBucketStorage {
...
+          await Promise.all(
+            s3Docs.map(async (doc) => {
+              try {
+                const buffer = await this.objectStorage!.get(doc.storage_ref.path);
+                const decompressed = await zstd.decompress(buffer);
+                const wrapper = bson.deserialize(decompressed, storage.BSON_DESERIALIZE_INTERNAL_OPTIONS);
+                doc.ops = wrapper.ops;
+              } catch (err) {
+                this.logger.warn(`Failed to fetch/decompress S3 object ${doc.storage_ref?.path}: ${err}`);
+                doc.ops = [];
```
rkistner: This should be a hard error - setting `doc.ops = []` may result in data inconsistencies.

rkistner (commented):
This looks quite promising, and I like the structure.

Some initial high-level comments:
1. Currently there are various places in the code doing the  same compression/decompression and serialize/deserialize logic. Should we perhaps do this in a wrapper class for ObjectStorage? E.g. a BucketDataObjectStorage that wraps ObjectStorage and does that logic?
2. NodeJS now has built-in zstd support. But I haven't checked how the APIs and performance compares with `@mongodb/zstd`. Since we're already using `@mongodb/zstd` implicitly, that should be fine.
3. We do need a threshold for inlining ops directly in mongodb storage, before we can merge & release this: S3 has too much overhead for storing say individual 100-byte operations.

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/BucketDataObjectStorage.ts
```diff
@@ -0,0 +1,28 @@
...
+import { ObjectStorage } from './ObjectStorage.js';
+
+export class BucketDataObjectStorage {
+  constructor(private readonly storage: ObjectStorage) {}
+
+  async store(path: string, ops: BucketOperation[]): Promise<{ compressedSize: number }> {
+    const bsonBuffer = Buffer.from(bson.serialize({ ops }));
+    const compressedUint8 = await zstd.compress(bsonBuffer);
+    const compressed = Buffer.from(compressedUint8);
+    await this.storage.put(path, compressed);
```
rkistner: I think we should include the content-type and content-encoding here:

```
  ContentType: "application/bson",
  ContentEncoding: "zstd",
```

This will allow us to for example changing the compression algorithm later, without causing compatibility issues with existing files.

ContentType is not strictly needed, but I think it's good practice to store anyway.

It may be good to add add an extension to the path as well, e.g. `.bson.zstd`. For in-memory or filesystem implementations, we can use the extension to indicate whether or not the file is compressed, instead of metadata (if we ever add compression support there).

Sleepful:
**Addressed**
- Extracted compression into `BucketDataObjectStorage` wrapper (single place for BSON+zstd)
- Replaced `compressed_size * 3` with `doc.size` — already stored on the metadata shell
- S3 fetch failures are now hard errors (throw), not silent `doc.ops = []`
- Orphan cleanup via `pending_s3_deletes` MongoDB collection — persisted inside compaction transaction, retried on next run
- Added `has_clear_op: boolean` top-level field on `BucketDataDocumentV3`, set during write and compaction
- Checksum pipeline: `$ifNull` wraps all `$ops` references
- `inline_threshold_bytes` config — chunks below threshold stay inline in MongoDB (default 256, catches tiny CLEAR docs and small write batches)
- Content-Type/Content-Encoding metadata + `.bson.zstd` suffix on S3 paths

**Gaps found and fixed**
- `clearBucketLeading` was missing S3 pre-fetch (`loadBucketDataDocument` needs `doc.ops`) — added same pattern as MOVE pass
- CLEAR pass aggregation projection was missing `storage_ref` — pre-fetch couldn't detect S3-backed docs
- **S3 straddle fixup in checksum pipeline**: when a checkpoint falls mid-document (`is_fully_included = false`) for an S3-backed doc, the pipeline's `$filter` on `$ops` returns empty (ops are on S3, not embedded). After the aggregation, for each doc matching this condition, we fetch ops from S3, filter by the range `(bucket_start, bucket_end]` in JS, and recompute `checksum_total`, `count_total`, and `has_clear_op` — overwriting the pipeline's zeroes.


**Tests added**
- 3 red-checksum tests: start straddle, end straddle, CLEAR checksum preservation
- 1 scoped delete test with partial `maxOpId` on S3-backed docs
- 1 inline threshold test (small ops survive S3 loss, large ops fail hard)
- All S3 tests toggleable: `MINIO_ENDPOINT=http://localhost:9000 pnpm test:core`
- Test count: 398 passing, 0 failures

**Manual tests**
- Performed manual tests against MINIO as substitute for S3 protocol. 

<details>
<summary>
Running MINIO with Docker
</summary>

```sh
docker rm -f minio 2>/dev/null
docker run -d --name minio -p 9000:9000 minio/minio server /data --console-address ":9001" &&
sleep 2 &&
docker run --rm --network host --entrypoint sh minio/mc -c "
  mc alias set local http://localhost:9000 minioadmin minioadmin &&
  mc mb local/powersync-s3-test --ignore-existing
" &&
MINIO_ENDPOINT=http://localhost:9000 pnpm --filter='./modules/module-mongodb-storage' test
```
</details>

**Deferred**

- **No explicit credentials in config** — deferred. Real AWS uses the default credential chain (IAM roles, env vars, `~/.aws/credentials`). Explicit `access_key_id`/`secret_access_key` fields are only needed for self-hosted S3-compatible endpoints in production, which isn't a target yet.

AlexGrafe:
Can any storage provider be used for this, like Google Cloud Storage (GCS) ?

rkistner:
> Can any storage provider be used for this, like Google Cloud Storage (GCS) ?

@AlexGrafe Our initial release will just focus on AWS S3, but we do have plans to support GCS, Azure Storage, and potentially some others shortly afterwards. Since we only use fairly simple storage operations, there should be no issue to support most providers - it just requires hooking up the relevant libraries and config.

rkistner:
Superseded by #728 (based on this PR).

PR #679, opened 2026-06-18
[docs] Replication Docs

stevensJourney:
## Overview / context

The replication path; database source connectors; the shared `service-core` loop; and the bucket storage writer are some of the more intricate parts of the service, and most of that knowledge lived in people's heads or in scattered notes. This PR attempts to start documenting the replication process in tracked docs files.

It starts by writing it down under `docs/replication` so the contracts/behaviours are documented, both so people can understand the existing modules and so a future source module (a new database connector) can be built against a documented contract rather than by reverse-engineering. It then surfaces those docs through a new `AGENTS.md` and a plan template, so agents can use the same context when building or changing replication modules.

These docs are not authoritative or complete, and aren't meant to be a final word on how replication works. They're a starting point to orient people and agents, and we expect to refine and extend them over time.

## What's included

**1. Replication specs (`docs/replication/`)**

A new numbered set of specs describing how replication works: the end-to-end flow (with a sequence diagram), core concepts and vocabulary, the replication lifecycle, the source-connector and storage-writer contracts, snapshots and streaming, checkpoints and write checkpoints, a comparison of the current Postgres, MongoDB, MySQL, SQL Server, and Convex modules, and the `resolveTables` lifecycle deep dive. Each page links to the real code anchors so the docs stay grounded in the implementation, and the README gives an outside-in map plus a reading path so individual pages can be loaded selectively.

**2. Documentation restructure**

The existing flat `docs/*.md` files were reorganized into a clearer taxonomy, each folder with an index README. Code comments and module READMEs that referenced the old paths were updated to match (no logic changes), and a pre-existing `parameters-lookups.md` → `parameter-lookups.md` filename typo was fixed along the way.

```
docs/
├── README.md                # top-level index
├── replication/             # replication protocol specs (numbered, outside-in) + resolveTables deep dive
├── specs/                   # client-facing protocol specs
│   └── sync-protocol.md
├── storage/                 # bucket storage data structures, invariants, compaction
└── modules/                 # source-module-specific notes (convex, postgres, ...)
```

**3. `AGENTS.md`**

A new root `AGENTS.md` that routes coding agents to the right spec page by task ("storage writer change → read these two pages") and tells them to load spec context selectively rather than pulling in everything. It also captures the day-to-day dev commands and testing conventions.

**4. AI agent plan template for new replication modules**

`08-ai-agent-plans.md` builds on the specs to give agents a phased plan for implementing a brand-new replication module — from feasibility notes through module skeleton, schema discovery, snapshot, streaming, write checkpoints, and hardening — with per-phase acceptance checks and a prompt template, so a new connector can be delivered as a series of small, reviewable changes rather than one large patch.

---
AI usage disclosure. These docs were generated by both Codex GPT 5.5 Claude Opus 4.8.  I instructed a general goal structure, reviewed the outputs and made various corrections additions as required.


docs/replication/03-source-connector-contract.md
```diff
@@ -0,0 +1,196 @@
...
+- MongoDB reads session cluster time and updates the `_powersync_checkpoints` collection.
+- SQL Server reads the latest LSN and updates the PowerSync checkpoints table.
+- Convex reads the head cursor and writes a checkpoint marker.
+- MySQL currently reads the executed GTID and calls the callback; the adapter notes a TODO for forcing a later replicated message.
+
+The adapter also exposes source schema or source metadata information in service-friendly formats:
+
+- `getConnectionSchema()` extracts the connected database schema as `DatabaseSchema[]` when the source has strict schema. Admin and sync-rules endpoints use this to validate sync config against actual source tables, collections, columns, and types.
+- `getDebugTablesInfo()` expands table patterns from sync config into concrete source table information and per-table warnings or errors. This is used by validation and diagnostics paths to explain what will be replicated.
+- `getParseSyncRulesOptions()` provides source defaults, such as the default schema or database used when sync config references an unqualified table name.
```
rkistner: I think documentation on this level of individual methods is better as just jsdoc comments in the source itself - it is very easy to get out of sync in a separate doc. We can keep the docs here as high-level conceptual docs, without describing the specific methods.

I'd make an exception for `createReplicationHead`, since it covers a fairly core concept.

docs/replication/03-source-connector-contract.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Replicator
+
+A replication module provides an `AbstractReplicator` subclass.
+
+It must implement:
+
+- `createJob(options)`: constructs the source-specific replication job for a locked storage stream.
+- `cleanUp(syncRuleStorage)`: removes source-side state for a stopped replication stream.
+- `testConnection()`: checks source connectivity using the decoded source config.
```
rkistner: Same here - maybe include a brief overview of the functionality, but I wouldn't reference the specific methods.

docs/replication/04-storage-writer-contract.md
```diff
@@ -0,0 +1,131 @@
+# Storage Writer Contract
+
+Source-specific streams started by replication jobs write through `SyncRulesBucketStorage` and `BucketStorageBatch`. This contract is implemented by bucket storage modules, currently Postgres and MongoDB storage.
```
rkistner: I haven't checked each individual detail here, but I think a "contract" like this is better covered by the interface and its jsdoc comments - I'd recommend rather putting effort in to make those easier to follow.

docs/replication/05-snapshots-and-streaming.md
```diff
@@ -0,0 +1,113 @@
+# Snapshots And Streaming
```
rkistner, line 1: As a comparison to "Storage Writer Contract", I think this one is fairly good. It does have some references to specific methods that may get out of date, but it's not that many, and the broad description here covers "what should the process look like" instead of "what does the individual APIs do", which is a good level for these docs.

docs/replication/07-source-modules.md
```diff
@@ -0,0 +1,89 @@
...
+Schema change handling is also source-specific. A module may automatically react to table, collection, column, or replica identity changes during streaming, or it may require operators to deploy a new sync config after source schema changes. Source-specific documentation should state which model applies.
+
+## Postgres
+
+Primary files:
+
+- [`WalStreamReplicator`](../../modules/module-postgres/src/replication/WalStreamReplicator.ts)
+- [`WalStreamReplicationJob`](../../modules/module-postgres/src/replication/WalStreamReplicationJob.ts)
+- [`WalStream`](../../modules/module-postgres/src/replication/WalStream.ts)
+- [`PostgresRouteAPIAdapter`](../../modules/module-postgres/src/api/PostgresRouteAPIAdapter.ts)
```
rkistner: I don't think the "primary files" list is very useful here - these are fairly easy to find already.

AGENTS.md
```diff
@@ -0,0 +1,61 @@
...
+
+- Prefer existing module patterns over new abstractions.
+- Keep changes scoped to the package or module relevant to the task.
+- For integration tests, prefer the existing test utilities and fixtures in the relevant module.
+- Prefer real implementations for tests that validate cross-module behavior. If a test only needs to observe calls or arguments, prefer Vitest spies such as `vi.spyOn()` before introducing mocks. Use mocks only when testing specific unit behavior or when a real dependency is impractical.
+
+## Development Commands
+
+- This workspace uses Node from [`.nvmrc`](.nvmrc) and pnpm from the root [`packageManager`](package.json). If the shell does not automatically load nvm or Corepack, run Node/pnpm commands through `source ~/.nvm/nvm.sh && nvm use && corepack pnpm ...`.
+- Install dependencies with `corepack pnpm install`. CI enables Corepack before installing dependencies; match that locally when possible.
```
rkistner: Are there cases where this is specifically required? In a typical local dev setup this should already be configured, and you can just do `pnpm install` / `pnpm build` / etc. But I'm not sure how good the cloud versions are at the setup.

rkistner (commented):
This looks very useful overall.

I added some specific comments, but to re-iterate the broad approach: I think these docs should only focus on the "big picture" things (general concepts, processes, etc), and use jsdoc comments for things like what methods to implement or what specific methods do. It's easier to follow the docs for the specific methods in the source itself, and it won't get out of date as easily there.

AGENTS.md
stevensJourney: It might be something related to my dev setup. I do have my local shells configured to automatically enable these, but I've noticed a few cases where an AI agent would trip-up and try to use the wrong version of Node/PNPM for some unknown reason - which would cause it to go into loops trying to solve the numerous issues that causes.

docs/replication/07-source-modules.md
stevensJourney: 👍  Removed those

docs/replication/04-storage-writer-contract.md
stevensJourney: 👍 I've updated the general flow to be more of an overview.

docs/replication/03-source-connector-contract.md
stevensJourney: 👍 updated

docs/replication/03-source-connector-contract.md
stevensJourney: 👍  Implemented this change

stevensJourney:
> This looks very useful overall.
> 
> I added some specific comments, but to re-iterate the broad approach: I think these docs should only focus on the "big picture" things (general concepts, processes, etc), and use jsdoc comments for things like what methods to implement or what specific methods do. It's easier to follow the docs for the specific methods in the source itself, and it won't get out of date as easily there.

I've updated this to now focus more on the broader "big picture" :) . I've also updated the docs after pulling in `main` - which included some incremental reprocessing updates - these incremental reprocessing doc updates were AI generated, I reviewed them and they LGTM - but it would be good to have another eye on the details there.

rkistner (approved):
I did not check all the details, but the structure looks good. If there are minor inaccuracies, we can fix those over time.

PR #683, opened 2026-06-23
[POC] Bucket storage report (operations vs rows)

bean1352:
# Bucket storage report

## What this adds

A new admin endpoint, `POST /api/admin/v1/bucket-report`. It shows how much operation history each bucket carries compared to its actual rows.

## How it works

The report only reads `bucket_state`. It never scans the operation history.

- `operations`, `operation_bytes`: kept up to date by writers and compactors. Exact.
- `rows`: the row count captured by the bucket's last full compact. `null` if the bucket has never been fully compacted.
- `fragmentation`: `operations / rows`. Near 1 is healthy. 167 means new clients download 167 operations per live row.
- `uncompacted_operations`: operations written after that compact, which is how stale `rows` might be.
- `last_full_compact_at`, `next_compact_at`: when the row stats were captured, and when the scheduler next considers the bucket. A suggested compact with a future date is already planned, just throttled until then.
- `suggested_action`: `compact`, `defragment`, `both`, `none`, or `unknown` when there are no compact statistics to reason from.

Worst buckets come first. There is also a rollup per definition and totals for the instance. `limit` defaults to 50 and caps at 1000; invalid values get a 400 error instead of being clamped.

## Example

```
POST /api/admin/v1/bucket-report
Authorization: Bearer <admin token>

{ "limit": 50 }
```

Response (from a real test run). This bucket was fully compacted at 07:35, then 15 more operations were written to it, so `rows` is a slightly stale snapshot and the report says so:

```json
{
  "buckets": [
    {
      "bucket": "by_user.1.3[\"u1\"]",
      "operations": 48,
      "operation_bytes": 9022,
      "uncompacted_operations": 15,
      "rows": 5,
      "fragmentation": 9.6,
      "last_full_compact_at": "2026-08-26T07:35:41.325Z",
      "next_compact_at": "2026-08-26T07:41:34.066Z",
      "suggested_action": "both"
    }
  ],
  "definitions": [
    {
      "definition": "by_user.1.3",
      "bucket_count": 4,
      "operations": 192,
      "operation_bytes": 36124,
      "uncompacted_operations": 60,
      "rows": 20,
      "fragmentation": 9.6,
      "suggested_action": "both"
    }
  ],
  "totals": {
    "bucket_count": 11,
    "operations": 281,
    "operation_bytes": 62302,
    "estimated": false
  },
  "buckets_truncated": false,
  "definitions_truncated": false
}
```

## Scaling

Up to 50k buckets the scan is exact. Above that the report ranks a sample of about 10k buckets, using only the `_id` index, and marks the totals `estimated: true`. Past 1M buckets it fails fast instead of scanning without bound.

Storage v1/v2 do not record compact statistics, so they report operation counts only: rows, fragmentation and the dates are `null` and the action is `unknown`. That keeps v1 cheap and nudges users towards v3.

## AI disclaimer

I developed this change using Claude, and reviewed and tested it myself.


modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
```diff
@@ -371,6 +441,359 @@ export abstract class MongoSyncBucketStorage
...
+    // estimatedDocumentCount is O(1) but ignores the match filter, so this is an upper bound on the active
+    // bucket count. That is fine for the sampling decision: over-estimating only switches to sampling sooner.
+    // It must NOT be used to scale the sampled totals though - the collection can hold buckets outside the
+    // match (other replication groups for v1/v2, inactive definitions for v3), which would over-scale.
+    const estimatedTotalBuckets = await collection.estimatedDocumentCount();
+    const sampled = estimatedTotalBuckets > BUCKET_SELECTION_SAMPLE_THRESHOLD;
+
+    const pipeline: mongo.Document[] = [{ $match: match }];
+    if (sampled) {
+      pipeline.push({ $sample: { size: BUCKET_SELECTION_SAMPLE_SIZE } });
```
rkistner: `$sample` does not help for performance unless it's the first stage in the pipeline.

Potential options:
1. `$sample` first, then filter. That would require the initial sample size to be higher than the limit we want.
2. Filter first, then use `$sampleRate`. I'm not actually what the performance is like for `$sampleRate` - would need some testing.

And if you go for option 2, node that current `_id.b` / `_id.g` filters aren't efficient either, and require a full collection scan. Do filter efficiently, you need to use a pattern such as `_id: {$gte: ..., $lt: ...}` - there should be a couple of examples like that in this repo you can use as a starting point.

packages/service-core/test/src/bucket-report.test.ts
```diff
@@ -0,0 +1,231 @@
...
+
+describe('resolveBucketReportLimit', () => {
+  it('defaults when no limit is given', () => {
+    expect(resolveBucketReportLimit(undefined)).toBe(DEFAULT_BUCKET_REPORT_LIMIT);
+  });
+
+  it('floors and clamps to a positive integer', () => {
+    expect(resolveBucketReportLimit(2.7)).toBe(2);
+    expect(resolveBucketReportLimit(-5)).toBe(1);
+    expect(resolveBucketReportLimit(0)).toBe(1);
```
rkistner: I'd recommend responding with an error for these cases, rather than clamping the value.

packages/service-core/src/storage/bucket-report.ts
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * Normalize a requested limit to a positive integer, falling back to {@link DEFAULT_BUCKET_REPORT_LIMIT}.
+ */
+export function resolveBucketReportLimit(limit?: number): number {
+  if (limit == null) {
+    return DEFAULT_BUCKET_REPORT_LIMIT;
+  }
+  return Math.max(1, Math.floor(limit));
+}
```
rkistner, line 206: We should also have a reasonable maximum, otherwise it could be easy to overload the service or storage database by passing in arbitrary large values.

rkistner (commented):
I had a quick check on `$sampleRate` performance, and unfortunately it's not a magic bullet in terms of performance. Essentially MongoDB still has to scan through the index at least before applying the `$sampleRate`. I _can_ work, as long as we ensure:
1. Limit the number of index entries a query is scanning to around 100k-1M _at most_.
2. Make sure no document lookup is performed before filtering using `$sampleRate`. E.g. if we scan through 100k index entries to sample around 1000 of them, the query must scan through 1000 documents, not 100k documents (fetching documents is much slower than the index entries). Use MongoDB's _explain_ to confirm this.

As an additional safeguard, we can use `readPreference: 'secondaryOnly'` for these queries, to make sure they don't affect performance on the primary node.

rkistner:
@bean1352 The MongoDB V3 storage had a significant change now in what we store in `bucket_state`: #752

I believe that should allow using _only_ the `bucket_state` collection to compute these reports, perhaps with minor tweaks. It would probably still require sampling on that collection to handle cases of say 100M+ buckets, but should not require reading `bucket_data` anymore.

More specifically:
1. bucket_state contains all the stats we use to determine whether or not to compact. Note that there are two types of compact per bucket now: A fast chunk-merge compact, and a full compact. In most cases, users should not need to compact manually, since a scheduled compact can run much more often now. But reporting on it could still be useful to see the cases such as "a compact would be beneficial, but it is throttled until X".
2. Fragmentation ratio can be read from the full compact state, something like `1.0 - (last_full_compact.puts / last_full_compact.count)`. That would not be quite a live counter, but should give a good enough indication. And since it's computed when the compact was run, each put operation is generally unique, so there's no need to try and count actual unique rows.

This would still leave cases with V1 storage. I'd recommend focusing on V3, and give more limited stats for V1, rather than attempting to do a more expensive scan for V1. That would further encourage moving users over to V3.

modules/module-mongodb-storage/src/storage/implementation/MongoSyncBucketStorage.ts
bean1352: The bucket report does not read `bucket_data` anymore.

The only sampling left is for picking the top buckets in `bucket_state (when above 50k buckets)` and it uses option 2

packages/service-core/src/storage/bucket-report.ts
bean1352: The maximum is 1000 now.

`export const MAX_BUCKET_REPORT_LIMIT: number = 1_000;`

bean1352:
> @bean1352 The MongoDB V3 storage had a significant change now in what we store in `bucket_state`: #752
> 
> I believe that should allow using _only_ the `bucket_state` collection to compute these reports, perhaps with minor tweaks. It would probably still require sampling on that collection to handle cases of say 100M+ buckets, but should not require reading `bucket_data` anymore.
> 
> More specifically:
> 
> 1. bucket_state contains all the stats we use to determine whether or not to compact. Note that there are two types of compact per bucket now: A fast chunk-merge compact, and a full compact. In most cases, users should not need to compact manually, since a scheduled compact can run much more often now. But reporting on it could still be useful to see the cases such as "a compact would be beneficial, but it is throttled until X".
> 2. Fragmentation ratio can be read from the full compact state, something like `1.0 - (last_full_compact.puts / last_full_compact.count)`. That would not be quite a live counter, but should give a good enough indication. And since it's computed when the compact was run, each put operation is generally unique, so there's no need to try and count actual unique rows.
> 
> This would still leave cases with V1 storage. I'd recommend focusing on V3, and give more limited stats for V1, rather than attempting to do a more expensive scan for V1. That would further encourage moving users over to V3.

The bucket report only reads `bucket_state` now.

The row count of a bucket comes from `last_full_compact.puts`, like you suggested. For the fragmentation score, we report `operations` / `rows`. Should I change that to `1.0 - (last_full_compact.puts / last_full_compact.count)`?

1. The `rows` field for a bucket stays `null` until the first full compact runs on that bucket. The compact that runs after initial replication only merges chunks and does not write the `last_full_compact` stats. Should it write them too, so rows are available from the start?
2. Above 50k buckets the report falls back to sampling and above 1M buckets it returns a 422 response. Is this fine for now?

PR #684, opened 2026-06-23
[WIP] Graceful replication job handover

rkistner:
The first issue here was with lock handling when using incremental reprocessing: Usually, when we update sync config, we take a lock immediately, so that the job updating the config is guaranteed to get the lock. The main use case that immediate lock solves is rolling deploys: Start a new replication process with new sync config, before terminating the existing replication process. However, when the sync config is added to an existing replication stream, that lock may already be held by another replication process, which then fails the lock acquisition, which fails the entire update.

The immediate fix is simple enough: Just don't attempt to lock in that case. But we can do better.

This adds a signal handler for SIGUSR2, which lets the replicator stop accepting any new replication jobs. If sync config is added to a replication stream, that stream is restarted _in the new process_. The existing process will detect that, and stop gracefully.

This is especially useful in Kubernetes environments: We can configure a preStop command that sends `kill -USR2 1`, then waits for say 30 seconds. That will continue processing the replication stream until the new pod is taking over, or times out after 30 seconds.

### DIFF
--- a/libs/lib-services/src/signals/termination-handler.ts
+++ b/libs/lib-services/src/signals/termination-handler.ts
@@ -13,7 +13,7 @@ type TerminationHandlerParams = {
   /**
    * A list of process signals to listen for.
    *
-   * @default ['SIGTERM', 'SIGINT', 'SIGUSR2']
+   * @default ['SIGTERM', 'SIGINT']
    */
   signals?: Signal[];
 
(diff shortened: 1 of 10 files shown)


rkistner:
Superseded by #686. The approach here doesn't actually work in practice: A kubernetes rolling update starts the new process first, _then_ calls `preStop` on the old one. That makes the signal handler fairly useless for Kubernetes rolling updates, which was the main case it attempted to cover.

PR #686, opened 2026-06-24
Restructure replication locks

rkistner:
The primary issue here was with lock handling when using incremental reprocessing: Usually, when we update sync config, we take a lock immediately, so that the job updating the config is guaranteed to get the lock. The main use case that immediate lock solves is rolling deploys: Start a new replication process with new sync config, before terminating the existing replication process. However, when the sync config is added to an existing replication stream, that lock may already be held by another replication process, which then fails the lock acquisition, which fails the entire update.

The immediate fix is simple enough: Just don't attempt to lock in that case. But that's not ideal for rolling updates: An older process can start the replication job, potentially running an older service version.

This changes how the replication process picks jobs: If a process updated the sync config from loaded config (filesystem or env vars), we ignore any replication streams that has a different "processing" sync config. We do still handle different "active" sync configs (i.e. older sync config).

This also tweaks the polling interval for replication streams: For the first 2 minutes, or until the first replication job is started, we use a shorter 500ms polling interval. This helps to reduce the handover time on rolling deploys, without increasing steady-state load. This does not yet address the slow polling interval for fail-over purposes - those are more complex to handle without increasing load.

Then we also re-order replication jobs: We now stop existing jobs before starting new ones. This specifically helps to restart a replication stream when the active config changes, without waiting for another refresh cycle.

## AI usage

Implemented by Codex gpt-5.5; manually reviewed and simplified.

### DIFF
--- a/.changeset/calm-squids-push.md
+++ b/.changeset/calm-squids-push.md
@@ -1,0 +1,7 @@
+---
+'@powersync/service-module-postgres-storage': patch
+'@powersync/service-module-mongodb-storage': patch
+'@powersync/service-core': patch
+---
+
+Restructure replication locks for updated sync config.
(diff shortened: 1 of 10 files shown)


stevensJourney (approved):
The changes here make sense. I can't think of any potential issues with this implementation :)

PR #689, opened 2026-06-25
Warn for `*` column after alias

simolus3:
As mentioned in https://github.com/powersync-ja/powersync-service/pull/685, queries like `SELECT public_id AS id, *` can be problematic as the `*` can overwrite `AS id`.

This adds a warning for that case. It's possible to be smarter about this by consulting the actual schema of resolved tables, but that can then cause issues for deployed Sync Streams if a conflicting column is added later. So this simple check should be good enough, and we probably don't want to encourage putting `*` last in any case given the risk.

I used Claude Code with minor manual revisions to generate this.

### DIFF
--- a/packages/sync-rules/src/compiler/parser.ts
+++ b/packages/sync-rules/src/compiler/parser.ts
@@ -471,7 +471,7 @@ export class StreamQueryParser {
             selectsFrom(resolved, column.expr);
             if (seenExplicitAlias) {
               this.errors.report(
-                `A '*' column after aliased columns may give unexpected results: If the synced row contains a column named '${seenExplicitAlias}', '*' would overwrite it.`,
+                `A '*' column after aliased columns may give unexpected results: If the source row contains a column named '${seenExplicitAlias}', '*' would overwrite it.`,
                 column.expr,
                 { isWarning: true }
               );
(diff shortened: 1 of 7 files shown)


packages/sync-rules/src/compiler/parser.ts
```diff
@@ -462,12 +462,20 @@ export class StreamQueryParser {
...
+    let seenExplicitAlias: string | undefined;
     for (const column of columns) {
       if (column.expr.type == 'ref' && column.expr.name == '*') {
         const resolved = this.resolveTableName(column.expr, column.expr.table?.name);
         if (resolved != null) {
           if (resolved instanceof BaseSourceResultSet) {
             selectsFrom(resolved, column.expr);
+            if (seenExplicitAlias) {
+              this.errors.report(
+                `A '*' column after aliased columns may give unexpected results: If the synced row contains a column named '${seenExplicitAlias}', '*' would overwrite it.`,
```
rkistner:
```suggestion
                `A '*' column after aliased columns may give unexpected results: If the source row contains a column named '${seenExplicitAlias}', '*' would overwrite it.`,
```

rkistner (approved):

PR #690, opened 2026-06-25
Release tweaks

rkistner:
These are minor tweaks to the release workflow:
1. Prevent command injection using github ref. The workflow is fairly safe since only approved users can run it, but it's better to do this properly.
2. Remove the `stevenontong/${{vars.DOCKER_REGISTRY}}:cache` cache in test builds - this cache did not work anyway, and I'm restricting the access of the default dockerhub user now.
3. Run dev images in the `dockerhub` environment, same as the next and stable releases.
4. For dev releases, add a check to ensure we don't accidentally publish a "stable" tag if there are no changesets.

This allows us to use separate users for dockerhub:
1. A read-only user for normal actions - only using an authenticated user here to avoid dockerhub rate limits.
2. A priviledged user in the `dockerhub` environment. This is only used when we need to push to dockerhub, and these workflows now require an additional approval step.


.github/workflows/test.yml
```diff
@@ -60,7 +60,6 @@ jobs:
       - name: Test Build Docker Image
         uses: docker/build-push-action@v5
         with:
-          cache-from: type=registry,ref=stevenontong/${{vars.DOCKER_REGISTRY}}:cache
```
stevensJourney, line 63: Wow, this must have survived for quite some time. Glad it's removed now :)

stevensJourney (dismissed):

stevensJourney (approved):

rkistner:
@stevensJourney I added a check to ensure the docker-dev workflow doesn't publish a stable version if there are no changesets. The "This branch had an error being deployed" is expected - that's this check in action.

stevensJourney:
Nice! The new changes look good to me :D

PR #691, opened 2026-06-25
Batch concurrent write checkpoint requests

rkistner:
We've seen some cases of users calling the `write-checkpoint2.json` API at 500 requests/s or more in peak periods. This easily exceeds the default router handler's concurrency limit of 10 and the queue limit of 20, especially if other db load is slowing down these requests. That results in high rates of 429 errors sent to clients. Clients automatically retry, but we'd rather avoid those completely.

This optimization notes that we can optimize this significantly under load:
1. If we have many concurrent write checkpoint requests, we only need to get a single "replication head" on the source database for all of them. The only condition here is that it needs to start _after_ the last call was received - we cannot add serve new requests from an existing in-flight replication head request.
2. On the storage database, we can similarly optimize this to do a single batched database call, instead of individual calls for each request.

These optimizations apply only if we get concurrent requests. Sequential requests behave the same as before, with no additional delay. Right now we support up to 2 in-flight batches - any new requests will be queued and form part of a new batch.

This now also places checkpoint requests on a separate fastify queue, with up to 50 requests processed concurrently, and an additional 100 requests queued. Since these requests typically have low overhead, and increased concurrency improves the efficiency of batching, this should not add significant overhead. The split is specifically to prevent the increased limits from affecting other APIs such as the relatively expensive diagnostic APIs.

## Benchmarks

Test case: Request 200x concurrent write-checkpoints for individual clients. Test with various fastify concurrency settings and batching settings. Added artificial MongoDB latency affecting both the source db and storage db, resulting in a single API call taking around 1s:

```js
db.adminCommand({
     configureFailPoint: "failCommand",
     mode: "alwaysOn",
     data: {
       failCommands: ["findAndModify"],
       blockConnection: true,
       blockTimeMS: 500
     }
   })
 ```

The first case here ran on the main branch, the rest on this branch. For all of these I increased the queue size to avoid 429 errors, for a fair comparison.

Not taken into account here:
1. Latency of individual requests
2. Source database and replication overhead from creating many individual checkpoint requests when not using batching.

| Fastify concurrency | Concurrent requests | Batching mode  | Duration |
| ------------------: | ------------------: | -------------- | -------: |
|                  10 |                 200 | Main branch    |    22.2s |
|                  10 |                 200 | No batching    |    21.5s |
|                  10 |                 200 | Max 2x batches |    25.4s |
|                  50 |                 200 | No batching    |    11.2s |
|                  50 |                 200 | Max 3x batches |     5.2s |
|                  50 |                 200 | Max 2x batches |     5.4s |
|                  50 |                 200 | Max 1x batch   |     6.2s |
|                 100 |                 200 | No batching    |    13.8s |
|                 **100** |                 200 | **Max 3x batches** |     **2.6s** |
|                 100 |                 200 | Max 2x batches |     2.6s |
|                 200 |                 200 | Max 2x batches |     1.6s |

It is clear from the results that increasing the fastify concurrency increases the throughput. But when not using batching, this is limited by the database connection pool size. Batching removes that bottleneck, reducing both the number of database commands and database connections in use. This would be even more relevant when there is high sync load, which typically co-occurs with high write checkpoint rates.

Note that the 200x requests here is significantly higher more than the 500/s observed in production: The 500/s is the _total_ for the instance, spread out over 20+ processes, so around 25/s per process.

## AI usage

Manually planned the implementation; used Codex gpt-5.5 to implement. Reviewed with Claude Opus 4.8.

### DIFF
--- a/packages/service-core/src/storage/WriteCheckpointAPI.ts
+++ b/packages/service-core/src/storage/WriteCheckpointAPI.ts
@@ -58,7 +58,7 @@ export type LastWriteCheckpointFilters = CustomWriteCheckpointFilters | ManagedW
 export interface BaseWriteCheckpointAPI {
   readonly writeCheckpointMode: WriteCheckpointMode;
   setWriteCheckpointMode(mode: WriteCheckpointMode): void;
-  createManagedWriteCheckpoint(checkpoint: ManagedWriteCheckpointOptions): Promise<bigint>;
+  createManagedWriteCheckpoints(checkpoints: ManagedWriteCheckpointOptions[]): Promise<Map<string, bigint>>;
 }
 
 /**
--- a/packages/service-core/src/system/ServiceContext.ts
+++ b/packages/service-core/src/system/ServiceContext.ts
@@ -19,7 +19,6 @@ export interface ServiceContext {
   storageEngine: storage.StorageEngine;
   migrations: PowerSyncMigrationManager;
   syncContext: SyncContext;
-  writeCheckpointBatcher: utils.WriteCheckpointBatcher;
   serviceMode: ServiceContextMode;
   eventsEngine: EventsEngine;
 }
@@ -51,7 +50,6 @@ export class ServiceContextContainer implements ServiceContext {
   eventsEngine: EventsEngine;
   syncContext: SyncContext;
   routerEngine: routes.RouterEngine;
-  writeCheckpointBatcher: utils.WriteCheckpointBatcher;
   serviceMode: ServiceContextMode;
 
   constructor(options: ServiceContextOptions) {
@@ -61,8 +59,16 @@ export class ServiceContextContainer implements ServiceContext {
 
     this.lifeCycleEngine = new LifeCycledSystem();
 
+    this.routerEngine = new routes.RouterEngine();
+    this.lifeCycleEngine.withLifecycle(this.routerEngine, {
+      stop: (routerEngine) => routerEngine.shutDown()
+    });
+
     this.storageEngine = new storage.StorageEngine({
-      configuration
+      configuration,
+      getCreateReplicationHead: () => {
+        return (callback) => this.routerEngine.getAPI().createReplicationHead(callback);
+      }
     });
     this.storageEngine.registerListener({
       storageFatalError: (error) => {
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

File: packages/service-core/src/storage/WriteCheckpointAPI.ts [hunk 1]
```diff
@@ -58,7 +58,7 @@ export type LastWriteCheckpointFilters = CustomWriteCheckpointFilters | ManagedW
 export interface BaseWriteCheckpointAPI {
   readonly writeCheckpointMode: WriteCheckpointMode;
   setWriteCheckpointMode(mode: WriteCheckpointMode): void;
-  createManagedWriteCheckpoint(checkpoint: ManagedWriteCheckpointOptions): Promise<bigint>;
+  createManagedWriteCheckpoints(checkpoints: ManagedWriteCheckpointOptions[]): Promise<Map<string, bigint>>;
```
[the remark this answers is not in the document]
rkistner: Not relevant - the NPM packages are not considered a public API.

packages/service-core/src/system/ServiceContext.ts
```diff
@@ -19,6 +19,7 @@ export interface ServiceContext {
   storageEngine: storage.StorageEngine;
   migrations: PowerSyncMigrationManager;
   syncContext: SyncContext;
+  writeCheckpointBatcher: utils.WriteCheckpointBatcher;
```
stevensJourney: Not a blocker: I was wondering if it would make sense for the `storageEngine` to contain `writeCheckpointBatcher`. That way storage engine would contain more of the storage specific items, but, if there are dependency issues - I think this is perfectly fine.

stevensJourney (dismissed):
The logic here looks sound to me. Happy with these changes.

packages/service-core/src/system/ServiceContext.ts
rkistner: Yeah, I think it makes sense. It's not fully clean since it also needs a reference to the RouteAPI createReplicationHead, but I think it's still better - see the latest changes.

packages/service-core/src/system/ServiceContext.ts
```diff
@@ -59,8 +59,16 @@ export class ServiceContextContainer implements ServiceContext {
 
     this.lifeCycleEngine = new LifeCycledSystem();
 
+    this.routerEngine = new routes.RouterEngine();
+    this.lifeCycleEngine.withLifecycle(this.routerEngine, {
+      stop: (routerEngine) => routerEngine.shutDown()
+    });
```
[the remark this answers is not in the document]
rkistner: Noted, but that's fine here.

_Update: It's not fine. Reverted the refactor in #698._

stevensJourney (approved):

packages/service-core/src/system/ServiceContext.ts
rkistner: Turns out my refactor caused an issue: #698. Codex actually picked it up and I incorrectly dismissed it: https://github.com/powersync-ja/powersync-service/pull/691#discussion_r3479851280 (I thought it was an edge case affecting write checkpoints only, but it affects all sync connections).

It may be possible to do a better refactoring here, but looking at this again, I don't think that refactoring is worth it. Since the `writeCheckpointBatcher` depends on both the `RouterEngine` and `StorageEngine`, we can't cleanly encapsulate it in one of those.

PR #693, opened 2026-06-27
fix(sync-rules): length() and instr() should count code points to match SQLite

sravan27:
Follow-up to the `substring()` code-point fix — the same UTF-16-vs-code-point divergence affects two sibling functions.

### Problem
`length()` used JS `String.length` (UTF-16 code units) and `instr()` used `String.indexOf` (a UTF-16 index), so both diverge from SQLite for text containing non-BMP characters (emoji, mathematical alphanumerics, etc.):

| expression | SQLite | sync-rules (before) |
|---|---|---|
| `length('😀')` | `1` | `2` |
| `length('a😀b')` | `3` | `4` |
| `instr('😀x', 'x')` | `2` | `3` |

Because sync-rule expressions are evaluated client-side and must agree with the source database, a filter like `length(name) > 10` or `instr(...)` over text containing emoji could include/exclude rows differently on the client than on the server — silent sync divergence.

### Fix
Both functions now iterate by code point (spreading the string / counting code points in the prefix), matching SQLite and the existing `substring()` behaviour. Blob handling is unchanged (still byte counts, which is correct per SQLite).

### Tests
Added regression cases to `sql_functions.test.ts` for `length` and `instr` covering non-BMP characters, including the case where a non-BMP character precedes the match (the one that actually diverges). Full sync-rules suite passes.

Verified against the system `sqlite3` as ground truth.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/sync-rules-length-instr-codepoints.md
+++ b/.changeset/sync-rules-length-instr-codepoints.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Fix `length()` and `instr()` to count characters (Unicode code points) rather than UTF-16 code units, matching SQLite. Previously `length('😀')` returned `2` instead of `1`, and `instr()` reported positions shifted by any non-BMP character before the match. This caused sync-rule expressions over text containing emoji or other non-BMP characters to diverge from the source database (the same class of issue as the earlier `substring()` code-point fix).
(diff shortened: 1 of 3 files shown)


rkistner:
@sravan27 As mentioned before, we're moving towards using the node:sqlite engine, and are not fixing further compatibility differences with our JS engine. See #666.

PR #694, opened 2026-06-27
fix(sync-rules): CAST(real AS INTEGER) should truncate towards zero like SQLite

sravan27:
### Problem
The number→integer path of `cast()` uses `BigInt(Math.floor(value))`, which rounds towards −∞. SQLite truncates a REAL towards **zero** when casting to INTEGER, so the two diverge on every negative non-integer real:

| expression | SQLite | sync-rules (before) |
|---|---|---|
| `CAST(-3.7 AS INTEGER)` | `-3` | `-4` |
| `CAST(-0.9 AS INTEGER)` | `0` | `-1` |
| `CAST(-2.1 AS INTEGER)` | `-2` | `-3` |

Since sync-rule expressions are evaluated client-side and must agree with the source database, an integer cast over a negative real (directly or via numeric coercion) can include/exclude rows differently on the client than on the server.

### Fix
Use `Math.trunc` instead of `Math.floor`. Positive reals are unaffected (floor == trunc there), and the string/bigint cast paths are unchanged. Added regression tests; verified against the system `sqlite3` as ground truth.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/sync-rules-cast-integer-truncation.md
+++ b/.changeset/sync-rules-cast-integer-truncation.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Fix `CAST(... AS INTEGER)` to truncate REAL values towards zero, matching SQLite. Previously a negative non-integer real was floored (`CAST(-3.7 AS INTEGER)` returned `-4` instead of `-3`), causing sync-rule expressions that cast negative reals to integer to diverge from the source database.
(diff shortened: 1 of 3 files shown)


rkistner:
@sravan27 As mentioned before, we're moving towards using the node:sqlite engine, and are not fixing further compatibility differences with our JS engine. See #666.

PR #695, opened 2026-06-29
Configurable bucket count cache duration

rkistner:
In some cases the bucket checksums and counts are expensive to calculate, so this helps reduce the overhead.

Note that this is a "TTL after use", not "TTL after calculation". So there may be some cases where the cached value is persisted for a long time.

Using a cached value should never result in an incorrect checksum. The main caveat of a longer TTL is that the count may be incorrect after a compact operation, resulting in client-side progress bars being slightly inaccurate. This is why I call this the "bucket count cache" duration instead of "checksum cache" duration.

The default here is still 60 minutes, but it can be increased to a couple of hours without significant risk.


stevensJourney (approved):
Looks good to me :)

PR #696, opened 2026-06-29
feat: Checkpoint Requests

stevensJourney:
# Overview

This adds the PowerSync service component for `requestCheckpoint`, as mentioned in these proposals:

- https://github.com/orgs/powersync-ja/discussions/324
- https://github.com/orgs/powersync-ja/discussions/317

This is related to the following open PRs:

- https://github.com/powersync-ja/powersync-sqlite-core/pull/198

This PR adds a `/sync/checkpoint-request` route which clients can use to create checkpoint requests.

Checkpoint requests currently flow through the standard write checkpoint flow in the sync protocol. The implementation here uses the existing collections/tables for write checkpoints for general checkpoint requests.

## Collections

Using the same collections for the previous and current checkpoint requests has a few advantages.

### Sync protocol

Checkpoint requests currently flow through the existing `write_checkpoint` marker in checkpoint events. We use the existing `lastWriteCheckpoint` logic for this. This works regardless of the checkpointing method used by the client.

### Migrations

Client and PowerSync service versioning migrations (upgrades and downgrades) are _almost_ compatible by default. Bucket storage migrations are required for MongoDB indexes and for the new Postgres storage columns. If an existing client has a current write checkpoint record, the `client_id` is preserved and future requests use monotonically increasing IDs (there are some exceptions to this, though—more on that later).

### Cleanup

One of the `Current Issues` in https://github.com/orgs/powersync-ja/discussions/317 is:

> If an app has many anonymous/temporary users, or regularly creates new temporary databases with unique client ids, it may end up with many write checkpoints on the service. We can never clean these up, since we don't know whether a client would ever connect again and need that write checkpoint. While each individual write checkpoint is slow, this can add up over time when you have hundreds of thousands of unique users/clients.

The goal is to have requested checkpoint records be temporary, so that records can be deleted after a period of time. For the current write checkpoint requests, `we can never clean these up`. Using the same collection allows us to update/mark existing write checkpoints as requested, allowing these records to be deleted.

This PR handles cleanup by marking checkpoint records associated with client checkpoint requests with a `checkpoint_requested_at` `Date` field. The existing `compact` job deletes expired request-derived records. The retention period defaults to 60 minutes and is configurable via `api.parameters.checkpoint_request_retention_minutes` (only configurable for self-hosted deployments).

Retrying the current checkpoint request refreshes `checkpoint_requested_at` without replacing its original source head, while stale requests with a lower ID do not refresh the timestamp. MongoDB storage also retains managed checkpoint requests until `processed_at_lsn` is set, preventing cleanup while replication is still lagging. Postgres storage and custom checkpoint requests use timestamp-based cleanup.

## Details

The Postgres and MongoDB bucket storage implementations have been updated to accommodate either the current auto-incrementing `write-checkpoint2.json` endpoint behaviour or the new ability to specify a requested checkpoint ID.

The storage update behaviour diverges based on whether a requested checkpoint ID has been supplied.

Supplied `checkpoint_request_id` values are validated as positive, signed 64-bit integers.

### No-ops

https://github.com/orgs/powersync-ja/discussions/317 mentions that checkpoint requests should be no-ops when the request ID is unchanged. This PR takes this slightly further by treating request IDs as a monotonic sequence. A larger value advances the stored request; retrying the current value refreshes its retention timestamp without replacing the stored source head; and a lower stale value leaves the record unchanged. The PowerSync service always returns the largest stored value as part of the `/sync/checkpoint-request` response. Clients can use this information to correct for certain edge cases.

For `write-checkpoint2.json` requests, we always make a change and always advance the replication head.

For client-supplied checkpoint requests, source-marker behaviour depends on storage:

- MongoDB tracks whether the stored request has been processed. It advances the source when a request is new or still pending, but can skip the marker for an already-processed duplicate or stale request.
- Postgres does not track a per-row processed state. It therefore advances the source for every supplied request, including stale or duplicate requests, to ensure a potentially pending checkpoint is observable. This can produce a harmless extra marker for an already-processed request.

### Client Migrations

The client needs to track and manage an increasing `checkpoint_request_id`. For existing users, this means they might have an existing write checkpoint record. The client should start its request sequence at or above this value (if it exists) in order to prevent setting a `target_op` below a consistency boundary.

Clients typically also re-issue their current checkpoint request on connect due to its temporary nature. This refreshes the retention timestamp without moving the request's original source head. If a newly migrated client does not have the checkpoint request ID sequence seeded, it is free to attempt a checkpoint request at `1`. If a record exists with a larger ID, the service will leave that record unchanged and return the current largest request ID, which the client can detect and use to re-seed its sequence.

One caveat remains if a checkpoint request record has already been deleted and the client no longer has a seeded checkpoint request value: the client has to start from `1`. This is acceptable because both sides have lost their previous state and the next request establishes a new sequence.

For MongoDB storage, a managed checkpoint request is not deleted while it is still pending replication. For Postgres storage, retention remains timestamp-based, so clients should retry their current request while waiting; each equal retry refreshes its retention timestamp without replacing the original source head.

One important factor to consider here depends on how we track the `checkpoint_request_id` on the client. If we store a single value per session and clear it in `disconnectAndClear`, we have the potential to reset the sequence if the service record has also been deleted.

> We could theoretically attempt to store a persisted table of sequences for the `user_id`/`client_id`—that would require extracting the `user_id` somehow, which could be more complicated.

We won't store a table of user sequences. Instead, we accept that it is fine for the checkpoint request ID sequence to reset to `1` if both the client and service have no last known state.

### Custom Write Checkpoints

For custom checkpoint requests (previously custom write checkpoints), the general flow is:

- The client generates the next `checkpoint_request_id`.
- The SDK calls a `BackendConnector` method to post the request ID to the application's backend.
- The backend should match the behaviour seen in our `/sync/checkpoint-request` route: only increment its write checkpoint/checkpoint request record if the supplied ID is larger than the stored value, refresh the current request when the supplied ID is equal, and always report the largest stored value (used for hydration in the SDK).

Core storage now accepts an optional `checkpoint_requested_at` value for custom checkpoints. Its presence indicates that the custom write checkpoint record stems from a temporary custom checkpoint request, allowing the compact job to delete the record after the configured retention period. Omitting or clearing this value keeps the existing persistent custom write checkpoint behaviour.

The external module which parses incoming sync events and calls storage has the freedom to decide how and when to pass this parameter. The event-definition specifics are not yet implemented, but we could do either of the following.

Add a new column to indicate checkpoint-request nature (untested):

```diff
event_definitions:
  # Note this event is only supported for customers on [Team and Enterprise](https://www.powersync.com/pricing) plans.
  write_checkpoints:
    payloads:
      # This defines where the replicated custom Write Checkpoints should be extracted from
-      - SELECT user_id, checkpoint, client_id FROM checkpoints
+      - SELECT user_id, checkpoint, client_id, true as is_checkpoint_request FROM checkpoints
```

Or use a new event definition name:

```diff
event_definitions:
  # Note this event is only supported for customers on [Team and Enterprise](https://www.powersync.com/pricing) plans.
-  write_checkpoints:
+  checkpoint_requests:
    payloads:
      # This defines where the replicated custom Write Checkpoints should be extracted from
      - SELECT user_id, checkpoint, client_id FROM checkpoints
```

---

AI Disclosure: The following was implemented by first doing a basic implementation by hand, then various changes were assisted by Claude Opus and Codex 5.5.


rkistner:
Looks good so far!

Some minor issues picked up by Codex listed below. I didn't double-check these, but these seem like plausible issues at a glance.

### 1. checkpoint_request_id accepts negative and unbounded values

The route accepts codecs.bigint directly at `packages/service-core/src/routes/endpoints/
checkpointing.ts:13,` and the codec accepts any signed integer string at libs/lib-services/src/codec/
codecs.ts:53. A first request with -1 can create a negative managed write checkpoint, and a huge
value passes route validation but fails later in Postgres at the ::int8 cast in modules/module-
`postgres-storage/src/storage/checkpoints/PostgresWriteCheckpointAPI.ts:109`. This should be validated
at the API boundary as a positive int64-compatible checkpoint id.

### 2. Performance: Postgres turns every supplied request into a source marker, including already-processed duplicates

`modules/module-postgres-storage/src/storage/checkpoints/PostgresWriteCheckpointAPI.ts:170`
intentionally sets shouldAdvance true for every supplied checkpoint because Postgres storage does not
track processed state. That means stale or duplicate /sync/checkpoint-request calls still execute the
keepalive/logical marker path at `modules/module-postgres/src/api/PostgresRouteAPIAdapter.ts:253`. This
undercuts the PR’s no-op goal and can add avoidable WAL/source writes under reconnect storms or
repeated client retries. Not necessarily a correctness blocker, but it is a real performance
implication to address or explicitly accept.

modules/module-mongodb-storage/src/storage/implementation/MongoWriteCheckpointAPI.ts
```diff
@@ -28,29 +28,35 @@ export class MongoWriteCheckpointAPI implements storage.WriteCheckpointAPI {
 
   async createManagedWriteCheckpoints(
     checkpoints: storage.ManagedWriteCheckpointOptions[]
-  ): Promise<Map<string, bigint>> {
+  ): Promise<storage.CreateManagedWriteCheckpointsResult> {
```
stevensJourney, line 31: One potential issue with this implementation is that managed write checkpoints appear to store only a single checkpoint association per full user/client id. That means a newer checkpoint request can replace the LSN association for an older pending request.

This can delay `write_checkpoint` emission for checkpoint requests. Unlike normal write checkpoints tied to a client-side `target_op` barrier, (ideally) checkpoint requests should not prevent the client from applying incoming changes while waiting for a later source position.

For example:
- replication is lagging
- the client sends checkpoint request `42`, associated with source LSN `A`
- before replication reaches `A`, the same client sends checkpoint request `43`, associated with later source LSN `B`
- storage updates the single record to `43 -> B`
- when replication reaches `A`, there is no longer a stored `42 -> A` association to emit
- the client only receives `43` once replication reaches `B`, so the earlier request is delayed by unrelated later work

rkistner: This should not be an issue for the original use case of blocking sync until the latest checkpoint request is acknowledged: Once the client sends checkpoint request 43, request 42 has no use on the client anymore.

But I guess this changes when we implement explicit checkpoint requests as proposed in https://github.com/orgs/powersync-ja/discussions/324?

stevensJourney: > But I guess this changes when we implement explicit checkpoint requests as proposed in

Yup, exactly correct. If both use cases use the same underlaying `write_checkpoint` record, _then_ the extra delay could occur.

rkistner: I think that's fine for now.

This only has an effect if all of the following is true:
1. The upload queue is empty.
2. There is some replication lag.
3. The client requests checkpoints at a lower interval than the replication lag.

I don't think it's worth changing the storage format to cater for that case right now, but we can consider addressing that when we do future storage changes.

stevensJourney:
> Looks good so far!
> 
> Some minor issues picked up by Codex listed below. I didn't double-check these, but these seem like plausible issues at a glance.
> 
> ### 1. checkpoint_request_id accepts negative and unbounded values
> The route accepts codecs.bigint directly at `packages/service-core/src/routes/endpoints/ checkpointing.ts:13,` and the codec accepts any signed integer string at libs/lib-services/src/codec/ codecs.ts:53. A first request with -1 can create a negative managed write checkpoint, and a huge value passes route validation but fails later in Postgres at the ::int8 cast in modules/module- `postgres-storage/src/storage/checkpoints/PostgresWriteCheckpointAPI.ts:109`. This should be validated at the API boundary as a positive int64-compatible checkpoint id.

I've added validation checks to ensure the supplied checkpoint_request_id is in the safe range.

> ### 2. Performance: Postgres turns every supplied request into a source marker, including already-processed duplicates
> `modules/module-postgres-storage/src/storage/checkpoints/PostgresWriteCheckpointAPI.ts:170` intentionally sets shouldAdvance true for every supplied checkpoint because Postgres storage does not track processed state. That means stale or duplicate /sync/checkpoint-request calls still execute the keepalive/logical marker path at `modules/module-postgres/src/api/PostgresRouteAPIAdapter.ts:253`. This undercuts the PR’s no-op goal and can add avoidable WAL/source writes under reconnect storms or repeated client retries. Not necessarily a correctness blocker, but it is a real performance implication to address or explicitly accept.

The Postgres bucket storage doesn't currently track the `processed_at_lsn`, which makes performing the optimisation for `shouldAdvance` difficult. This optimisation is currently disabled for Postgres bucket storage. I've added a comment in the code for now.

modules/module-mongodb-storage/src/migrations/db/migrations/1782950400001-unique-write-checkpoint-user-id.ts
```diff
@@ -0,0 +1,112 @@
...
+  const {
+    service_context: { configuration }
+  } = context;
+  const db = storage.createPowerSyncMongo(configuration.storage as MongoStorageConfig);
+
+  try {
+    // Existing databases may already have duplicate rows because the old
+    // user_id index was not unique and checkpoint writes use upserts. Clean
+    // those duplicates before creating the unique replacement index.
+    await deduplicateWriteCheckpoints(db.write_checkpoints);
```
stevensJourney: This was picked up by Codex, and seems to be a legitimate potential issue. We previously didn't have any uniqueness index based off the `user_id`. I don't think it was intentional to ever have multiple records for this. I believe it wasn't an issue before since we'd just query a write checkpoint based off the corresponding LSN.
In the new model, we'd like to track a fixed reference per `user_id`.

rkistner: This migration looks risky:
1. The aggregation query can be slow and need to use a lot of db memory if there are many existing write checkpoints on the instance.
2. In theory, new duplicates can be introduced between performing the de-duplication and creating the unique index.
3. There is a period between dropping the old index and creating the new index, in which checkpoints queries will not have any per-user index to use.

In practice it could be okay:
1. I'm not aware of any current case of more than 100k or so write checkpoints in a single instance, but I don't have any data on self-hosted instances.
2. In practice, duplicates should be _very_ rare, so new ones should not be introduced in this period.
3. You can work around this by using a different name for the new index, and creating the new index before dropping the old one. A caveat is that it needs an unique set of fields for each index - you can't create a second one on just `{user_id: 1}`.

modules/module-mongodb-storage/src/migrations/db/migrations/1782950400000-checkpoint-requested-at-index.ts
```diff
@@ -0,0 +1,55 @@
...
+        checkpoint_requested_at: 1
+      },
+      {
+        name: 'checkpoint_requested_at',
+        // Only client-requested checkpoints have this field; generated
+        // checkpoints leave it unset. This keeps the index limited to the
+        // documents the compact job's retention delete scans.
+        partialFilterExpression: { checkpoint_requested_at: { $exists: true } }
+      }
+    );
```
rkistner, line 23: Instead of a normal index, you can create a TTL index, that will automatically remove them. That would avoid the need for having a separate compact task. But that may make it more difficult to keep the retention period configurable, so not sure if this is worth it.

packages/types/src/config/PowerSyncConfig.ts
```diff
@@ -409,6 +409,16 @@ export const powerSyncConfig = t
...
 
+            checkpoint_request_retention_days: t.number
+              .meta({
+                description: dedent`
+              Number of days to keep client-requested write checkpoint records.
+              Expired records are removed by the compact job.
+              Must be a positive integer. Default of 30.
+            `
+              })
+              .optional(),
```
rkistner: I wonder if days is the best unit here - I imagine we could remove records in as little as 15-60 minutes?

rkistner (commented):
I'm happy with this for the most part - the only potential issue I could see is the migration for unique user_id.

I'm wondering, what is the effect if there are duplicates? Is it more significant from before this PR?

In practice, since we always use `findOneAndUpdate` on a user_id, duplicates should be very rare. So I'm wondering if we should do that migration as a separate project, potentially as part of a storage version change.

stevensJourney:
> I'm wondering, what is the effect if there are duplicates? Is it more significant from before this PR?

I think this was less of an issue for the old generated write checkpoint flow. Even if duplicate rows existed, the client used the client_id returned from the row that was just updated, so as long as that row had the correct lsns, the acknowledgement could still work.

For the new client-supplied flow, duplicates are more awkward because storage needs to answer “what is the latest monotonic request id for this full user id?” If we query/seed from the wrong duplicate row, we could return a lower checkpoint request id than another row already has, which would put the client in a worse state.
In practice, duplicate creation should still be unlikely because SDKs don’t make concurrent initial requests for the same full user id: they make one initial request, then later requests only after initialization/state is established. The unique index is mainly a guardrail to enforce the one-row-per-full-user-id invariant.

> In practice, since we always use findOneAndUpdate on a user_id, duplicates should be very rare. So I'm wondering if we should do that migration as a separate project, potentially as part of a storage version change.

I think there is a very slight risk that duplicate records could exist, but, I think it's so theoretical - that we probably could push this migration out for later.

packages/types/src/config/PowerSyncConfig.ts
stevensJourney: The thought process here was that it depends on the replication lag. For the write-checkpoint case, we want to make sure the server actually sends the corresponding `write-checkpoint` back which the client is waiting for. 

For a basic case: if the lag is larger than the retention period, we might miss those events.

I guess this could be solved if the client was constantly retrying the last checkpoint request though. I was wondering what a good reason would be for retrying those requests, this might be one of them.

modules/module-mongodb-storage/src/storage/implementation/MongoWriteCheckpointAPI.ts
```diff
@@ -52,36 +57,128 @@ export class MongoWriteCheckpointAPI implements storage.WriteCheckpointAPI {
...
+        // every duplicate row for the user_id that is lower than this request id.
+        const shouldApplyRequestId = {
+          $or: [
+            { $eq: [{ $ifNull: ['$client_id', null] }, null] },
+            { $lt: ['$client_id', { $literal: checkpointRequestId }] }
+          ]
+        };
+
+        return {
+          updateMany: {
```
stevensJourney, line 157: The change here should cater for the very rare possibility that we might have duplicate `write_checkpoint` records for a full `user_id`.

Instead of doing a migration to deduplicate, this now will handle those cases as part of the `sync/checkpoint-request` flow.

If the client doesn't have the latest client_id/next_checkpoint_request_id sequence. It will make a checkpoint request with the value of `1`. Which will funnel through to this logic which will attempt to update _all_ records for the user_id - skipping updates if the currently stored value is larger than the supplied value.

We then query the write checkpoint records for the user and return the MAX `client_id` for that user (taking all potential duplicates into account).

The client is then free to start its `next_checkpoint_request_id` at that MAX value. Which will prevent any missed updates in future. 

The client's next request will then update all the user's `write_checkpoint` records, which will mark them as requested - which means we could automatically clean these up at a later stage.

modules/module-mongodb-storage/src/migrations/db/migrations/1782950400001-unique-write-checkpoint-user-id.ts
stevensJourney: I've worked around possible duplicate and removed the need for this migration. For context, see https://github.com/powersync-ja/powersync-service/pull/696#discussion_r3542782899

packages/types/src/config/PowerSyncConfig.ts
stevensJourney: I've implemented retries on the client side now. It feels safe to have a shorter retention period now. I've updated the default to 60 minutes.

modules/module-mongodb-storage/src/migrations/db/migrations/1782950400000-checkpoint-requested-at-index.ts
stevensJourney: That is pretty cool! It feels like, at least for the first release- having the ability to configure the retention period is a nice to have. At this stage, I'd like to keep it configurable as-is.

rkistner (approved):

PR #697, opened 2026-06-29
[MongoDB Storage] Read from secondaries

rkistner:
This adds a `read_preference` option for MongoDB Storage:

```yaml
storage:
  type: mongodb
  uri: ...
  bulk_read_preference: secondaryPreferred # primary | primaryPreferred | secondary | secondaryPreferred | nearest
```

This read preference is only used for:
1. Bucket data range queries and checksums via the sync API.
2. Only for non-incremental updates.

Specifically, when a request reads data for bucket not seen before _in that request_, it uses the specified read preference. Any checkpoint_diff after that would use the default _primary_ read preference, unless a new bucket is added in the checkpoint_diff.

Additionally, when this option is specified, this uses `readConcern: 'majority'`. That should have a minimal impact.

This allows using `secondaryPreferred` or `nearest` to spread the load of bulk reads over the storage cluster's nodes. We use causal consistency (for data) or snapshot consistency (for checkpoints) using the `operationTime` of the checkpoint we read, which ensures that we don't get stale data from a secondary node, at the cost of potentially needing to wait if that secondary is behind.

Effectively, MongoDB picks a node to read from for each _operation_:
1. For `secondaryPreferred`, it would pick a random secondary if available, otherwise use the primary.
2. For `nearest`, it would pick any random node.

The "nearest" part only has an effect when the nodes are distributed enough to have more than a 15ms latency difference, which you typically shouldn't see within a single region. For multi-cloud or multi-region setups, that may become relevant, and the [MongoDB docs](https://www.mongodb.com/docs/manual/core/read-preference-mechanics/) should be consulted for details.

I don't have a good use case for `primaryPreferred` - it is only added for completeness. If the primary is unavailable, other reads in the sync path will fail, so this will not help for redundancy.

`secondary` can be used as a stronger signal to not use the primary node for these reads.

The default for now is roughly equivalent to `primary`, but without the changes to `readConcern`.

## Latency impact

Right now, the loop to check for new checkpoints is using snapshot reads, which imply `readConcern: 'majority'`. This means any read change must be written to at least 2 out of 3 nodes in a typical 3-node replicaset before it is visible. At that point, in a healthy cluster, the secondary we read from is very likely to be up to date, in which case there would be no added latency.

To ensure minimal latency for incremental streaming updates, we still read from the primary for most cases where we sync a `checkpoint_diff`.

There is one case where there could be a significant performance impact: If only one of the two secondaries is degraded and falls behind, and reads hitting that secondary would be blocked. Once it is behind more than 90s, it will be excluded from reads, so this primarily affects cases where a node has a small but consistent replication lag.

## Performance impact

Since bulk reads are spread randomly across nodes, the effectiveness of MongoDB's cache is not increased: Each node would still need to keep the same amount of data in the cache to use it effectively. Where this can help is cases where:
1. Many operations miss the cache, in which case the disk read load can be spread across nodes.
2. Operations result in high CPU usage, in which case the CPU usage can be spread across nodes.

Note that if `secondaryPreferred` is used and one node goes down for maintenance, reads would still hit only one of the nodes. `nearest` may be better for that reason.

## Experimental

This option is considered experimental until we've done more thorough testing on the performance impact. We may remove it again in a future version.

## Testing

Tested using a local replicaset running in Docker.

MongoDB has an option `secondaryDelaySecs` to configure replication delay. Unfortunately this forces the node to be marked as "hidden", meaning it is never used by the client, so it doesn't help for testing reads.

What does work is to pause & resume replication in a loop - run this on secondaries in mongosh:

```js
while (true) {
  db.adminCommand({ configureFailPoint: "rsSyncApplyStop", mode: "alwaysOn" });
  print(new Date().toISOString(), "replication paused");

  sleep(10000);

  db.adminCommand({ configureFailPoint: "rsSyncApplyStop", mode: "off" });
  print(new Date().toISOString(), "replication resumed");

  sleep(500);
}
```

Run the above on one or both of the secondaries to see the behavior. The biggest visible impact for me was when running that on only _one_ of the secondaries, simulating the case of one degraded secondary, described in "latency impact" above.

## AI usage

Manually planned the options and consistency properties; Used Codex gpt-5.5 to implement and Claude opus 4.8 to review.


modules/module-mongodb-storage/src/storage/implementation/v3/MongoChecksumsV3.ts
```diff
@@ -169,8 +181,7 @@ export class MongoChecksumsV3 extends MongoChecksums implements DefinitionChecks
     const pipeline = this.buildPartialChecksumPipeline(requests);
     const aggregate = await collection
       .aggregate(pipeline, {
-        session: undefined,
-        readConcern: 'snapshot',
+        ...this.checksumReadOptions(context),
```
stevensJourney: I was chatting to Codex about the PR to get a better understanding. While asking it a few questions, it flagged this, seems like it could be worth something:

This drops `readConcern: 'snapshot'` from the checksum aggregation, and on the default path (no read_preference set) `checksumReadOptions` returns `{}`, so existing deployments fall back to `local` — not just the new secondary-read path.

The snapshot was there so the aggregate sees a consistent point-in-time view; without it, a long-running aggregate can observe a mixed pre/post-compaction (CLEAR) view and cache an inconsistent checksum. Note majority doesn't substitute for this — it guarantees majority-committed reads, not per-operation snapshot isolation.

Is the change intentional, and how does it interact with concurrent compaction? The description mentions the move to majority but not this part.

rkistner: This is indeed an undocumented change, but should be safe:
1. During normal replication, we only append changes, and it would only be _after_ the checkpoint queried for here.
2. During compacting, every atomic (transactional) operation preserves checksums for the affected operations.

rkistner: After checking again, I think you're right: Doing a non-snapshot read on large ranges of data can in theory read data from both before and after a compact operation. Even if that compact operation is atomic, this can lead to inconsistencies. I can't tell whether this actually happens in practice, but MongoDB doesn't appear to give guarantees against this unless you use snapshot reads.

I changed this to use snapshot reads again, but will have to review again thoroughly to make sure I didn't miss something.

rkistner: After lots of back and forth, I confirmed with some tests:
https://gist.github.com/rkistner/723c6c2e8f7b267070243c723e1d32d4

The above is a simplified scenario of what we have: A single transaction updates multiple documents, changing the checksum of individual ones, while keeping the checksum total the same. This all happens atomically: There is never a db state where the checksum total is different.

However, the big point that I missed, is that a normal db read is _not_ atomic. This is quite different from what you expect with modern SQL databases. Essentially, MongoDB reads a batch at a time, and may apply updates in between those batches.

There are two ways around that:
1. Use snapshot reads.
2. Use a transaction.

Each of those use a single timestamp for the entire read, which avoids this issue.

modules/module-mongodb-storage/src/storage/implementation/v3/MongoChecksumsV3.ts
```diff
@@ -169,8 +181,7 @@ export class MongoChecksumsV3 extends MongoChecksums implements DefinitionChecks
...
         maxTimeMS: lib_mongo.MONGO_CHECKSUM_TIMEOUT_MS
```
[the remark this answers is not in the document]
rkistner: Fixed.

stevensJourney (approved):
Happy with the changes :)

PR #701, opened 2026-06-30
Log checkpoint_complete / Fix checkpoint invalidation bug

rkistner:
This adds additional logs to help debug sync issues and performance.

## Example Scenarios

Scenario 1: A user says sync is not working - they're not seeing the latest data. We're trying to debug from service logs. Logs show no errors, shows there was a checkpoint was sent, and an eventual "Sync stream complete". But we have no further visibility: Did the client finish syncing? Were they still downloading data?

Scenario 2: A user says there is a significant delay in getting new data, and we're trying to find the source of the delay. There is no replication lag. Storage database shows high load, but we can't tell whether that is an issue. Logs show checkpoints are sent, but no timing info. We see some `Sync concurrency limit reached, waiting for lock` messages, which give some indication that there could be high load, but there is nothing definitive.

## New logs

This add logs for:
1. checkpoint_complete
2. partial_checkpoint_complete
3. checkpoint_invalidated

With these, we include stats on operations and bytes synced, as well as timing info. This helps to get better visibility into what is happening while syncing, instead of only at the end of the sync connection.

This then removes logs for `Got sync lock`, `Releasing sync lock` and `Sync concurrency limit reached, waiting for lock`. These were extremely noisy, are not actionable, and introduce confusion. The new logs above indicate aggregate timings that indicate (1) how long the connection was waiting for locks, and (2) how long the connection was holding a lock (implicitly via `bucket_data` and `client` timings), which is much easier to interpret than the individual sync lock messages.

In some cases, these changes could lead to doubling log volumes. But there are also many cases where the sync lock logs were a lot more noisy than the checkpoint logs, so removing those compensates for the increased volumes somewhat. I've seen cases where 86% of log lines are "sync lock" lines, with only 12% being checkpoint-related. In cases like that, even with the checkpoint lines being doubled, the total log lines would still be around 25% of what it was before.

### Future potential: Reducing log volumes

If the log volumes are still too high, we can consider merging the "new/updated checkpoint" logs into these new logs, instead of effectively logging two messages for each checkpoint. We have much more useful timing and data volume info at the time of "checkpoint_complete" than at the time we send the checkpoint. However, there are many potential edge cases where that change can result in reduced visibility. For example, if a client takes 10 minutes to download the data, we won't get any info on what is happening until the end.

One possibility may be to _delay_ the "new/updated checkpoint" messages, and merge the two messages when they're close together, but still keep the current split if it takes longer than a couple of seconds to complete the checkpoint. That would reduce noise if there are many tiny checkpoint updates, while preserving the useful split when large volumes of data is synced.

## Fixed bug: missing data on checkpoint invalidation with priorities

There was one rare consistency bug that is fixed as a side-effect from this. This scenario could happen:
1. Bucket priorities are used.
2. Client syncs a high-priority bucket.
3. Due to a concurrent compaction, the checkpoint is invalidated (via target_op).
4. The service picks up that invalidation, and the batch stops, without emitting a "partial_checkpoint_complete". The intention is continuing with the next checkpoint.
5. However, the loop handling priorities does not see that invalidation, it only sees "bucket data for this priority is done".
6. The priority loop incorrectly continues with the next priority.
7. The service emits the data, as well as a final checkpoint_complete.
8. The client now received a checkpoint_complete without the full data for that checkpoint. It would pick up the missing data in the checksum check, requiring a full re-download of the affected buckets.

This bug is now fixed by properly aborting the priority loop if the checkpoint is invalidated.

## Example logs

_Some of these logs are from tests, since it's tricky to produce the exact scenario in practice._

Example timings under contention (many concurrent bulk sync requests):
```js
info: checkpoint_complete: 4125837 {"bson":false,"checkpoint":4125837,"client_id":"load-test-20","data_synced_bytes":90432986,"ms":{"acquiring_lock":7025,"bucket_data":2989,"checksum":62,"other":1,"sending":751,"total":10828},"operations_synced":5984,"rid":"h/019f1cab-32c6-7758-9a34-800dade4fa7e","route":"/sync/stream","user_id":"3"}
```

Example for bucket priorities:

```js
info: New checkpoint: 2 | write: null | buckets: 2 | param_results: 0 ["b0.1.1[]","b1.1.2[]"] {"buckets":2,"checkpoint":2,"parameter_query_results":0,"user_id":"test"}
info: partial_checkpoint_complete: 2 {"checkpoint":2,"data_synced_bytes":234,"operations_synced":1,"priority":1,"user_id":"test"}
info: checkpoint_complete: 2 {"checkpoint":2,"data_synced_bytes":123,"ms":{"bucket_data":1,"checksum":4,"total":6},"operations_synced":1,"user_id":"test"}
```

Checkpoint interrupted with higher-priority data:

```js
info: New checkpoint: 10001 | write: null | buckets: 2 | param_results: 0 ["b0.1.1[]","b1.1.2[]"] {"buckets":2,"checkpoint":10001,"parameter_query_results":0,"user_id":""}
info: partial_checkpoint_complete: 10001 {"checkpoint":10001,"data_synced_bytes":0,"operations_synced":1,"priority":1,"user_id":""}
info: checkpoint_invalidated: 10001 {"checkpoint":10001,"data_synced_bytes":0,"ms":{"bucket_data":29,"checksum":6,"sending":7,"total":43},"operations_synced":3000,"reason":"checkpoint_superseded","user_id":""}
info: Updated checkpoint: 10002 | write: null | buckets: 2 | param_results: 0 | updated: ["b1.1.2[]"] | removed: [] {"buckets":2,"checkpoint":10002,"parameter_query_results":0,"removed":0,"updated":1,"user_id":""}
info: partial_checkpoint_complete: 10002 {"checkpoint":10002,"data_synced_bytes":0,"operations_synced":1,"priority":1,"user_id":""}
info: checkpoint_complete: 10002 {"checkpoint":10002,"data_synced_bytes":0,"ms":{"bucket_data":43,"checksum":6,"other":7,"total":57},"operations_synced":7000,"user_id":""}
```

Connection closed by client before completing checkpoint:

```js
info: Sync stream started {"bson":false,"client_id":"load-test-0","rid":"h/019f1ca7-4656-723a-bf3d-084bc26f96bd","route":"/sync/stream","user_id":"3"}
info: New checkpoint: 4125837 | write: null | buckets: 5 | param_results: 0 ["4#global2|0[]","4#by_user1|0[3]","4#by_user3|0[3]","4#location|0[3]","4#synth|0[3]"] {"bson":false,"buckets":5,"checkpoint":4125837,"client_id":"load-test-0","parameter_query_results":0,"rid":"h/019f1ca7-4656-723a-bf3d-084bc26f96bd","route":"/sync/stream","user_id":"3"}
info: Sync stream complete {"bson":false,"client_id":"load-test-0","close_reason":"client closing stream","data_sent_bytes":53137612,"data_synced_bytes":53137612,"large_buckets":{"4#location|0[3]":5535},"operation_counts":{"clear":0,"move":0,"put":5798,"remove":0},"operations_synced":5798,"rid":"h/019f1ca7-4656-723a-bf3d-084bc26f96bd","route":"/sync/stream","stream_ms":653,"user_id":"3"}
info: POST /sync/stream {"bson":false,"client_id":"load-test-0","duration_ms":654,"method":"POST","path":"/sync/stream","rid":"h/019f1ca7-4656-723a-bf3d-084bc26f96bd","route":"/sync/stream","status":200,"user_id":"3"}
info: checkpoint_invalidated: 4125837 {"bson":false,"checkpoint":4125837,"client_id":"load-test-0","data_synced_bytes":53137612,"ms":{"bucket_data":530,"sending":129,"total":660},"operations_synced":5798,"reason":"checkpoint_cancelled","rid":"h/019f1ca7-4656-723a-bf3d-084bc26f96bd","route":"/sync/stream","user_id":"3"}
```

## AI Usage

Manually designed the new behavior and tracking. Used Codex gpt-5.5 for some of the implementation.

### DIFF
--- a/.changeset/cyan-readers-bathe.md
+++ b/.changeset/cyan-readers-bathe.md
@@ -1,0 +1,6 @@
+---
+'@powersync/service-core-tests': patch
+'@powersync/service-core': patch
+---
+
+Add logs for checkpoint_complete to assist with debugging. Remove "sync lock" logs.
--- a/.changeset/modern-socks-pull.md
+++ b/.changeset/modern-socks-pull.md
@@ -1,0 +1,6 @@
+---
+'@powersync/service-core-tests': patch
+'@powersync/service-core': patch
+---
+
+Fix edge case where the service can return incomplete data. This could happen when using bucket priorities, and a checkpoint is invalidated by the compact process while syncing.
--- a/packages/service-core-tests/src/tests/register-sync-tests.ts
+++ b/packages/service-core-tests/src/tests/register-sync-tests.ts
@@ -1238,6 +1238,150 @@ bucket_definitions:
     });
   });
 
+  test('compacting high-priority data invalidates checkpoint before lower priorities', async (context) => {
+    // Scenario:
+    // 1. Bucket priorities are used.
+    // 2. Client syncs a high-priority bucket.
+    // 3. Due to a concurrent compaction, the checkpoint is invalidated (via target_op).
+    // 4. The service picks up that invalidation, and the batch stops, without emitting a "partial_checkpoint_complete". The intention is continuing with the next checkpoint.
+    // 5. However, the loop handling priorities does not see that invalidation, it only sees "bucket data for this priority is done".
+    // 6. The priority loop incorrectly continues with the next priority.
+    // 7. The service emits the data, as well as a final checkpoint_complete.
+    // 8. The client now received a checkpoint_complete without the full data for that checkpoint. It would pick up the missing data in the checksum check, requiring a full re-download of the affected buckets.
+    await using f = await factory();
+
+    const syncRules = await updateSyncRules(f, {
+      content: `
+bucket_definitions:
+  high_priority:
+    priority: 1
+    data:
+      - SELECT * FROM test WHERE substring(id, 1, 4) = 'high';
+  low_priority:
+    priority: 2
+    data:
+      - SELECT * FROM test WHERE id = 'low';
+    `
+    });
+
+    const bucketStorage = await f.getInstance(syncRules);
+    await using writer = await bucketStorage.createWriter(test_utils.BATCH_OPTIONS);
+    const testTable = await test_utils.resolveTestTable(writer, 'test', ['id'], config);
+
+    await writer.markAllSnapshotDone('0/1');
+    await writer.save({
+      sourceTable: testTable,
+      tag: storage.SaveOperationTag.INSERT,
+      after: {
+        id: 'high1',
+        description: 'High priority 1'
+      },
+      afterReplicaId: 'high1'
+    });
+    await writer.save({
+      sourceTable: testTable,
+      tag: storage.SaveOperationTag.INSERT,
+      after: {
+        id: 'high2',
+        description: 'High priority 2'
+      },
+      afterReplicaId: 'high2'
+    });
+    await writer.save({
+      sourceTable: testTable,
+      tag: storage.SaveOperationTag.INSERT,
+      after: {
+        id: 'low',
+        description: 'Low priority'
+      },
+      afterReplicaId: 'low'
+    });
+    await writer.commit('0/1');
+
+    const stream = sync.streamResponse({
+      syncContext,
+      bucketStorage,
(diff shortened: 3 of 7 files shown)


stevensJourney (approved):
This looks very nice. The new logs should be very helpful.

PR #703, opened 2026-07-01
feat(sync-rules): wildcard schema support for Sync Streams

henriquekraemer:
Opened at the PowerSync team's request following a positive review of the approach (support request #40221).

### Summary

Adds support for a **wildcard schema** in a Sync Stream's `FROM` clause, e.g.:

```sql
SELECT * FROM "%".assets
WHERE assets.schema() = auth.parameter('tenant_schema')
```

One stream definition can then span every matching Postgres schema, with the active schema resolved per client. This enables **schema-per-tenant** replication (a single database with one schema per tenant, e.g. the Rails `ros-apartment` gem) from a single replication slot, with no per-tenant sync-rules change or redeploy.

### How it works

`schema()` and `table_suffix()` calls are qualified by a table in scope (a name or alias, resolved through the same scope lookup as column references, so joins and aliases behave as expected: `SELECT * FROM "users%" AS users WHERE users.table_suffix() = ...`). They compile to a new `RowMetadata` expression input (sharing a superclass with `ColumnInRow`, so dependency analysis is unchanged). Sync plans get a matching `RowMetadataSqlValue` alternative in `TableProcessorData`, and the evaluators resolve those inputs against the source table of the row being processed, for both bucket data sources and parameter index lookup creators. Since these are function calls rather than column references, they cannot shadow real columns and need no special handling in schema validation.

### Changes

- **`service-sync-rules`**
  - `RowMetadata` input in the compiler, parsed from the `call` case in `PostgresToSqlite`, with diagnostics for unknown qualified functions, unexpected arguments, unresolved table qualifiers, and `table_suffix()` on non-wildcard tables.
  - `RowMetadataSqlValue` alternative in `TableProcessorData`, serialized as-is in sync plans (no version gating for now, as discussed).
  - Evaluators resolve metadata inputs from the source table, in `PreparedStreamBucketDataSource` and `PreparedParameterIndexLookupCreator`.
  - `TablePattern`: schema-wildcard matching (`isSchemaWildcard` / `schemaPrefix`).
- **`service-module-postgres`**
  - `getQualifiedTableNames` discovers tables across all matching schemas for a wildcard-schema pattern (excluding internal schemas), using each table's actual schema for the publication check and the source-table descriptor.

### Behavior note

`TablePattern.matches()` is shared with the older sync rules editions. Previously a `"%"` schema only matched a schema literally named `%` (i.e. nothing in practice); it now matches any schema. The metadata functions themselves are only available in the edition-3 compiler.

### Tests

- `sync-rules`: compiler diagnostics, per-schema bucketing, static bucket resolution from the request, metadata as selected columns, and parameter lookups with metadata (the compiler test helper also round-trips plans through serialization).
- `module-postgres` integration (`schema_per_tenant.test.ts`, real Postgres + MongoDB storage via `describeWithStorage`): cross-schema table discovery and per-schema bucket routing.

Changeset included (`minor` for both packages).


packages/sync-rules/src/streams/from_sql.ts
```diff
@@ -524,6 +524,17 @@ class SyncStreamCompiler {
   }
 }
 
+function syntheticWildcardColumns(sourceTable: TablePattern): ColumnDefinition[] {
```
simolus3: Note that `streams/` is an older / legacy implementation of Sync Streams that isn't used with `edition: 3`. You'd want to update `compiler/`.

(we should rename it, the current state is confusing)

packages/sync-rules/src/sync_plan/evaluator/bucket_data_source.ts
```diff
@@ -150,10 +151,13 @@ class PendingStreamDataSource {
 
   instantiate(engine: ScalarExpressionEngine) {
     const evaluator = engine.prepareEvaluator(this.statement);
+    const pattern = this.tablePattern;
 
     return (options: EvaluateRowOptions, results: UnscopedEvaluationResult[]) => {
       try {
-        const inputInstantiation = this.evaluatorInputs.map((input) => options.record[input.column]);
+        // Synthetic columns feed filters/bucket parameters via inputRecord; `star` reads the original record (never synced).
+        const inputRecord = addSpecialColumns(pattern, options.sourceTable, options.record);
```
simolus3: We might need something similar for `PreparedParameterIndexLookupCreator`.

I'm not sure if exposing `_table_suffix` and `_schema` as regular columns is the best way for this, because:

- They would shadow existing columns with the same name.
- We have a validator matching referenced columns against the actual schema in the source database, those tools now need to be aware of this.

Even though it's more effort, I think it might make sense to be explicit about when a reference is resolved to a table/schema name:

- In `src/compiler/expression.ts`, we should have a third type of `ExpressionInput` like `RowMetadata` or something to represent these two special cases. It can share a superclass with `ColumnInRow` since the dependency analysis for both cases is the same.
- In sync plans (`src/sync_plan/plan.ts`), these expressions should also have explicit representations. We could add a new alternative to `TableProcessorData` (representing external data for bucket data sources and parameter index creators).
- We'd then change `evaluatorInputs` to be aware of these references, and resolve them against the source table we already have in `TableRow` as an input to the prepared evaluator.

packages/sync-rules/src/streams/from_sql.ts
henriquekraemer: Thanks, good catch. I went for `streams/` because that's where the existing `_table_suffix` handling lives, and didn't realize it was the legacy path. Do you want me to drop the legacy changes entirely and make this edition-3 only? That would also shrink the diff.

simolus3: Yeah, there's no reason to update the old code IMO.

packages/sync-rules/src/sync_plan/evaluator/bucket_data_source.ts
henriquekraemer: Makes sense. The synthetic column approach was the least invasive way I found, but I agree the shadowing and validator issues make it fragile.

I'll rework it the way you described: a `RowMetadata` input type in `compiler/expression.ts` sharing a superclass with `ColumnInRow`, an explicit alternative in `TableProcessorData`, and the evaluators resolving it from the source table on `TableRow` instead of injecting values into the record. I'll cover `PreparedParameterIndexLookupCreator` as well, since `_schema` in parameter queries doesn't work at all today.

Two questions before I start:

1. If a table actually has a column named `_schema` or `_table_suffix`, what should win? My plan was to resolve to metadata on wildcard sources and treat it as a normal column reference otherwise, but a compile-time error is also an option if you prefer something stricter.
2. Serialized sync plans look version-agnostic today. Is "a plan using the new input type requires a newer service version" acceptable, or should this be gated somehow?

Moving the PR to draft while I work on this.

simolus3: > If a table actually has a column named `_schema` or `_table_suffix`, what should win?

What if we change the syntax to function calls, e.g. `table.schema()` and `table.table_suffix()`? We could parse those via the `call` case in `PostgresToSqlite`, allowing us to handle them separately from columns with the same name.

> Serialized sync plans look version-agnostic today. Is "a plan using the new input type requires a newer service version" acceptable, or should this be gated somehow?

I think we can treat it as acceptable. Changing the sync plan format needs to be done carefully because existing deployments need to keep working, but adding new syntax doesn't affect that.

We will need gates to check for this, probably by bumping the plan version to v2 and allowing both versions when deserializing them (the rest of the deserialization logic wouldn't need changes since this only adds things, we just need older service versions to reject these plans). Ideally we can keep emitting a v1 version if this feature isn't used.

But don't worry about that initially, I think it's something we can figure out later (we don't have established procedures for this yet, so coming up with something here is on us).

henriquekraemer: Sounds good. I'll go with the `table.schema()`/` table.table_suffix()` function syntax and leave plan versioning out for now.

henriquekraemer: @simolus3 `table` is a reserved keyword in pgsql-ast-parser, so the bare form `table.schema()` doesn't parse ("Unexpected kw_table token"). The quoted form `"table".schema()` works. The namespace is checked in a single place, so it's trivial to change. If requiring quotes feels awkward, alternatives that parse unquoted include `source.schema()`, `row.schema()` and `meta.schema()`. Happy to use to whichever you prefer.

packages/sync-rules/src/streams/from_sql.ts
henriquekraemer: Done, dropped the legacy `streams/` changes entirely and reworked this in `compiler/` only (714d66c). The latest merge also picks up your `legacy/` rename from #707.

packages/sync-rules/src/sync_plan/evaluator/bucket_data_source.ts
simolus3: Sorry, I meant to use the actual identifier of the table in scope. E.g. `SELECT * FROM "users%" AS users WHERE users.table_suffix() = ...`. That form is also explicit about the table from which the suffix should be extracted when there are multiple tables in scope (e.g. for a join).

henriquekraemer: The qualifier now resolves through the same scope lookup as column references, so aliases and multiple tables in scope behave as you'd expect. Tests updated accordingly.

packages/sync-rules/src/sync_plan/evaluator/bucket_data_source.ts
```diff
@@ -191,3 +194,16 @@ class PendingStreamDataSource {
     };
   }
 }
+
+export function resolveRowMetadata(
```
simolus3: Nit: Maybe move this to `table_processor_to_sql.ts` since it's shared between buckets and parameters.

simolus3 (commented):
The changes to the Sync Stream compiler and plan look great to me.

Since the plan changes are purely additive, I'm not sure if we need to be concerned about backwards compatibility. The only thing that could go wrong is if a user upgrades to a service version with `.schema()` and `.table_suffix()` support, deploys Sync Streams and later downgrades again, they'd get confusing errors because the old version doesn't understand the serialized plan.

Maybe we're fine with this, if not we'd have to bump the plan version to `2` which would make older service versions fail early and with a more recognizable error message. I can open a follow-up PR to update that, it would be good to hear thoughts from @rkistner on this too.

I also haven't checked whether this might need changes for other source databases too.

packages/sync-rules/src/sync_plan/evaluator/bucket_data_source.ts
henriquekraemer: Thanks! I moved `resolveRowMetadata` to `table_processor_to_sql.ts` as suggested. On plan versioning: happy to defer to you and @rkistner since it affects the serialized format generally, not just this feature.

stevensJourney:
On plan compatibility: I don't have a strong opinion on whether this needs a plan version bump. 

On other source databases: I checked this briefly and I think this is currently Postgres-only at the source discovery layer. The shared sync-rules/compiler changes are generic, but MongoDB, MySQL, MSSQL, and Convex still appear to treat `tablePattern.schema` as an exact database/schema/default schema. I think we should add an explicit guardrail there. We could log a clear error or throw a hard error when `tablePattern.isSchemaWildcard` is used for those modules; I’d lean toward throwing so the config does not look valid while the module silently discovers nothing.

henriquekraemer:
> On plan compatibility: I don't have a strong opinion on whether this needs a plan version bump.
> 
> On other source databases: I checked this briefly and I think this is currently Postgres-only at the source discovery layer. The shared sync-rules/compiler changes are generic, but MongoDB, MySQL, MSSQL, and Convex still appear to treat `tablePattern.schema` as an exact database/schema/default schema. I think we should add an explicit guardrail there. We could log a clear error or throw a hard error when `tablePattern.isSchemaWildcard` is used for those modules; I’d lean toward throwing so the config does not look valid while the module silently discovers nothing.

Something like this? Adding the check at each module's table pattern resolution point. For MongoDB, in `MongoSnapshotter.resolveQualifiedTableNames`:

```ts
private async resolveQualifiedTableNames(
  batch: storage.BucketStorageBatch,
  tablePattern: TablePattern
): Promise<storage.SourceTable[]> {
  if (tablePattern.connectionTag != this.connections.connectionTag) {
    return [];
  }

  if (tablePattern.isSchemaWildcard) {
    throw new ServiceError(
      ErrorCode.PSYNC_Sxxxx,
      'Schema wildcards ("%") in table patterns are not supported for MongoDB connections.'
    );
  }

  const schema = tablePattern.schema;
  // ...
}
```

Same check in `getTablesFromPattern` for MySQL (`common/schema-utils.ts`) and MSSQL (`utils/schema.ts`), and in the Convex adapter's pattern loop. I'd register a proper error code for it instead of the placeholder above.

One question: do you want this surfaced at validation time too (route API adapters), so the error shows up when deploying the sync config rather than only when replication starts? Happy to add it in both places.

stevensJourney:
> Same check in getTablesFromPattern for MySQL (common/schema-utils.ts) and MSSQL (utils/schema.ts), and in the Convex adapter's pattern loop. I'd register a proper error code for it instead of the placeholder above.

This sounds good to me! From looking at the `codes.ts` file. I think this range might be applicable

```typescript
  // ## PSYNC_R22xx: SQL supported feature issues

  // OR Maybe even

  // ## PSYNC_R23xx: SQL schema mismatch issues
```

> One question: do you want this surfaced at validation time too (route API adapters), so the error shows up when deploying the sync config rather than only when replication starts? Happy to add it in both places.

This sounds like it would be very nice to have. I'm not sure how much effort this would entail though. I think throwing an error would be fine for the first release.

henriquekraemer:
> > Same check in getTablesFromPattern for MySQL (common/schema-utils.ts) and MSSQL (utils/schema.ts), and in the Convex adapter's pattern loop. I'd register a proper error code for it instead of the placeholder above.
> 
> This sounds good to me! From looking at the `codes.ts` file. I think this range might be applicable
> 
> ```ts
>   // ## PSYNC_R22xx: SQL supported feature issues
> 
>   // OR Maybe even
> 
>   // ## PSYNC_R23xx: SQL schema mismatch issues
> ```
> 
> > One question: do you want this surfaced at validation time too (route API adapters), so the error shows up when deploying the sync config rather than only when replication starts? Happy to add it in both places.
> 
> This sounds like it would be very nice to have. I'm not sure how much effort this would entail though. I think throwing an error would be fine for the first release.

Done in 6924d32. Registered `PSYNC_R2201` in the `PSYNC_R22xx` range and added the hard error at each module's table pattern resolution point: `MongoSnapshotter.resolveQualifiedTableNames` (MongoDB), `getTablesFromPattern` (MySQL and SQL Server) and `ConvexStream.resolveQualifiedTableNames` (Convex). The check runs after the `connectionTag` filter where there is one, so patterns belonging to other connections are unaffected. I'll leave the validation-time surfacing for a follow-up as suggested.

simolus3:
> Maybe we're fine with this, if not we'd have to bump the plan version to `2` which would make older service versions fail early and with a more recognizable error message

We've discussed this offline, and our plan is to bump sync plans to v2 if they use the new functions in expressions. I will open a follow-up PR for that, hopefully sometime next week. That probably doesn't include too many changes in this case, but it also helps us establish the process of increasing sync plan versions, as we've not done that before.

We want to keep this PR open until the follow-up PR is in place, but it looks good to me in its current shape.

henriquekraemer:
> > Maybe we're fine with this, if not we'd have to bump the plan version to `2` which would make older service versions fail early and with a more recognizable error message
> 
> We've discussed this offline, and our plan is to bump sync plans to v2 if they use the new functions in expressions. I will open a follow-up PR for that, hopefully sometime next week. That probably doesn't include too many changes in this case, but it also helps us establish the process of increasing sync plan versions, as we've not done that before.
> 
> We want to keep this PR open until the follow-up PR is in place, but it looks good to me in its current shape.

Sounds great. Happy to rebase once the follow-up lands.

simolus3 (dismissed):
I'm happy to merge this (and https://github.com/powersync-ja/powersync-service/pull/717 for sync plan versioning before we release this) in its current state.

One thing I wonder is whether it could make sense to also expose the full table name, perhaps via `.table_name()`, even when no suffix is used? Obviously one can reconstruct it via `SELECT * FROM "users_%" AS users WHERE 'users_' || users.table_suffix() == ...`, but it still feels like a convenient thing to have in case the auth parameter contains the full table name and it should be relatively easy to add now before we commit to the new sync plan version. What do you think?

henriquekraemer:
> I'm happy to merge this (and #717 for sync plan versioning before we release this) in its current state.
> 
> One thing I wonder is whether it could make sense to also expose the full table name, perhaps via `.table_name()`, even when no suffix is used? Obviously one can reconstruct it via `SELECT * FROM "users_%" AS users WHERE 'users_' || users.table_suffix() == ...`, but it still feels like a convenient thing to have in case the auth parameter contains the full table name and it should be relatively easy to add now before we commit to the new sync plan version. What do you think?

Good idea, and agreed it is easier to add now. `.table_name()` returning the full replicated table name (`table.name`) is a small addition alongside the existing two, so I'll add it. Want me to push it here on #703 so it lands before the v2 plan version in #717?

simolus3:
Yes, let's add that here. I can then merge this PR and change the target branch for the follow-up to `main`.

henriquekraemer:
> Yes, let's add that here. I can then merge this PR and change the target branch for the follow-up to `main`.

Added in 1736a0c: `users.table_name()` returns the full replicated table name, on any table (no wildcard required, so no warning like `table_suffix()` has on non-wildcard tables). Changeset updated as well.

simolus3 (approved):

PR #706, opened 2026-07-01
[MySQL] Fix idle keepalive LSN stall and multi-server-UUID GTID parsing

### LINKED ISSUE #704: [MySQL] Checkpoint creation stalls on idle servers: heartbeat keepalive LSN is behind the last commit LSN
### Background

This comes out of a PowerSync support case: a customer running Service 1.22.0 with a MySQL source and MongoDB bucket storage reported replication lag on a quiet test instance that never recovered on its own. Waiting did nothing, compaction did nothing, and only restarting the service cleared it. We root-caused it from their logs, then reproduced it in our support workbench against the pinned `journeyapps/powersync-service:1.22.0` image, and verified the code is unchanged on current `main`.

### Symptom

On an idle server (no writes arriving), the service logs this every ~30 seconds, indefinitely:

```
INF [powersync_1_429d] Waiting before creating checkpoint, currently at
0000000039489899|2e35321d-...:39489899|mysql-bin.000027|565835613 / 0.
Current state: {"snapshot_done":true,
"last_checkpoint_lsn":"0000000039489899|2e35321d-...:39489899|mysql-bin.000027|565836177", ...}
```

Note the signature: the current LSN and `last_checkpoint_lsn` share the same padded transaction id (39489899), but the current binlog offset (565835613) is *behind* the checkpoint offset (565836177).

### Root cause

MySQL LSNs are strings compared lexicographically (for example `canCheckpointState` in `modules/module-mongodb-storage/src/storage/implementation/CheckpointState.ts`, which requires `lastCheckpointLsn <= lsn`). In `modules/module-mysql/src/replication/zongji/BinLogListener.ts`:

- The GTID event handler builds `currentGTID` with the position of the GTID event, i.e. the *start* of the transaction.
- The Xid handler builds the commit LSN from `evt.nextPosition`, i.e. the *end* of the transaction. That LSN becomes `last_checkpoint_lsn`.
- Heartbeat keepalives re-send `currentGTID.comparable`, the stale start-of-transaction LSN.

So the moment the last transaction commits on a quiet server, every subsequent heartbeat presents an LSN that sorts below the stored checkpoint, and checkpoint creation is blocked until the next real transaction arrives. The same applies to the DDL auto-commit and non-transactional query commit paths in `processQueryEvent`.

### Reproduction

Fully scripted in our workbench; the minimal manual version:

1. Start the service (1.22.0 or main) with a MySQL source, let the snapshot complete
2. Commit one transaction, then leave the server idle
3. From the first heartbeat (~30s, `KEEPALIVE_INACTIVITY_THRESHOLD = 30`) onward, the service logs `Waiting before creating checkpoint` with the same transaction id and a smaller binlog offset than `last_checkpoint_lsn`
4. A single new write unblocks it immediately

In our repro the commit stored offset 607 while every idle keepalive re-sent offset 276 for the same transaction, matching the customer's log pattern digit-for-digit at their offsets.

### Impact

- Blocked (empty) checkpoints and log spam on idle instances
- Confusing for operators: it presents as stuck replication lag on quiet instances, and in this customer's case led to an unnecessary bucket storage volume deletion
- The replication lag metric itself stays at 0 in this state (the last commit succeeded), so the noise is worse than the harm, but the checkpoint gate is genuinely closed until the next write

A fix PR follows shortly: all commit paths now advance `currentGTID` to the commit position, so keepalive LSNs are never behind the last checkpoint LSN.

---

🤖 AI disclosure: this issue was generated by Claude (via Claude Code). The investigation, reproduction, and write-up were produced by Claude working from the customer's logs, directed and reviewed by @michaelbarnes.

### LINKED ISSUE #705: [MySQL] Multi-server-UUID GTID sets are parsed into NaN LSNs and can permanently block checkpoints
### Background

Found in the same PowerSync support case as the idle keepalive stall issue: a customer on Service 1.22.0 (MySQL source, MongoDB bucket storage) whose MySQL server has GTID history from two server UUIDs (typically the result of a failover, restore, or host migration). Their stored `no_checkpoint_before` LSN contained a literal `NaN`:

```
"no_checkpoint_before":"0000000000000NaN|2e35321d-...:1-37729572,\n314306f3-...:1-2734181|mysql-bin.000026|537764557"
```

We reproduced this against the pinned 1.22.0 image and verified the parsing code is unchanged on current `main`.

### Root cause

`ReplicatedGTID.comparable` (`modules/module-mysql/src/common/ReplicatedGTID.ts`) assumes the raw GTID is a single `uuid:ranges` value. But `readExecutedGtid()` feeds it `Executed_Gtid_Set` from `SHOW MASTER STATUS`, and MySQL joins multiple UUID sets with `,\n`. The parser then does:

- `raw.split(':')` and takes only the segment between the first two colons, so the second UUID lands in the transaction range list
- The "range" `\n314306f3-ff7b-...` splits on `-` into `\n314306f3` and `ff7b`, and `parseInt('ff7b', 10)` is `NaN`
- `Math.max(..., NaN)` is `NaN`, padded to the literal `0000000000000NaN`

The same single-UUID assumption also silently drops intervals for multi-interval sets (`uuid:1-5:11-18` parses as 5 instead of 18) and produces `NaN` for MySQL 8.4 tagged GTIDs (`uuid:tag:23`).

### Severity

LSNs are compared as plain strings, and `'N'` sorts above every digit, so the corrupted value has two modes:

- On servers with a transaction counter of 1000 or more, the `NaN` LSN sorts low and stays mostly latent (this customer's case)
- On servers with fewer than 1000 transactions (fresh or low-traffic instances), `0000000000000NaN` sorts above every real LSN. In our reproduction it was stored into both `no_checkpoint_before` and `last_checkpoint_lsn` at snapshot time, after which every committed write replicated but was never published: no checkpoint was ever created, the `powersync_replication_lag_seconds` metric climbed monotonically without bound, and a service restart did not recover the instance. The only way out is clearing bucket storage for a fresh snapshot, or repairing the GTID history

That second mode plausibly explains why the customer ended up deleting their bucket storage volume after a restart alone did not stick.

### Reproduction

On a fresh MySQL 8.0 server, before the service's initial snapshot:

```sql
SET @@GLOBAL.gtid_purged = '+314306f3-ff7b-11ef-a0e0-566fbaa00002:1-2734181';
```

Then start the service and write rows. Observed: `NaN` in the stored LSNs, all 12 test writes blocked (`Waiting before creating checkpoint` with the NaN state on every commit), lag metric climbing 10s to 327s across the run, and no recovery after a service restart.

### Fix

A fix PR follows shortly: `comparable` now parses full GTID sets (multiple UUIDs, whitespace and newline tolerant, multiple intervals per UUID) and takes the maximum transaction id across the set, never producing `NaN`. The output format is byte-identical for previously-correct inputs, since persisted LSN strings in bucket storage must keep comparing correctly across the upgrade. Instances already poisoned self-heal once their transaction id reaches 1000; low-transaction instances still need a resync, which the fix cannot avoid retroactively.

---

🤖 AI disclosure: this issue was generated by Claude (via Claude Code). The investigation, reproduction, and write-up were produced by Claude working from the customer's logs, directed and reviewed by @michaelbarnes.

michaelbarnes:
Fixes #704
Fixes #705

### Background
These fixes address two MySQL replication issues found in the same support case: idle replication could leave checkpoints blocked, and GTID sets containing historical server UUIDs could produce invalid NaN LSNs.

#### Changes

- Keepalive LSNs now use the end position of the last committed transaction across Xid, DDL auto-commit, and non-transactional commit paths.
- Replication heads are derived from the active MySQL server UUID instead of storing the complete Executed_Gtid_Set.
- Multi-interval and tagged GTID entries are reduced to a single uuid:transaction_id, preventing invalid or NaN comparable LSNs.
- Servers with no transactions for the active UUID use a synthetic uuid:0 GTID while retaining the current binlog coordinates.
- Resume validation now checks:
  * The source server UUID has not changed.
  * The binlog file and stored offset remain available.
  * Non-zero GTIDs are still present in gtid_executed.

- Invalid or rewound resume positions trigger a fresh snapshot.
- MySQL replicas are rejected during source configuration checks.
- Transactions originating from a server UUID other than the active source are rejected until multi-origin ordering is supported.
- Serialized GTIDs and binlog offsets are now validated rather than silently accepting malformed values.

#### Tests

Added and updated unit coverage for GTID parsing, zero GTIDs, resume validation, replica detection, malformed serialization, and committed keepalive positions.

#### Current limitations

PowerSync still does not support replication directly from a MySQL replica or binlogs containing transactions from multiple active origins. MySQL 8.4 GTID tags are not preserved; tagged entries are normalized to uuid:transaction_id.

---

🤖 AI disclosure: Investigated and reproduced issues with Claude Code working from the customer's logs. 
Implementation, tests and review were done using Claude Opus 5 and ChatGPT 5.6.  Directed and reviewed by @michaelbarnes, @Rentacookie @bean1352


Rentacookie:
Hi @michaelbarnes 
Thank you these changes look great! I just want to investigate the `multi-server-UUID GTID sets` issue a bit more.
Its good that we parse the GTID sets correctly now, but I am actually not sure what effect it has for our replication consistency were we to process GTIDs from multiple servers 😬

michaelbarnes:
@Rentacookie, when you mention "am actually not sure what effect it has for our replication consistency were we to process GTIDs from multiple servers"

Do you mean multiple servers connecting to a single PowerSync Service instance?

Rentacookie:
@michaelbarnes 
> Do you mean multiple servers connecting to a single PowerSync Service instance?

No, I mean more that events from multiple MySQL servers are appearing on binlog for the MySQL server that PowerSync is actually connected to. I believe this can happen if the MySQL DB we are connecting to is a replica itself, replicating from other MySQL servers.

Rentacookie (approved):

Rentacookie (changes requested):
Since the multi server uuid is unsupported, for now we'll add a check that does not allow replication to continue if multiple server uuids are found on the binlog.

bean1352:
> Since the multi server uuid is unsupported, for now we'll add a check that does not allow replication to continue if multiple server uuids are found on the binlog.

I went with a warning rather than a hard stop. Happy to make it a hard error instead if you'd prefer.

Rentacookie (approved):

Rentacookie:
> > Since the multi server uuid is unsupported, for now we'll add a check that does not allow replication to continue if multiple server uuids are found on the binlog.
> 
> I went with a warning rather than a hard stop. Happy to make it a hard error instead if you'd prefer.

I reckon that is fine for now until we implement the full multi server uuid handling.

Rentacookie (changes requested):
Discussed strategy for selecting current server uuid for choosing which uuidset to use.

modules/module-mysql/src/common/ReplicatedGTID.ts
```diff
@@ -80,32 +85,73 @@ export class ReplicatedGTID {
...
+   *
+   * The raw GTID can be a full GTID set consisting of multiple comma-separated
+   * (optionally whitespace/newline padded) UUID sets, each of the form
+   * `server_uuid:interval[:interval...]` where an interval is `n` or `n-m`.
+   * The maximum transaction id across all UUID sets is used for ordering.
+   *
+   * Note: this assumes the currently writing server has the highest transaction
+   * counter in the set. If a stale server UUID in the set has a higher counter
+   * than the active server (e.g. after a restore to a new server), checkpoints
+   * can be delayed until the active server's counter catches up.
```
stevensJourney: Will this catch up automatically, or would actual additional transactions be required to get there?

bean1352: It would not have caught up automatically. The latest commit removes this case entirely. We now query `@@server_uuid` and order LSNs by the connected server's own counter, so a stale uuid with a higher count is ignored.

modules/module-mysql/src/common/ReplicatedGTID.ts
```diff
@@ -80,35 +91,72 @@ export class ReplicatedGTID {
...
+   * When the connected server's UUID is known (see {@link ReplicatedGTIDSpecification.serverUuid}) and has
+   * transactions in the set, ordering uses that server's transaction counter: it is the counter that grows
+   * with new transactions. Otherwise the maximum counter across all UUID sets is used, which can pin the
+   * LSN to a stale server's higher counter (e.g. after a restore to a new server) and delay checkpoints
+   * until the active counter catches up.
    *
    * @returns A comparable string in the format
    *   `padded_end_transaction|raw_gtid|binlog_filename|binlog_position`
    */
   get comparable(): string {
```
Rentacookie, line 119: If it is possible to supply the active server uuid here as a parameter. Then the `ReplicatedGTID` does not need to store the serverUuid at all.  ie
```
getComparable(serverUuid?: string)
```

That then greatly simplifies the usage.

Rentacookie: Ah nvm that affects too many places..

modules/module-mysql/src/common/ReplicatedGTID.ts
```diff
@@ -31,40 +36,56 @@ export type BinLogGTIDEvent = {
  * and position where this GTID could be located.
  */
 export class ReplicatedGTID {
+  private options: ReplicatedGTIDSpecification;
+
+  constructor(options: ReplicatedGTIDSpecification) {
```
Rentacookie, line 41: The main focus here is to ensure that a `ReplicatedGTID` only represents a single GTID since that preserves the integrity of our checkpoint LSNs. 

That means no GTID sets and no transaction ranges are allowed. GTIDs read from events on the Binlog were always already in this format.

When reading the executed GTID set to determine the replication head and for snapshot checkpoints we now determine the correct single GTID based on the active server uuid.

modules/module-mysql/src/replication/BinLogStream.ts
```diff
@@ -220,16 +222,23 @@ export class BinLogStream {
...
-          const isAvailable = await common.isBinlogStillAvailable(connection, lastKnowGTID.position.filename);
+          // Check if the active server uuid matches the one in the GTID
+          if (this.activeServerUuid !== lastKnowGTID.serverUuid) {
+            this.logger.info(
+              `The source server uuid has changed. Active server uuid ${this.activeServerUuid} does not match the server uuid from the resume checkpoint: ${lastKnowGTID.serverUuid}, re-snapshotting to ensure consistency.`
+            );
+            return false;
+          }
+
+          const isAvailable = await common.isGtidPositionStillAvailable(connection, lastKnowGTID);
```
Rentacookie, line 235: In addition to checking if the binlog is still available, we now do the following validation on the resume LSN's GTID as well:
- Confirm that the server uuid from the GTID matches the active server uuid
- That the transaction id from the GTID is actually still available on the server

If these checks fail a full re-snapshot is required.

Rentacookie (approved):

stevensJourney (approved):
Happy with these changes.

PR #711, opened 2026-07-07
Fix 'evicted' errors

rkistner:
When LRUCache entries are evicted while the fetch is still in progress, it cancels the request and results in an unhelpful `'evicted'`. We've already sent the db query, so it makes more sense to just wait for the results. This is what the `ignoreFetchAbort: true` config flag does.

This also change the log message for errors like these, to include the original stack trace instead of just the place where it's wrapped in socket-route.ts.


stevensJourney (approved):
Changes look good to me :)

PR #713, opened 2026-07-08
[MongoDB Storage V3] Tweak size estimates and chunking to use actual BSON sizes

rkistner:
This uses actual BSON sizes for estimating operation sizes for:
1. Estimating chunk sizes when chunking operations.
2. Calculating the `size` field for the chunk when persisting.

Previously only the `data` length was used, which meant that MOVE and REMOVE operations did not get considered in the size at all. In theory that could lead to a situation where a chunk contains 100k MOVE operations and then exceeds the 16MB limit.

The calculations here are not meant to be exact, but it should be a better estimate than what we had.

This does add some overhead, but that should be small compared to the actual persistence overhead.

## AI Usage

Used Codex gpt-5.5 to investigate and suggest changes, but implemented the changes myself.

### DIFF
--- a/.changeset/neat-colts-switch.md
+++ b/.changeset/neat-colts-switch.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-module-mongodb-storage': patch
+---
+
+[MongoDB Storage V3] Tweak size estimates and chunking to use actual BSON sizes.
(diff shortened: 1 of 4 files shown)


stevensJourney (approved):
Changes make sense and look good to me

PR #717, opened 2026-07-15
Add sync plan version 2

simolus3:
#703 introduced syntax exposing table and schema names to bucket data and parameter creators, a change that requires changes to the serialized form of sync plans.

Older service versions would not load these correctly, so they should reject those plans outright. The version field added to sync plans helps here, older services throw if it's not `1`. This PR:

1. Introduces version `2` for sync plans, used when `.schema()`, `.table_suffix()` or `.table_name()` are used in an expression of the sync plan. Existing deployments not using the new row metadata functions are unaffected.
2. Improves the error message when we encounter sync plan versions the current service doesn't understand (this only starts being relevant when we increase the version again in the future, but it's helpful to have).

By continuing to use plan version 1 where possible, we allow service downgrades as long as the new functions aren't used in sync plans (if there are, there's no way an older service could implement those streams either way).

### DIFF
--- a/.changeset/yummy-showers-agree.md
+++ b/.changeset/yummy-showers-agree.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Improve error message when trying to load Sync Streams deployed with a more recent service version on older services.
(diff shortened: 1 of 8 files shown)


stevensJourney (approved):
The changes here look good to me.

PR #719, opened 2026-07-20
fix(postgres-storage): Fix stale checkpoint events when connections a…

### LINKED ISSUE #718: API instances silently stop receiving checkpoint updates under load while replication stays healthy (Postgres storage, split api/sync roles)
All API instances independentlt updates while the
replication instance keeps advancing normally. Each API instance then serves its
own frozen checkpoint to everytely. There is no error,
no crash, no failing health probe, and CPU usage *drops*. Clients believe they
are fully synced.

This happened twice in one day API deployment resolves
it, but only for a few hours — the second occurrence came 3h22m after the restart.

## Environment

- **PowerSync Service:** 1.22.0 (`journeyapps/powersync-service:1.22.0`)
- **Source database:** Postgre
- **Bucket storage:** PostgreSQL
- **Deployment:** Kubernetes,
  - `start -r api` — 6 replicas, 1 CPU / 1Gi each
  - `start -r sync` — 1 replic
- **Sync rules:** edition 3, streams. Group id 36, `state=ACTIVE`,
  `snapshot_done=t`, `last_fatboth incidents.
- `api.parameters.max_concurrent_connections: 150`
- **Clients:** `powersync-dart Android (~900 POS devices)

## Incident 1

Each API pod's last checkpoints (UTC):

| Pod   | Last advance | Froze
|-------|--------------|------------|
| mllnx | 00:13:51     | 13522
| sbv4b | 00:20:30     | 135220685  |
| mld9v | 00:31:19     | 13522
| 4x4cn | 00:58:39     | 135221126  |
| rxhpp | 00:59:12     | 13522
| vvwns | 01:06:35     | 135221511  |

Six instances stopped **one at a time over 53 minutes**, each at a different
value. Meanwhile `powersync.sy

04:43:20Z   last_checkpoint =
04:55:30Z   last_checkpoint = 135245994     (~86 ops/min, healthy)

Gap at discovery: **~24,500 operations**, widening. Total unnoticed duration:
~4h40m. A rollout restart of t0Z brought all six pods
back to within 20–160 operations of the replicator.

### Clients can receive a lower checkpoint than they already applied

Because each instance froze at a different value, a client reconnecting to a
different pod moved backwards.

04:49:33Z  pod rxhpp  →  13522
04:49:37Z  pod mllnx  →  135220613     (537 ops backwards, 4 seconds later)

## Incident 2 — progressive degradation, then full stall

After the restart the gap did not stay at zero. It grew steadily, then stopped
entirely:

05:16Z   gap 20–160          (
08:33Z   gap ~12,250         (still advancing, but ~2.5h behind)
08:32Z   last advance on every

Two consecutive samples ten mitical checkpoints across
all six instances while the replicator continued at ~81 ops/min:

=== sample 1  08:43:50Z          === sample 2  08:53:51Z
7wl9z  135251041
hdbx5  (no checkpoint lines)     hdbx5  135250965
jkrjc  135251045
lb9c5  135250995                 lb9c5  135250995
qhsvg  135251053
zgtgh  135250971                 zgtgh  135250971

The instances were still actively serving clients during this window — `hdbx5`
logged no checkpoint lines at 8:53Z, serving the same
stale checkpoint.

**Time from restart to full stall: 3h22m.**

So the failure is not a sudden stop. Instances fall progressively further behind
and then halt. Both stalls begg load: 07:13–08:06 local
time (stores opening) and 15:32 local time.

## What still works while an instance is stalled

This is the part we find most surprising:

- **Storage DB access is fine.** On pod `mllnx`, which stopped receiving
  checkpoints at 00:13:51Z, `Wucceeded at **00:16:32Z**
  and **00:18:16Z** — writes to the same Postgres storage still worked.
- **HTTP and streaming work.**tream` is served, and
  per-client checkpoint computation runs — just against a stale value.
- **Probes return 200** the enr restarts anything.
- **CPU drops to 2–4 millicores** against a 1000m limit (0.2–0.4%); memory
  231–299Mi against 1Gi.

## Ruled out

We eliminated these with measu

- **Replication health** — `sy `last_fatal_error` null,
  `last_keepalive_ts` fresh, `last_checkpoint` advancing throughout.
- **Storage connection exhaustnnections` in use.
- **Dead storage connections** — writes succeeded after the freeze (see above).
- **CPU / memory pressure** —
- **The `compact` job** — ran 9 hours before the first stall.
- **Client tokens / connectiviged near the freeze times
  were `PSYNC_S2103 JWT has expired` from clients retrying with stale tokens,
  unrelated to the checkpoint
- **`no_checkpoint_before` gating** — value `00000097/B679F8E8`, far below the
  current LSN `000000A4/AC63BC

## Not a duplicate of #704

We checked #704 (`[MySQL] Checidle servers`). It is a
different failure:

| | #704 | This report |
|---|---|---|
| Source | MySQL (binlog/GTID) | PostgreSQL |
| Storage | MongoDB | PostgreS
| Trigger | Idle server | Under load |
| What stalls | Replicator sto| Replicator is healthy; APIinstances stop *receiving* them |
| `last_checkpoint` | Frozen |
| Logs | `"Waiting before creating checkpoint"` every 30s | No message at all |

The failure *shape* is similar though, and #704's observation that
`"The replication lag metric itate"` matches our
experience exactly: nothing in the system reports a problem.

## Questions

1. Is the API instance's checkpoint update path expected to log an error or fail
   its health probe if it term no observable signal —
   a 5-hour data outage is indistinguishable from healthy operation.
2. Is there a supported way toess per API instance? We
   currently compare `sync_rules.last_checkpoint` against checkpoint values
   scraped from each pod's logtended usage.
3. Should the protocol prevent a client from receiving a checkpoint lower than
   one it has already applied,tate can diverge?

## Next steps on our side

We are upgrading to 1.23.3, prrite-checkpoint2`
handling in 1.23.2 (our logs are dense with those requests and the failure
correlates with load) and for te` /
`partial_checkpoint_complete` / `checkpoint_invalidated` events, which we expect
will make this far easier to d

We can provide full pod logs fing both incidents.

stevensJourney:
Fixes https://github.com/powersync-ja/powersync-service/issues/718

# Overview

Postgres bucket storage uses `LISTEN` to notify API runners about new checkpoints. This communication channel bridges the `sync` and `api` runners, which can run in different processes, machines, or pods.

If the underlying Postgres connection is terminated, for any reason, we no longer receive checkpoint notifications. This can cause clients to continue using stale checkpoint state and lead to serious syncing issues.

One way for this connection to close is through `idle_session_timeout`. The linked issue was reproducible by setting the idle session timeout to a short value.

## Fix

The initial fix used the existing pgwire `whenDestroyed` event to detect when the notification connection had closed. We then reconnect that slot and restore its `LISTEN` subscriptions.

Reconnecting by itself was not enough, since checkpoints may be committed while no connection is listening. Once `LISTEN` has been restored, we now notify the checkpoint watcher so that it queries the latest persisted checkpoint and catches up on anything it missed. The listener is registered before the initial checkpoint query, and stale or duplicate events are ignored, which covers checkpoints changing around the initial query and reconnection boundaries.

Checkpoint notifications are now parameterized and committed atomically with the state they describe. Processing sync configs no longer emit active-checkpoint notifications, and activating a replacement config closes the old checkpoint stream immediately.

The reconnect path also exposed a few assumptions in `ConnectionSlot`. Previously, `poke()` issued a `SELECT 1` query to check whether a connection was still usable. The pgwire `whenDestroyed` event now provides that lifecycle signal directly, allowing us to clear and restore a destroyed connection without the additional query.

A lease belongs to the slot rather than a particular underlying connection, so a replacement connection must remain unavailable until the outstanding slot lease is released. Connection setup is deduplicated, late connections are closed during disposal, and failures are tracked per slot so one failed notification connection does not reject work that another healthy slot can serve. Notification connections reconnect proactively to restore `LISTEN`; ordinary transaction connections reconnect lazily so server-side idle timeouts do not create a permanent reconnect loop. A short delay is applied between failed connection attempts.

## Testing

The integration tests reproduce the original idle-timeout failure and pause the replacement connection before `LISTEN` is restored. They verify that missed checkpoints are recovered in order, checkpoints changing during the recovery query do not regress the stream, unchanged checkpoints are not emitted twice, and the old stream closes when a replacement sync config becomes active.

Additional connection-slot tests cover lease ownership across replacement connections, proactive and lazy reconnection, concurrent connection attempts, slot-scoped failures, retry exhaustion, and disposal while a connection is still opening.

---

AI disclosure: Codex GPT 5.6 assisted in fixing this issue. I reviewed and verified the work.

### DIFF
--- a/.changeset/quiet-lions-listen.md
+++ b/.changeset/quiet-lions-listen.md
@@ -5,4 +5,4 @@
 '@powersync/service-module-postgres-storage': patch
 ---
 
-Reconnect PostgreSQL notification connections and restore `LISTEN` subscriptions when the underlying connection is terminated.
+Reconnect PostgreSQL notification connections and restore `LISTEN` subscriptions when the underlying connection is terminated. Harden connection-slot retries and lease handling, and publish checkpoint notifications atomically with checkpoint updates.
--- a/.changeset/quiet-lions-listen.md
+++ b/.changeset/quiet-lions-listen.md
@@ -1,6 +1,6 @@
 ---
 '@powersync/lib-service-postgres': patch
-'@powersync/service-core': patch
+'@powersync/service-image': patch
 '@powersync/service-jpgwire': patch
 '@powersync/service-module-postgres-storage': patch
 ---
--- a/.changeset/quiet-lions-listen.md
+++ b/.changeset/quiet-lions-listen.md
@@ -1,0 +1,8 @@
+---
+'@powersync/lib-service-postgres': patch
+'@powersync/service-core': patch
+'@powersync/service-jpgwire': patch
+'@powersync/service-module-postgres-storage': patch
+---
+
+Reconnect PostgreSQL notification connections and restore `LISTEN` subscriptions when the underlying connection is terminated.
--- a/libs/lib-postgres/src/db/connection/ConnectionSlot.ts
+++ b/libs/lib-postgres/src/db/connection/ConnectionSlot.ts
@@ -16,6 +16,15 @@ export interface NotificationListener {
   notificationChannelsRegistered?: () => Promise<void>;
 }
 
+/**
+ * Listener keys that operate on the notification connection. These are handled
+ * on the dedicated notification slot rather than the general connection pool.
+ */
+export const NOTIFICATION_LISTENER_KEYS = [
+  'notification',
+  'notificationChannelsRegistered'
+] as const satisfies readonly (keyof NotificationListener)[];
+
 export interface ConnectionSlotListener extends NotificationListener {
   connectionAvailable?: () => void;
   connectionError?: (exception: any) => void;
@@ -71,7 +80,11 @@ export class ConnectionSlot extends framework.BaseObserver<ConnectionSlotListene
      * Subscribing to notifications, even without a registered listener, should not add much overhead.
      */
     await this.configureConnectionNotifications(connection);
-    connection.whenDestroyed.then(() => this.handleConnectionDestroyed(connection));
+    // whenDestroyed normally resolves, but guard against a rejection becoming an
+    // unhandled promise rejection. Either outcome means the connection is gone.
+    connection.whenDestroyed
+      .catch((error) => framework.logger.debug('Postgres connection destroyed with an error', error))
+      .then(() => this.handleConnectionDestroyed(connection));
     return connection;
   }
 
@@ -120,7 +133,7 @@ export class ConnectionSlot extends framework.BaseObserver<ConnectionSlotListene
   };
 
   protected hasNotificationListener() {
-    return !!Object.values(this.listeners).find((l) => !!l.notification);
+    return !!Object.values(this.listeners).find((l) => NOTIFICATION_LISTENER_KEYS.some((key) => !!l[key]));
   }
 
   /**
--- a/libs/lib-postgres/src/db/connection/ConnectionSlot.ts
+++ b/libs/lib-postgres/src/db/connection/ConnectionSlot.ts
@@ -1,5 +1,6 @@
 import * as framework from '@powersync/lib-services-framework';
 import * as pgwire from '@powersync/service-jpgwire';
+import * as timers from 'node:timers/promises';
 
 export type NotificationEvent =
   | { type: 'notification'; notification: pgwire.PgNotification }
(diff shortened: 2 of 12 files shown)


rkistner (approved):
Changes make sense to me on a high level.

PR #722, opened 2026-07-21
Improve yaml errors when parsing sync configs

simolus3:
When parsing a sync config from YAML, we report errors in two phases:

1. First, we traverse known keys to parse them, manually reporting errors for stuff that doesn't make sense.
2. As a catch-all, we then run ajv to validate the yaml against a sync config schema. This catches additional things like additional keys or wrong types.

This setup is not ideal:

- ajv errors provide no source span, so it's hard for users to tell what's caused an error.
- Some of the errors should really be caught as we try to parse the document. For example, we regularly cast nodes to a subtype without checking the type at runtime. If the user provides an array where we expect a map, we crash in the first step and never get to the second ajv scan.

To improve this, this PR catches all errors as we traverse the YAML: Instead of calling `as YamlMap` for example, we use a helper structure on which we can call `requireMap()`, which returns `null` if it's not a map and automatically reports an error. This alone makes our parsing code much more robust against wrong types. By having the helper structure track read keys, it can also detect illegal extra keys in the source map in a `Symbol.dispose` function.

AI use: Parts of the implementation and test helpers were written by Claude with manual adjustments. Test changes are manual.


stevensJourney (approved):
The updated errors look nice to me :)

packages/sync-rules/src/yaml_validation.ts
```diff
@@ -0,0 +1,224 @@
...
+      requireSequence: unexpectedType(node, 'a sequence'),
+      [Symbol.dispose]() {
+        // Treat keys that haven't been extracted as invalid an report an error.
+        if (matchedKeys != null) {
+          for (const item of node.items) {
+            const key = item.key;
+            const keyName = key instanceof Scalar ? String(key.value) : undefined;
+            if (keyName == null || !matchedKeys.has(keyName)) {
+              report(createYamlError(key as Node, `Unknown key '${keyName}'.`));
+            }
```
rkistner, line 152: Just out of interest, in which cases would key not be a `Scalar`? Is it worth having a better error message for those than `Unknown key 'undefined'`?

simolus3: In almost all cases it would be a scalar string, but YAML allows non-scalar keys:

```yaml
foo:
  [a, b]: c
```

`Unknown key 'undefined'` is not a great error message, I'll open a small follow-up to fix that. But either way, the span of of the error should cover `[a, b]` which helps for context.

PR #723, opened 2026-07-22
style: Improve diagnostics when a connection exceeds the bucket limit

stevensJourney:
## Context

Some users create enough buckets to exceed the default limit of 1,000 buckets per connection. When that happens, the
service raises `PSYNC_S2305` and logs a breakdown intended to show which parts of the sync configuration produced those
buckets.

That breakdown is valuable because the total alone does not tell a user where to start simplifying their configuration
or subscriptions. The connection limit is evaluated against the total number of distinct buckets, while the breakdown
groups those buckets by the Sync Stream or legacy bucket definition that produced them. The counts in those groups add
up to the buckets represented by the breakdown.

However, the existing log only includes the 10 largest definition entries. On configurations with many Sync Streams or
legacy bucket definitions, this can potentially hide some of the information needed to understand the problem.

The existing wording is also _slightly_ ambiguous. It labels the entries as "definitions," even though the names being
printed are Sync Stream or Bucket definition names. For example, a user could previously see a log resembling:

```text
[PSYNC_S2305] Too many buckets: 60 (limit of 50)
Buckets by definition:
  projects: 30
  tasks: 20
  comments: 10
```

Here, `projects`, `tasks`, and `comments` are Sync Stream names, and the numbers are the distinct buckets produced by
each stream.

## What this changes

This PR does **not** change the bucket limit itself. The default remains 1,000 actual buckets per connection. It only
changes how many grouped Sync Stream or bucket-definition entries can be printed when reporting that the bucket limit
was exceeded.

The bucket-limit diagnostic becomes more useful while retaining a safeguard against excessively large log messages:

- It raises the number of logged definition entries from 10 to 100.
- It identifies whether the entries came from Sync Streams or legacy bucket definitions and labels the log accordingly.
- It retains a summary for entries beyond the 100 largest contributors.
- It preserves the existing structured `buckets_by_definition` field for compatibility with current log consumers.

The original limit of 10 was likely intended to prevent unusually large log messages. That remains a reasonable
concern, but logging more of the available counts _could_ be valuable when diagnosing. A diagnostic-entry cap of 100 keeps the output bounded while making it much more likely that the relevant stream or definition is visible.

For example, `projects: 30`, `tasks: 20`, and `comments: 10` represent 60 actual buckets across three logged definition
entries. The bucket limit applies to the total of 60; the diagnostic-entry cap applies to the three grouped lines.

## Example output

For Sync Streams, the log now names the source correctly:

```text
[PSYNC_S2305] Too many buckets: 60 (limit of 50)
Buckets by sync stream:
  projects: 30
  tasks: 20
  comments: 10
```

For a legacy Sync Rules configuration, it continues to use bucket-definition terminology:

```text
[PSYNC_S2305] Too many buckets: 60 (limit of 50)
Buckets by bucket definition:
  projects: 60
```

When more than 100 Sync Streams contribute buckets, the 100 largest entries are included and the remainder is
summarized:

```text
[PSYNC_S2305] Too many buckets: 105 (limit of 100)
Buckets by sync stream:
  def1: 1
  def2: 1
  def3: 1
  def4: 1
  def5: 1
  ...
  def96: 1
  def97: 1
  def98: 1
  def99: 1
  def100: 1
  ... and 5 more buckets from 5 sync streams
```

## Testing

The updated tests cover:

- A Sync Stream breakdown with counts from multiple streams.
- A legacy bucket-definition breakdown with the appropriate label.
- A 105-stream breakdown that logs exactly 100 entries and summarizes the remaining five.

--- 
AI usage: The changes here were implemented with Codex 5.6 Sol - I manually guided and tested the result.

### DIFF
--- a/packages/service-core/src/sync/BucketChecksumState.ts
+++ b/packages/service-core/src/sync/BucketChecksumState.ts
@@ -2,6 +2,7 @@ import {
   BucketParameterQuerier,
   BucketPriority,
   BucketSource,
+  BucketSourceType,
   HydratedSyncConfig,
   mergeBuckets,
   QuerierError,
@@ -164,8 +165,9 @@ export class BucketChecksumState {
         const count = bucketsByDefinition.get(definition) ?? 0;
         bucketsByDefinition.set(definition, count + 1);
       }
-
-      const breakdown = formatBucketDefinitionBreakdown(bucketsByDefinition);
+      // Only 1 type is allowed per sync config
+      const bucketSourceType = this.parameterState.syncRules.bucketSourceDefinitions[0].type;
+      const breakdown = formatBucketDefinitionBreakdown(bucketsByDefinition, bucketSourceType);
       errorMessage += breakdown.message;
       logData.buckets_by_definition = breakdown.countsByDefinition;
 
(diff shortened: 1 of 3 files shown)


rkistner (approved):
I'm happy with this, but would be good to get @simolus3's input on whether we need additional improvements here.

packages/service-core/src/sync/BucketChecksumState.ts
```diff
@@ -164,8 +165,9 @@ export class BucketChecksumState {
         const count = bucketsByDefinition.get(definition) ?? 0;
         bucketsByDefinition.set(definition, count + 1);
       }
-
-      const breakdown = formatBucketDefinitionBreakdown(bucketsByDefinition);
+      // Only 1 type is allowed per sync config
```
simolus3, line 168: Technically you can have both sync streams and bucket definitions in a single config, with `edition: 2`. But we consider that deprecated, so since this only affects the message I think it's fine.

simolus3 (approved):
In terms of additional things worth adding, we might want to include subscription parameters for streams too (in case only some subscriptions cause a large amount of buckets). But that is a much larger change and this is a nice improvement on its own. And I'm not aware of a report where not including subscription parameters made the issue harder to diagnose, so I'm not sure if we really need to do that.

PR #727, opened 2026-07-24
[MSSQL]: Disable Automatic Schema change handling

stevensJourney:
## Summary

This PR pins each MSSQL source-table binding to a specific CDC capture instance and keeps the replicated table set and captured schema fixed for the lifetime of a replication stream. This lets an existing stream continue using its original capture schema while a newly deployed stream binds to a replacement capture instance and snapshots the current source table.

MSSQL replication no longer handles source schema changes automatically. Table wildcards (`%`) are therefore no longer supported: tables must be listed explicitly, and schema changes that affect replication require deploying the updated sync configuration as a new replication stream.

The PR also adds generic, source-owned `SourceTable` reconciliation and opaque `sourceMetadata` persistence to MongoDB and PostgreSQL storage.

### Why streams no longer adopt schema changes

SQL Server CDC does not include schema-change events in its replication stream. PowerSync can only discover them by polling database metadata, and that poll cannot be made atomic with source transactions or CDC reads. There is therefore no reliable source position at which PowerSync can say that one schema ended and another began.

Changing a stream's schema or table set in place could let it commit a checkpoint using metadata observed after the corresponding source changes. That can skip unread changes or expose a state that never existed atomically in the source database. MSSQL replication streams now keep their initial table set and captured schemas for their entire lifetime instead of trying to infer an unsafe transition boundary.

### Capture-instance pinning

Each new source-table binding selects the newest available CDC capture instance and persists the object ID of its change table. Legacy bindings are assigned and persist the instance the previous selection logic would have used. Restarts restore that binding, while polling refreshes its current LSN range without changing its identity.

When a newer capture instance appears, PowerSync warns but the existing stream remains pinned to its original captured schema. If the pinned instance is removed while streaming, replication stops with `PSYNC_S1601`; automatically adopting its replacement could silently change the captured columns without a safe transition boundary. If no capture instance is available when a job starts, initial table validation reports `PSYNC_S1602`.

### Adding tables

Table wildcards are no longer supported because a matching table could appear between metadata polling and checkpointing, with no atomic position from which to begin replicating it. Every configured table must be listed explicitly, exist, and have CDC enabled before the stream can start. A table that is unavailable or not CDC-enabled stops replication with `PSYNC_S1602` rather than being silently omitted.

Adding a table requires deploying the updated sync configuration as a new replication stream so that the table is included in that stream's initial snapshot and consistency boundary.

### Changing columns

Adding, removing, or changing source columns does not alter an existing CDC change table's captured schema. PowerSync warns when it observes a difference but continues reading the schema belonging to the pinned capture instance.

Adopting the new columns requires a replacement capture instance and a new replication stream. The new stream binds to the replacement capture instance and snapshots the current source table, while the active stream continues serving updates from the old instance.

### Dropping or renaming tables

If an active replication stream detects that a replicated table was dropped or renamed, replication stops with `PSYNC_S1603`. Continuing without it could advance the shared checkpoint past changes that remain unread from that table. Existing replicated data is retained until a replacement sync configuration is deployed. If the table is already absent when a job starts, initial exact-table validation reports `PSYNC_S1602`.

For all schema changes, keep the old tables and capture instances available until the replacement stream has completed its snapshots and become active.

## Storage and reconciliation

`resolveTables()` now loads all persisted records overlapping a discovered source table and lets the source connector classify isolated clones as compatible or incompatible through a read-only TypeScript interface.

Storage compares source metadata changes against its original hydrated records and rematerializes reconciliation results from its canonical `SourceTable` values. Only source metadata changes are accepted from the connector; snapshot state, memberships, identity, and cleanup remain storage-owned even if connector code bypasses the read-only type boundary.

Source metadata is nullable and opaque to storage. MongoDB v1/v3 persist it on source-table documents, and PostgreSQL adds a nullable `source_metadata` JSONB column. Connectors without custom reconciliation retain the existing generic identity comparison.

## Testing

Coverage includes source-table reconciliation and clone isolation, metadata persistence in all storage implementations, capture-instance selection and restart restoration, legacy binding backfill, newer or missing capture instances, fixed table-set validation, wildcard diagnostics, and table/column schema changes.

---

AI usage: The implementation was planned and iteratively reviewed with Codex. Claude Opus implemented an initial plan, followed by manual review and further Codex-assisted refinements.


modules/module-mssql/src/replication/CaptureReconciler.ts
```diff
@@ -0,0 +1,119 @@
...
+ * legacy metadata-free records.
+ */
+export function readCaptureMetadata(value: JsonValue | undefined): MSSQLSourceMetadata | null {
+  if (value == null || typeof value !== 'object' || Array.isArray(value)) {
+    return null;
+  }
+  const captureTableObjectId = (value as Record<string, unknown>).captureTableObjectId;
+  if (typeof captureTableObjectId !== 'number') {
+    return null;
+  }
```
rkistner: I think we should treat these cases (apart from `value == null`) as a hard error instead of silently returning null. If those cases are hit then generally either the data was corrupted, or a future version changed the format and the user downgraded again. In both those cases it's typically better to fail with a hard error, rather than silently continuing.

modules/module-mssql/src/replication/CaptureReconciler.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    // Pinned binding. Resolve the persisted capture identity against what is available.
+    const [captureTableObjectId] = [...pinnedObjectIds];
+    const available = availableInstances.some((instance) => instance.objectId === captureTableObjectId);
+    if (!available) {
+      throw new ReplicationAssertionError(
+        `The pinned capture instance (object id ${captureTableObjectId}) for source table ` +
+          `${source.schema}.${source.name} is no longer available. Re-enable CDC or redeploy the ` +
+          `sync configuration to adopt a new capture instance; PowerSync will not silently switch instances.`
```
rkistner:
```suggestion
          `sync configuration to adopt a new capture instance; PowerSync will not automatically switch instances.`
```

rkistner (commented):
I'm happy with these changes overall. I did not check the comments on minLSN in detail.

Mentioned by Codex on the existing recovery path:

>    modules/module-mssql/src/replication/CDCStream.ts:811 tells operators to re-enable CDC, while modules/
>    module-mssql/src/replication/CaptureReconciler.ts:107 gives similar guidance. Re-enabling CDC creates
>    a replacement change table with a different object ID, which the pinned stream deliberately refuses to
>    adopt. The actionable recovery is restoring the original capture instance if possible, or deploying a
>    new replication stream. The current message can leave operators expecting an indefinitely stopped
>    stream to recover.

modules/module-mssql/src/replication/CDCStream.ts
```diff
@@ -292,6 +305,24 @@ export class CDCStream {
     return resolvedTable;
   }
 
+  /**
+   * Select the capture instance a resolved binding must poll.
+   *
+   * Pinned bindings resolve to their persisted capture identity; legacy/metadata-free bindings use
+   * the newest available instance, matching origin/main behavior.
+   */
+  private selectBoundCaptureInstance(
```
Rentacookie: This looks like it could maybe be a convenience method on the `MSSQLSourceTable` model instead?

modules/module-postgres-storage/src/types/models/SourceTable.ts
```diff
@@ -14,7 +15,7 @@ export const ColumnDescriptor = t.object({
   /**
    *  Some data sources have a type id that can be used to identify the type of the column
    */
-  typeId: t.number.optional()
+  type_oid: t.number.optional()
```
Rentacookie, line 18: I'm a bit worried about divergence with this rename. Everywhere else the `ColumnDescriptor` definition still used `typeId`

modules/module-mssql/src/replication/CaptureReconciler.ts
stevensJourney: Agreed. `null` is now the only legacy/no-metadata value. Non-null values are decoded with the `MSSQLSourceMetadata` codec, so malformed or incompatible persisted metadata fails instead of being silently treated as a legacy record.

modules/module-mssql/src/replication/CaptureReconciler.ts
stevensJourney: Agreed. I removed the ambiguous wording. The current error explains that the stream cannot adopt a replacement capture instance and directs the operator to deploy the sync configuration as a new replication stream.

modules/module-postgres-storage/src/types/models/SourceTable.ts
stevensJourney: The public/core `ColumnDescriptor` still uses `typeId`; `type_oid` is the field stored in PostgreSQL's `replica_id_columns` JSON. The storage-row codec previously expected `typeId`, so it did not hydrate the persisted `type_oid` value into `SourceTable.replicaIdColumns` correctly.

That appears to have been a latent bug: the old resolution path compared the persisted values inside storage, so the missing type ID on the hydrated `SourceTable` was not used. Candidate reconciliation now performs that identity comparison in JavaScript, which means the hydrated `SourceTable` must contain the type ID. The storage boundary therefore decodes `type_oid` and maps it to the core `typeId` property (and maps `typeId` back to `type_oid` when writing). The shared descriptor API remains unchanged.

.changeset/mssql-fixed-table-set.md
```diff
@@ -0,0 +1,28 @@
...
+Pin the replicated schema at the start of replication for SQL Server connections.
+
+SQL Server does not deliver schema changes in the replication stream, so they can only be found by
+polling, and a poll is never atomic with a commit. That leaves a gap where checkpoints are committed
+against a schema that has already changed — producing a state the source database was never in, which
+clients cannot detect. PowerSync therefore no longer adopts schema changes automatically.
+
+What this means when making schema changes:
+
+- **Adding a table to replication:** deploy a new sync configuration. Table wildcards (`%`) are no
```
rkistner: Were wildcards ever supported? There was previously a check that rejects wildcards in `getTablesFromPattern`, but I haven't traced the actual usage.

.changeset/mssql-fixed-table-set.md
```diff
@@ -0,0 +1,28 @@
...
+- **Changing a table's columns:** deploy a new sync configuration. Replication continues against the
+  original captured schema and warns, since the bound capture instance keeps its own column list.
+- **Dropping or renaming a replicated table:** deploy a new sync configuration. Replication stops
+  with `PSYNC_S1603` — changes committed before the table went away may not have been read yet, so
+  continuing would commit past them and skip those rows. Already-replicated data for the table is
+  retained until the redeploy.
+- **Disabling and re-enabling CDC on a replicated table:** deploy a new sync configuration as a new
+  replication stream. Replication stops with `PSYNC_S1601`. The binding is pinned to the CDC change
+  table's object ID, and re-enabling CDC creates a new change table with a different ID, which the
+  existing stream will not adopt.
```
rkistner: I wonder if we should still re-snapshot these automatically, rather than stopping replication completely?

Re-snapshotting automatically when the CDC table changed was problematic: It could block replication while re-snapshotting, even when we did not need to resnapshot immediately.

But for the other cases, it could perhaps be better to recover automatically, rather than permanently stalling?

This could specifically help during development, where it is for example common to drop and re-create all tables to reset the db.

.changeset/mssql-fixed-table-set.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+[Breaking Change]
+
+Pin the replicated schema at the start of replication for SQL Server connections.
+
+SQL Server does not deliver schema changes in the replication stream, so they can only be found by
+polling, and a poll is never atomic with a commit. That leaves a gap where checkpoints are committed
+against a schema that has already changed — producing a state the source database was never in, which
+clients cannot detect. PowerSync therefore no longer adopts schema changes automatically.
```
rkistner, line 12: Does the new behavior (blocking rather than re-snapshotting) actually help here? It looks like it could still get into the same inconsistent state, but then stop further replication instead of recovering automatically (see further comments below).

.changeset/mssql-fixed-table-set.md
stevensJourney: It looks like on `main` we don't support `schema` wildcards. But we do support trailing wildcards for table names, e.g. `dbo.orders_%` - which support for that has been dropped in this PR.

.changeset/mssql-fixed-table-set.md
stevensJourney: You’re right that a hard failure cannot retroactively repair a checkpoint exposed before the schema change was detected. 

The intended behaviour currently is narrower: once detected, we do not apply the schema change at all. Replication fails and retains the last checkpointed state rather than assigning the change an arbitrary CDC position, resnapshotting, and then continuing as though ordering had been preserved.
This prevents further checkpoints until an explicit redeploy and makes our inability to order the schema transition visible instead of silently converging and presenting the flow as reliable.

That said, polling can still detect a schema change late - which can cause its own issues. This circles back to the underlying limitation: if a schema change is detected late, an inconsistent checkpoint may already have been exposed. Automatic recovery can converge afterward, but it cannot either undo that exposure.

.changeset/mssql-fixed-table-set.md
stevensJourney: > This could specifically help during development, where it is for example common to drop and re-create all tables to reset the db.

I think there might be room to add a few more strategies, but my current inclination is to expose this as a hard error, including during development. Given the current constraints, I don’t think we should encourage dropping and recreating a replicated table without explicitly deploying a new replication stream.

For context, schema changes are not part of the SQL Server CDC stream. They are detected by periodically querying the database’s current schema, independently of the CDC position being processed.
The case I particularly want to avoid is a schema poll observing the intermediate dropped state while CDC replication is lagging. The previous flow would immediately queue removal of the replicated table and then continue processing the CDC backlog. Its next commit could therefore be at an LSN from before the source table was dropped while already exposing the table’s removal: a state the source database never had.
If the drop and recreation are observed together, the replacement snapshot barrier should prevent that. A typical development reset may therefore recover correctly, but correctness depends on the schema poll not landing between the drop, recreation, and CDC enablement. A later snapshot can converge the current data, but at that point - we've already exposed the inconsistent data - which may cause difficult to debug issues in apps. 

We could add a recovery flow that waits for and snapshots the replacement, but that takes us back towards timing-dependent resnapshotting and blocking the replication stream. I think a clear, predictable failure could be less confusing during development and avoids presenting drop/recreate as a reliable production workflow.

modules/module-mssql/src/replication/CaptureReconciler.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    const [captureTableObjectId] = [...pinnedObjectIds];
+    const available = availableInstances.some((instance) => instance.objectId === captureTableObjectId);
+    if (!available) {
+      // A replacement may capture a different schema, so do not adopt it in place.
+      throw new CaptureInstanceMissingError(
+        `The CDC capture instance (object id ${captureTableObjectId}) for source table ` +
+          `${source.schema}.${source.name} is no longer available. Deploy the sync configuration as a ` +
+          `new replication stream to replicate this table against a new capture instance.`
```
benitav: "Deploy the sync configuration as a new replication stream" - can't we just say redeploy your sync config?

.changeset/mssql-fixed-table-set.md
rkistner: Ah, I missed that the previous check was for _schema_ wildcards and not _table_ wildcards.

modules/module-mssql/src/common/MSSQLSourceTable.ts
```diff
@@ -47,12 +55,26 @@ export class MSSQLSourceTable {
...
+   * Persisted capture-table object id, or null for a legacy binding.
+   */
+  get pinnedCaptureObjectId(): number | null {
+    return this.captureObjectId;
+  }
+
+  /**
+   * Find the available capture instance matching this table's persisted binding.
+   */
+  findPinnedCaptureInstance(availableInstances: readonly CaptureInstance[]): CaptureInstance | null {
```
Rentacookie: Looks like combining the functionality of findPinnedCaptureInstance and setCaptureInstance would result in a nice simplification. Example:
```
setCaptureInstance(availableInstances: readonly CaptureInstance[]) {
    this.captureInstance = availableInstances.find((instance) => instance.objectId === this.pinnedCaptureObjectId) ?? null;
   // Can consider throwing an exception here if we don't want the capture instance to be null at this point
  }
```
This simplifies the `MSSQLSourceTable` usage in CDCStream considerably since  most of the added pinnedCaptureObject logic remains inside in this class.

modules/module-mssql/src/replication/CaptureReconciler.ts
stevensJourney: Jip :) I've updated the description now

modules/module-mssql/src/common/MSSQLSourceTable.ts
stevensJourney: That is nicer! I've implemented that now.

stevensJourney:
The recommended MSSQL schema-change workflow is to treat a sync-config deployment as the consistency boundary. A replication stream keeps the exact table set and CDC capture-instance identities it selected when it started; it does not adopt schema or table changes in place. Tables must therefore be listed explicitly rather than selected with wildcards.

- **Add a table:** Create the table, enable CDC, add its exact qualified name to the sync config, and deploy. The new stream (sync config) snapshots the table before becoming active.
- **Change captured columns:** Where SQL Server allows a rolling change, apply the DDL and create a second capture instance with the desired captured columns. Update explicit sync queries if needed, deploy a new stream (sync config with changes), wait for its snapshots to finish and for it to become active, and only then remove the old capture instance.
- **Make an identity-breaking change:** For a primary-key change, identity-column rename, or another change that requires CDC to be disabled, disable CDC, apply the DDL, re-enable CDC, update the sync config if needed, and deploy a new stream. The old stream cannot adopt the replacement capture instance - this will result in downtime. 
- **Drop a table:** Prefer removing it from the sync config, deploying and activating the replacement stream, and only then dropping the source table.
- **Rename or drop/recreate a table:** Update the config to express the intended table explicitly, ensure the resulting table is CDC-enabled with the expected schema and replica identity, and deploy a new stream.
- **No deployment is needed** for changes that do not affect the captured row shape or replica identity, such as many index, constraint, or default changes. A non-identity column change also needs no PowerSync action if that changed column does not need to be replicated; otherwise use the rolling capture-instance workflow above.

This sequencing keeps the old table/capture instance available while the replacement stream snapshots and activates whenever a rolling transition is possible. If a pinned capture instance or bound table is removed first, the existing stream stops and a fresh deployment is required.

The [`temp-mssql-tests` branch](https://github.com/powersync-ja/powersync-service/tree/temp-mssql-tests) contains the [full schema-change analysis and workflow](https://github.com/powersync-ja/powersync-service/blob/temp-mssql-tests/docs/replication/10-mssql-schema-change-flows.md), along with [integration-test reproductions](https://github.com/powersync-ja/powersync-service/blob/temp-mssql-tests/modules/module-mssql/test/src/schema-change-inconsistencies.test.ts) of the old behavior on `main`. Those tests deliberately assert the reproduced inconsistent states, so a passing test means the inconsistency was successfully reproduced. The risks depend on timing, replication lag, and the specific transition; they are possible under the old flow, but not every occurrence of the corresponding DDL necessarily produces an inconsistent checkpoint.

--- 
This comment was generated by prompting Codex with context and test generated in the temporary branch. The tests in the temporary `temp-mssql-tests` branch were generated with help from Codex 5.6 and reviewed for accuracy with Claude Opus 5.

rkistner (dismissed):

.changeset/mssql-fixed-table-set.md
```diff
@@ -0,0 +1,28 @@
...
+
+What this means when making schema changes:
+
+- **Adding a table to replication:** deploy a new sync configuration. Table wildcards (`%`) are no
+  longer supported, so list each table explicitly, and every table must exist with CDC enabled when
+  replication starts.
+- **Changing a table's columns:** deploy a new sync configuration. Replication continues against the
```
benitav: To understand the "Replication continues against..." sentence better  - I assume that case happens when one _doesn't_ deploy new sync config? If that's the case this would be less confusing if we start the sentence with "Without this replication continues..." or similar.

benitav: Similar comment for the other cases listed here

.changeset/mssql-fixed-table-set.md
```diff
@@ -0,0 +1,28 @@
...
+  original captured schema and warns, since the bound capture instance keeps its own column list.
+- **Dropping or renaming a replicated table:** deploy a new sync configuration. Replication stops
+  with `PSYNC_S1603` — changes committed before the table went away may not have been read yet, so
+  continuing would commit past them and skip those rows. Already-replicated data for the table is
+  retained until the redeploy.
+- **Disabling and re-enabling CDC on a replicated table:** deploy a new sync configuration as a new
```
benitav: "deploy a new sync configuration as a new replication stream" - I think I pointed this out previously. Maybe this is due to some internals leaking from our incremental reprocessing work, or us not explaining those mechanics well yet, but if I read this as a user I have no idea what I need to do about the "as a new replication stream" part of this sentence. For all user-facing messaging about this shouldn't we just say "deploy new sync config", and then optionally explain how the "new replication stream" fits in?

.changeset/mssql-fixed-table-set.md
stevensJourney: > I assume that case happens when one doesn't deploy new sync config

It's not necessarily this. I agree we probably don't describe the relationship between sync config and replication streams well enough yet.

Essentially. When one deploys a sync config. That config is processed in a job which streams changes from some replication stream. 

When making changes to a table's columns we recommend doing a sync config deploy. That will start a new job which runs against the new replication stream (in this case). The old replication job continues running using the old replication stream (the previous capture table). Connected clients continue getting updates from this stream. When the new sync config is in the ready state, we make it the active sync config - causing clients to get updates from here.

docs/replication/09-resolve-tables-flow.md
```diff
@@ -86,14 +87,43 @@ Each matching pattern is resolved independently.
...
+SQL Server can keep two capture instances for one table. Each binding is pinned to the capture table's object id:
+
+- New and legacy bindings use the newest available instance.
+- A pinned binding keeps using its instance when a newer one appears and logs a warning.
+- If the pinned instance is removed, replication fails with `PSYNC_S1601`. A replacement is not adopted automatically because it may capture a different schema.
+- Conflicting persisted pins are treated as invalid state.
+
+Every configured table is resolved before streaming starts, which also backfills pins for legacy records. A table that does not exist or does not have CDC enabled fails with `PSYNC_S1602`.
+
+Adopting a new capture instance requires a new replication stream. Keep the old instance available until the new stream has finished its snapshots and is active.
```
benitav: Is this something that we should also document in the "What this means when making schema changes:" section of the release notes? Something about old capture instances must remain until reprocessing completes if I understand this correctly? 

And on that note, my AI previously pointed out that a similar restriction applies to dropping tables - apparently new sync config needs to be deployed first, then reprocessed, then table can be dropped?

.changeset/mssql-fixed-table-set.md
stevensJourney: 👍 I've updated the terminology throughout. I've added context for Agents to refer to it as "sync config".

docs/replication/09-resolve-tables-flow.md
stevensJourney: Yes, that's correct. I've updated the changeset entry to align more with the comment which mentioned this https://github.com/powersync-ja/powersync-service/pull/727#issuecomment-5202713616 

> apparently new sync config needs to be deployed first, then reprocessed, then table can be dropped

Yes, that's correct.

benitav (dismissed):

Rentacookie:
@stevensJourney I manually tested the schema changes and nearly all look to be working as expected.
Just one test case resulted in some "unexpected" behaviour: Deleting and recreating a replicating table.
For this I deleted an already replicating table `test_data` with 4 rows. Then as expected I got this log entry:
error: [powersync_1_90d7] Replication error [PSYNC_S1602] Source table dbo.test_data from the sync configuration does not exist. Create the table and enable CDC for it. {"errorData":{"code":"PSYNC_S1602"

I then did the following:
1. Stopped the powersync service. (Note that this step is not necessary)
2. Recreated the `test_data` table with CDC enabled and added 3 rows to it.
3. Restarted the powersync service without changing the sync config

The service then automatically detected the table, deleted the old table data and started re-snapshotting the table:
```
Locked replication stream for processing
info: [Flushed: 4 ops, 0 index entries, 4 records. 1kb in 6ms. Last op_id: 9 {"flushed":{"bucket_ops_count":4,"duration":6,"parameter_indexes_count":0,"size":1000,"source_records_count":4}}
info: Detected new table(s) [[dbo].[test_data]] that have not been snapshotted yet.
warn: The checkpoint LSN:[0003FFFE:0001A9D0:0006] is older than the minimum LSN:[00040005:00007EB0:000C] for table [dbo].[test_data]. This indicates that the checkpoint LSN is outside of the retention window.
```
This does it on the same replication thread so replication is paused until the data is finished deleting and the table finishes snapshotting. This results in an eventual consistent state, but diverges from the pattern of having to redeploy the sync config before schema changes are processed. But maybe that is ok in this case?

stevensJourney:
@Rentacookie 

> Just one test case resulted in some "unexpected" behaviour

I think it's probably more consistent with the new implementation to not re-snapshot in this case. The general fix for the issue is to throw if we already have a source table and detect that the actual table is gone. I had to shuffle the logic a bit to make the errors consistent between service restarts. This essentially involves checking if the table exists, and passing that as additional context to the new source table reconcillator. I've updated the changeset to reflect this behaviour.

modules/module-mssql/src/replication/CDCStream.ts
```diff
@@ -291,6 +301,21 @@ export class CDCStream {
...
+    batch: storage.BucketStorageBatch,
+    context: MSSQLUnavailableTableReconciliationContext
+  ): Promise<never> {
+    await batch.resolveTables({
+      connection_id: this.connectionId,
+      source: context.source,
+      reconcileSourceTables: createCaptureReconciler(context)
+    });
+    // Unavailable-state reconciliation must reject; this enforces the Promise<never> contract if it returns.
+    throw new ReplicationAssertionError(
```
Rentacookie, line 314: Why do we have to throw an error here? Will batch.resolveTables not already throw an error via the reconciler?

stevensJourney: We don't need to throw here. `resolveTables` _should_ always throw in this case. There is nothing enforcing that behaviour though. This throw is mainly a catch-all, which has the small benefit, as mentioned in the comment, to satisfy in TypeScript that this function never returns.

Rentacookie (approved):
The drop and recreate case now works as expected.

PR #728, opened 2026-07-24
Object Storage/S3 for bucket data

rkistner:
Based off and supersedes #673.

This moves operation storage batches to object storage (S3) for MongoDB storage V3, keeping mostly metadata shell in MongoDB. No protocol changes at this point. Object storage is optional at configuration level; when not configured, all ops remain inline in MongoDB as today. Small operations remain inline in MongoDB storage.

To start with, this can only be used with MongoDB storage, and only using the experimental V3 format. Support will be expanded to Postgres storage in the future.

Compared to #673:
1. Fixes merge conflicts.
2. Re-implements object orphan tracking and cleanup.
3. Re-implements compacting.
4. Removes zstd compression - may add configurable compression later.
5. Many small fixes and refactorings.

## Compacting

This rewrites the compaction logic to reduce churn. Previously, when even a single new operation was added to a bucket, it could lead to the boundaries for every single chunk changing, rewriting the entire bucket. Now, this optimizes to preserve current chunks in many cases:
1. Compacting MOVE operations modify the single chunk.
2. Two or more chunks can be merged into larger chunks after the above.
But we merge specific chunks - we don't re-calculate boundaries from scratch.

This means that if we have many chunks of say 600KB each, they will remain as-is, even though it could be re-chunked to get closer to the 1MB limit. This is a deliberate trade-off to avoid high churn rates.

## Orphaned data on S3

If there is even a transient failure while replicating data, it may cause orphaned files on S3. The compact job has logic to protect against that, but it's more complicated with replication. Leaving this for a future PR.

The clear logic when deleting a replication stream does remove these orphaned files.

## AI Usage

Implemented most of this using Codex gpt-5.6-sol. Reviewed using Claude Opus 5.0. Manually guided the architecture and reviewed the code.


modules/module-mongodb-storage/src/storage/implementation/v3/PersistedBatchV3.ts
```diff
@@ -218,14 +219,53 @@ export class PersistedBatchV3 extends PersistedBatch {
...
+            if (serialized.size <= this.inlineThresholdBytes) {
+              // Small enough to store inline
+              inserts.push({
+                insertOne: {
+                  document: serialized
+                }
+              });
+            } else {
+              const path = lifecycle.allocatePath(definitionId, bucket, minOp, maxOp);
+              const { fileSize } = await lifecycle.bucketData.store(path, bucketOps!);
```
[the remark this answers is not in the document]
rkistner: As mentioned in the description, this will be addressed in a future PR.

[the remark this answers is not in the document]
rkistner: Duplicate.

modules/module-mongodb-storage/src/storage/implementation/v3/MongoCompactorV3.ts
```diff
@@ -138,18 +212,19 @@ export class MongoCompactorV3 extends MongoCompactor {
     const context = { replicationStreamId: this.group_id, definitionId: resolvedDefinitionId };
 
     const lowerBound = bucketContext.minId;
-    let upperBound = bucketContext.maxId;
+    let upperBound = bucketContext.docId(this.maxOpId + 1n);
```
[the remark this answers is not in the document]
rkistner: This works by design. The next call should use a higher `maxOpId`.

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/BucketDataObjectStorage.ts
```diff
@@ -0,0 +1,54 @@
...
+): Promise<void> {
+  if (!objectStorage) {
+    return;
+  }
+
+  const store = new BucketDataObjectStorage(objectStorage);
+  const storedDocuments = documents.filter((document) => document.storage_ref);
+  await Promise.all(
+    storedDocuments.map(async (document) => {
+      signal?.throwIfAborted();
```
stevensJourney: Nit: Should we call this once before `.map()` instead? My understanding is that all the callbacks are invoked synchronously, so checking inside each callback won’t meaningfully catch an abort while iterating. The check is cheap, though; this mostly made me wonder what the intended behavior was.

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/ObjectStorage.ts
```diff
@@ -0,0 +1,25 @@
...
+    this.retryable = options.retryable;
+  }
+}
+
+export function isRetryableObjectStorageError(error: unknown): error is ObjectStorageError {
+  return error instanceof ObjectStorageError && error.retryable;
+}
+
+export interface ObjectStorage {
+  put(path: string, data: Uint8Array, metadata?: ObjectStoragePutMetadata): Promise<void>;
```
stevensJourney: Nit (opinionated) /non-blocking: Since this is a new interface, would it be worth using parameter objects here? I find call sites such as `put({ path, data, metadata })` easier to read because the argument names are visible, and it gives us more flexibility to evolve the interface. Another option would be to keep the primary inputs positional and use an options object, e.g. `get(path, { signal })`. The number of arguments here are quite small, so I don't have a very strong opinion FWIW.

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/ObjectStorageLifecycle.ts
```diff
@@ -0,0 +1,160 @@
...
+      objectCount++;
+      if (paths.length === 500) {
+        await flush();
+      }
+    }
+    await flush();
+    return { objectCount };
+  }
+
+  async cleanup(logger: { warn(message: string, error?: unknown): void }): Promise<void> {
```
stevensJourney: Nit: Could we use the shared `Logger` type here instead of declaring the `warn` signature inline?

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/S3ObjectStorage.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  async *list(prefix: string, signal?: AbortSignal): AsyncIterable<string> {
+    const fullPrefix = this.fullPath(prefix);
+    let continuationToken: string | undefined;
+
+    do {
+      signal?.throwIfAborted();
+      let response;
+      try {
+        response = await (async () => {
```
stevensJourney: Is the async IIFE needed here? It looks like we could place the `await using` and request directly inside the existing try block. The guard would still be disposed before we start yielding the response contents.

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/S3ObjectStorage.ts
```diff
@@ -0,0 +1,247 @@
...
+          Bucket: this.bucket,
+          Key: fullPath
+        }),
+        { abortSignal: signal }
+      );
+      const chunks: Uint8Array[] = [];
+      const stream = response.Body as AsyncIterable<Uint8Array>;
+      for await (const chunk of stream) {
+        signal?.throwIfAborted();
+        chunks.push(chunk);
```
stevensJourney: Could we use `response.ContentLength` to preallocate a buffer and write each chunk into it directly? That might avoid storing all the chunks and then allocating another buffer when merging them. Not sure whether ContentLength is guaranteed to be available here.

modules/module-mongodb-storage/src/storage/implementation/MongoStorageProvider.ts
```diff
@@ -53,7 +70,10 @@ export class MongoStorageProvider implements storage.StorageProvider {
       // Right now, only MongoDB source databases supports incremental reprocessing.
       // Remove this filter when we support it for other source databases.
       // This assumes a single source connection - revisit if we ever support multiple connections.
-      supportsMultipleSyncConfigs: resolvedConfig.connections?.[0]?.type == lib_mongo.MONGO_CONNECTION_TYPE
+      supportsMultipleSyncConfigs: resolvedConfig.connections?.[0]?.type == lib_mongo.MONGO_CONNECTION_TYPE,
+
+      objectStorage,
+      inlineThresholdBytes: decodedConfig.object_storage?.inline_threshold_bytes
```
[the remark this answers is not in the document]
rkistner: Noted, will address in future PR.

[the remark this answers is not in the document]
rkistner: Implemented in #735.

modules/module-mongodb-storage/test/src/storage_s3_compaction_operations.test.ts
```diff
@@ -0,0 +1,406 @@
...
+      moveBatchLimit: 10,
+      moveBatchQueryLimit: 10,
+      moveBatchByteLimit: 1024,
+      minBucketChanges: 1,
+      minChangeRatio: 0
+    });
+
+    const batchAfter = await test_utils.fromAsync(bucketStorage.getBucketDataBatch(checkpoint, [request]));
+    const dataAfter = test_utils.getBatchData(batchAfter);
+    // 1 MOVE (superseded A-v1) + 2 PUT (B, A-v2) = 3 total
```
stevensJourney: The comment here seems incorrect.

I was wondering why `A-v1` would be superseded by `A-v2` - since they have different replica IDs. 

I ran the test locally and confirmed the `dataAfter` is actually

```
[{"op_id":"1","op":"PUT","object_id":"A","checksum":586019761},{"op_id":"2","op":"PUT","object_id":"B","checksum":3910985394},{"op_id":"3","op":"PUT","object_id":"A","checksum":3143967475}]
```

So no operation is converted to MOVE. The length assertion passes in either case and therefore doesn’t verify the behavior described by the test. Should the two A operations use the same replica ID and assert their operation types?

modules/module-mongodb-storage/test/src/storage_s3_compaction_operations.test.ts
```diff
@@ -0,0 +1,406 @@
...
+      moveBatchByteLimit: 1024,
+      minBucketChanges: 1,
+      minChangeRatio: 0
+    });
+
+    const batchAfter = await test_utils.fromAsync(bucketStorage.getBucketDataBatch(checkpoint, [request]));
+    const dataAfter = test_utils.getBatchData(batchAfter);
+
+    // 3 ops for same row: 2 superseded MOVEs + 1 latest PUT.
+    // CLEAR collapses leading MOVEs, leaving 1 CLEAR + 1 PUT = 2 ops.
```
stevensJourney: The same seems to be true for this test, the `dataAfter` is actually

```
[{"op_id":"1","op":"PUT","object_id":"A","checksum":470977906},{"op_id":"2","op":"PUT","object_id":"A","checksum":849345711},{"op_id":"3","op":"PUT","object_id":"A","checksum":847656399}]
```

stevensJourney (commented):
Had a few questions and nits. Overall the general flow makes sense and looks good.

modules/module-mongodb-storage/src/storage/implementation/v3/MongoCompactorV3.ts
```diff
@@ -5,20 +5,94 @@ import { BucketDefinitionId } from '@powersync/service-sync-rules';
...
+        await this.objectStorageLifecycle.cleanup(this.logger);
+      } catch (e) {
+        // In this case, still continue normal compact process
+        this.logger.error(`Failed to clean up object storage deletion markers before compaction`, e);
+      }
+    }
+    await super.compact();
+    if (this.storage.objectStorage) {
+      // Cleanup for any produced during compacting.
+      await this.objectStorageLifecycle.cleanup(this.logger);
```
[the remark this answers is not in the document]
rkistner: That's fine.

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/ObjectStorageLifecycle.ts
```diff
@@ -0,0 +1,160 @@
...
+    const markers = this.db.pendingObjectStorageDeletes(this.replicationStreamId);
+    const cursor = markers.find({ delete_after: { $lte: new Date() } });
```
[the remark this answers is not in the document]
rkistner: Since this is only run twice per day, it's not an issue.

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/ObjectStorage.ts
rkistner: I did this partially: The `signal` is now wrapped in `options`.

For the others (path, data, metadata), where the options are required and the patterns fairly common in other APIs, I feel their usage already makes it easy enough to read.

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/S3ObjectStorage.ts
rkistner: You're right. The IIFE perhaps made sense in a previous version, but not after other refactorings were done.

modules/module-mongodb-storage/test/src/storage_s3_compaction_operations.test.ts
rkistner: Indeed, some of these tests were quite incorrect. I overhauled it a bit: Removed a bunch that duplicated testing of base compact functionality, and restructured others.

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/S3ObjectStorage.ts
rkistner: Yeah, it should be slightly more efficient (keeps overall buffer + 1 chunk in memory at a time, instead of overall buffer + all chunks). The type for ContentLength is optional, but the AWS docs indicate it's always present. The code now validates that it's present either way.

stevensJourney (approved):
Happy with the changes. Looks very cool. Excited to see S3 in action :D

PR #732, opened 2026-07-27
[MongoDB Storage] Remove some compactor abstractions

rkistner:
Previously, the `SingleBucketStore` was a helper class to handle some bucket data queries within a single bucket, covering both V1 and V3 formats within a single `MongoCompactor`. In the meantime, the V3 format and implementation diverged enough that this abstraction no longer makes sense. It now just makes the code harder to read, and adds type casts that increases the possibility of bugs. This removes the abstraction.

Additionally, `MongoCompactor` had some methods that were only used by `MongoCompactorV1`. These are moved to `MongoCompactorV1` now.

This is a pure refactoring with no behavioral changes.

This is a follow-up to #728. I identified the abstraction issues during that work, but did not want to introduce further churn in an already-large PR.

No changeset needed here.

## AI Usage

Planned and implemented using Codex gpt-5.6. Reviewed using Claude Opus 5.0 to confirm no behavior changes.


stevensJourney (approved):
This looks much nicer!

PR #734, opened 2026-07-27
fix(module-postgres-storage): delete rows in batches when clearing storage

### LINKED ISSUE #729: PostgresSyncRulesStorage.clear() deletes entire groups in single unchunked DELETEs, which never complete on large groups
Follow-up from https://github.com/orgs/powersync-ja/discussions/448.

### Problem

When a sync rules group is terminated, `PostgresSyncRulesStorage.clear()` deletes all rows for the group with one statement per table:

```sql
DELETE FROM bucket_data WHERE group_id = $1;
DELETE FROM bucket_parameters WHERE group_id = $1;
DELETE FROM current_data WHERE group_id = $1;  -- via deleteGroupRows
DELETE FROM source_tables WHERE group_id = $1;
```

On large groups this never completes. Observed on a group with 79M current_data rows (schema-per-tenant deployment, full re-replication): the DELETE runs for minutes until the connection drops (`postgres query failed` in `PgConnection.stream`), `retriedQuery` gives up after 2 attempts, `terminate()` never reaches the `state = TERMINATED` update, and `AbstractReplicator`'s refresh loop re-issues the whole thing every few seconds, indefinitely. Each attempt adds dead tuples, making the next one slower. The active stream is unaffected since the cleanup runs async, but the group's storage is never reclaimed and the logs fill with `Failed clean up replication config`.

`group_id` leads the PK on all of these tables, so it is not an indexing problem, just statement size.

### Suggested interim fix

Batch the deletes with commits in between, the way the truncate path already does (`streamTruncateRows` with `BATCH_LIMIT = 2000`). Something like a loop of

```sql
DELETE FROM current_data
WHERE ctid IN (SELECT ctid FROM current_data WHERE group_id = $1 LIMIT $2);
```

per table until 0 rows. Partitioned tables in the v4 storage (dropping the partition instead of deleting rows) would eliminate this entirely, as noted in #448; this is just the interim.

Happy to send a PR for this if useful.

henriquekraemer:
Fixes #729.

`clear()` deleted all rows for the group in a single DELETE per table. With millions of rows in `current_data` that DELETE can run for hours and gets killed by statement or socket timeouts before finishing. When this happens during `terminate()`, the replicator retries the full delete from scratch every time, so it never completes and keeps accumulating dead tuples.

Now `clear()` deletes in batches of 50k rows, selecting `ctid` by `group_id` and running each batch as its own autocommit statement. A retry continues where the previous attempt stopped. The snapshot truncate path already works this way, and this is the interim batching mentioned in #729 until the partitioned tables planned for v4.

This also implements the existing TODO for the abort signal in `clear()`: the signal is checked between batches, like the MongoDB storage does in `clearDeleteMany`.

## Tests

Added a test seeding more than one batch (`CLEAR_BATCH_LIMIT + 100` rows) plus rows for a second group. It checks that `clear()` empties all four tables for the group, leaves the other group alone, and that an aborted signal stops the operation.

### DIFF
--- a/.changeset/clear-storage-batched-deletes.md
+++ b/.changeset/clear-storage-batched-deletes.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-module-postgres-storage': patch
+---
+
+Delete rows in batches when clearing storage for a replication stream. A single DELETE covering the entire group can run for hours on large deployments and never complete once it exceeds statement or socket timeouts; batched deletes make durable progress and can be safely retried or aborted.
--- a/modules/module-postgres-storage/src/storage/PostgresSyncRulesStorage.ts
+++ b/modules/module-postgres-storage/src/storage/PostgresSyncRulesStorage.ts
@@ -49,6 +49,11 @@ export type PostgresSyncRulesStorageOptions = {
   checksumCacheTtlMs?: number;
 };
 
+/**
+ * Number of rows deleted per batch when clearing storage for a replication stream.
+ */
+export const CLEAR_BATCH_LIMIT = 50_000;
+
 export class PostgresSyncRulesStorage
   extends framework.BaseObserver<storage.SyncRulesBucketStorageListener>
   implements storage.SyncRulesBucketStorage
@@ -540,7 +545,11 @@ export class PostgresSyncRulesStorage
   }
 
   async clear(options?: storage.ClearStorageOptions): Promise<void> {
-    // TODO: Cleanly abort the cleanup when the provided signal is aborted.
+    const signal = options?.signal;
+    if (signal?.aborted) {
+      throw new framework.ReplicationAbortedError('Aborted clearing data', signal.reason);
+    }
+
     await this.db.sql`
       UPDATE sync_rules
       SET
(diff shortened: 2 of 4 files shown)


rkistner (commented):
Thanks for the PR!

Overall the approach looks good, but there is one potential issue with concurrency: If another process happens to delete data at the same time, the `SELECT count(*) as count FROM deleted` part may return fewer results than the provided limit, and prematurely stop the deletion.

Now the main cause of that would be concurrent replication processes not using a lock for clearing, which is fixed in #736. But it would still be better to handle this case on a query level.

I believe using `SELECT count(*) as count FROM batch` should protect against this case. Alternatively, a final query such as `SELECT EXISTS (SELECT 1 FROM bucket_data    WHERE group_id = $1) AS exists`, continuing if there are still rows remaining, would also work.

henriquekraemer:
> I believe using `SELECT count(*) as count FROM batch` should protect against this case.

Changed it to count from `batch` (in both the generic helper and the current_data variant). The loop now stops based on what the scan found rather than what the delete removed, so a concurrent delete can no longer end it early. Also updated the comments with the reasoning.

rkistner (approved):

PR #739, opened 2026-07-29
feat(module-postgres): configurable snapshot socket timeout

henriquekraemer:
Adds a `snapshot_socket_timeout` connection option (in seconds) for the idle timeout on snapshot connection sockets, defaulting to the current fixed 30 seconds. Same pattern as the `replication_socket_timeout` option proposed in #715, applied to the snapshot connection instead.

```yaml
replication:
  connections:
    - type: postgresql
      # ...
      snapshot_socket_timeout: 90
```

## Background

The snapshot socket timeout is an idle timeout: it fires when the snapshot loop stops consuming the source stream for 30 seconds. The main way that happens is storage backpressure, since a storage flush blocks the chunk loop. We hit this reliably while re-replicating a deployment where the new sync rules instance snapshots while the previous instance keeps streaming on the same storage: on wide tables, a flush under that combined load can take longer than 30s, and the idle timeout kills the source connection mid-snapshot. It shows up either as a socket timeout inside the chunk query or as `postgres query failed` on the next chunk, when the timeout fired between queries and destroyed the connection.

With resumable snapshots the table resumes from the last key, so no work is lost, but each kill costs the retry delay plus re-reading the current chunk. Snapshot concurrency (#731) makes the window easier to hit, but the exposure exists in the sequential path as well, since flushes have always run between chunk reads.

The default stays at 30 seconds, so behavior is unchanged unless the option is set.

## Tests

Added config normalization tests (seconds to milliseconds, unset by default, invalid values ignored), mirroring the tests in #715.


henriquekraemer:
Small update from staging: we found a second way to hit this timeout that has nothing to do with storage backpressure. Running with initial snapshot filters (a Postgres port of #502) on a large sparse table without a supporting index, a single chunk query (`WHERE <filter> ORDER BY id LIMIT 10000`) can take minutes on the server before it returns the first row. The socket sits idle the whole time, so the fixed 30s timeout kills the connection mid snapshot and the table gets stuck retrying. On a 13.6M row table with ~2% selectivity our first chunk took 8m38s.

The real fix in that case was a partial index matching the filter predicate (2.6s afterwards), but the configurable timeout is what turned an endless retry loop into something we could see and diagnose. We're running with `snapshot_socket_timeout: 90`.

rkistner (approved):

rkistner (approved):
Thanks for this. We've seen other cases where something on Postgres causes specific queries to run just over the 30-second limit, and making this configurable is a simple workaround.

It's a good note on indexes when using initial snapshot filters. When we add official support we may look into either requiring indexes on the filtered field, or producing a strong warning if there isn't one.

PR #742, opened 2026-07-31
[MSSQL] Ensure consistent ordering of CDC operations

Rentacookie:
### Summary

This fixes an issue that was discovered where deferred updates could result in missing row updates.
Deferred updates are recorded in the CDC tables as a paired `DELETE` and `INSERT` operation, with the 2 operations having the same sequence number in the same transaction. This caused problems with the ordering of operations in the same transaction and the `INSERT` could be processed before the `DELETE` causing the replicated row to disappear.

#### Changes

- Use `operation` as a tie breaker when ordering CDC results within a transaction.
- Identify deferred updates and process them as normal updates, which reduces bucket operation.
- Replaced previous optimistic transaction count estimate with an accurate count. The transactions replicated metric should now be more accurate.
- Switched to streaming for the CDC Polling query

#### Testing

Added new CDCPoller.test.ts unit tests which helped diagnose the main issue and confirms the fixes.
Moved MSSQLSourceTable test cases to own test file.

#### AI Disclosure

Used Clause Opus 5 to replicate the issue with tests and implement then implement the fix. Also reviewed and verified changeds with GPT 5.6.


modules/module-mssql/src/replication/CDCPoller.ts
```diff
@@ -248,66 +249,53 @@ export class CDCPoller {
...
+      return transactionLSNs;
     } else if (minLSN >= bounds.startLSN) {
       bounds.startLSN = minLSN;
     }
 
     try {
       const { recordset: results } = await this.connectionManager.query(
         `
-        SELECT * FROM ${table.allChangesFunction}(@from_lsn, @to_lsn, 'all update old') ORDER BY __$start_lsn, __$seqval
+        SELECT * FROM ${table.allChangesFunction}(@from_lsn, @to_lsn, 'all update old') ORDER BY __$start_lsn, __$seqval, __$operation
```
stevensJourney, line 280: Looking at this query here. I'm wondering if this should be paginated or limited/batched in some form. From the looks of it, one large transaction with many rows could be loaded into memory at-once?

I imagine we'd have to be clever about pagination if update ops are represented as multiple ops - but, I think that should be doable.

modules/module-mssql/src/replication/CDCPoller.ts
```diff
@@ -443,3 +431,110 @@ export class CDCPoller {
...
+  let currentTransactionLSN: Buffer | null = null;
+  let currentSequence: Buffer | null = null;
+
+  for (const row of rows) {
+    const nextTransactionLSN: Buffer = row.__$start_lsn;
+    const nextSequence: Buffer = row.__$seqval;
+
+    if (
+      currentRows.length > 0 &&
+      !(nextTransactionLSN.equals(currentTransactionLSN!) && nextSequence!.equals(currentSequence!))
```
stevensJourney: Nit: There's a lot of unwrapping here.

A slightly cleaner approach, at least IMO, could be to group the values into a PendingGroup object. (example was generated by Codex)

```typescript
interface PendingGroup {
  transactionLSN: Buffer;
  sequence: Buffer;
  rows: any[];
}

let current: PendingGroup | undefined;

for (const row of rows) {
  const transactionLSN: Buffer = row.__$start_lsn;
  const sequence: Buffer = row.__$seqval;

  if (
    current &&
    !(transactionLSN.equals(current.transactionLSN) && sequence.equals(current.sequence))
  ) {
    yield toLogicalChange(current.rows, current.transactionLSN, table);
    current = undefined;
  }

  current ??= {
    transactionLSN,
    sequence,
    rows: []
  };

  current.rows.push(row);
}

if (current) {
  yield toLogicalChange(current.rows, current.transactionLSN, table);
}
```

modules/module-mssql/src/replication/CDCPoller.ts
```diff
@@ -443,3 +431,110 @@ export class CDCPoller {
...
+    currentRows.push(row);
+  }
+
+  if (currentRows.length > 0) {
+    yield toLogicalChange(currentRows, currentTransactionLSN!, table);
+  }
+}
+
+function toLogicalChange(rows: any[], startLSN: Buffer, table: MSSQLSourceTable): LogicalChange {
+  // The rows are ordered by operation in the query, but this is an extra safeguard
```
stevensJourney: This seems unnecessary and like a somewhat red-flag that there might be some other potential issue elsewhere. Does anything break if we remove this?

modules/module-mssql/src/replication/CDCPoller.ts
```diff
@@ -443,3 +431,110 @@ export class CDCPoller {
...
+    rows: orderedRows
+  };
+}
+
+function resolveLogicalChangeType(orderedRows: any[], transactionLSN: LSN, table: MSSQLSourceTable): LogicalChangeType {
+  if (orderedRows.length === 1) {
+    const operation = orderedRows[0].__$operation;
+    if (operation === Operation.UPDATE_BEFORE || operation === Operation.UPDATE_AFTER) {
+      throw new ReplicationAssertionError(
+        `Incomplete update for table ${table.toQualifiedName()} in transaction LSN ${transactionLSN}: an update must have both a before and an after image.`
```
stevensJourney: For my interest sake: Can this ever happen in practice?

modules/module-mssql/src/replication/CDCPoller.ts
Rentacookie: That does look cleaner - will make the change.

modules/module-mssql/src/replication/CDCPoller.ts
Rentacookie: No should be fine to remove it, the order by operation in the original query should be good enough I reckon.

modules/module-mssql/src/replication/CDCPoller.ts
Rentacookie: No seems like this should not be possible unless we split up the rows in the logical change incorrectly

modules/module-mssql/src/replication/CDCPoller.ts
Rentacookie: True - I'll switch this to use streaming instead.(We already use that for the snapshot queries) Pagination would be quite a bit more complicated.

stevensJourney (dismissed):

packages/service-core/src/storage/SourceEntity.ts
```diff
@@ -25,8 +25,6 @@ export interface SourceEntityDescriptor extends SourceTableRef {
    * If specified, this is specifically used to detect renames.
    */
   objectId: number | string | undefined;
-  schema: string;
```
Rentacookie, line 28: Noticed that these were already declared in SourceTableRef

modules/module-mssql/test/src/CaptureReconciler.test.ts
Rentacookie, line 1: Functionally unchanged - I just moved the CDCPoller and MSSQLSourceTable test cases to their own corresponding test files

packages/service-core/src/storage/SourceEntity.ts
stevensJourney: I guess there is a subtle `readonly` declaration difference between the two. But I guess if declaring `schema` and `name` as `readonly` in `SourceEntityDescriptor` doesn't break anything, that's probably fine.

stevensJourney (approved):
LGTM

packages/service-core/src/storage/SourceEntity.ts
Rentacookie: Yeah it didn't seem to from what I could tell. Looks like we had been treating them as readonly properties to begin with.

PR #744, opened 2026-08-04
Node v24.18.1

rkistner:
Notable changes:
  * Node’s bundled SQLite moves from 3.51.3 to 3.53.1. This may have a minor effect when using the `unstable_sqlite_expression_engine` option. Specifically, SQLite 3.53 changed floating-point-to-text conversion from 15 to 17 significant digits, so expressions involving casts, concatenation, JSON, or floating-point bucket parameters may produce different strings.
 * Buffer.poolSize increases from 8 KiB to 64 KiB; allocations below 32 KiB now come from pooled
    slabs. No significant impact expected.
 * TLS, DNS, HTTP agent, certificate-store, and hostname validation were tightened across
    24.17.0–24.18.1. These could expose invalid hostnames, embedded NULs, unusual session reuse, or
    certificates chaining to removed/distrusted roots. No significant impact expected.

## AI Usage

The above description was generated with the help of Codex gpt-5.6; manually filtered.


stevensJourney (dismissed):
Happy with the changes. 

> Node’s bundled SQLite moves from 3.51.3 to 3.53.1. This may have a minor effect when using the unstable_sqlite_expression_engine option. Specifically, SQLite 3.53 changed floating-point-to-text conversion from 15 to 17 significant digits, so expressions involving casts, concatenation, JSON, or floating-point bucket parameters may produce different strings.

I guess `unstable_sqlite_expression_engine`, is by definition `unstable`, so this seems like not a large concern :D . I'm not sure if we need to mention this in the Changesets to be reflected in the release notes, but it seems like we should mention this potentially in the release notes.

stevensJourney (approved):

PR #745, opened 2026-08-04
TypeScript 7 [BLOCKED]

rkistner:
_Blocked until typescript 7.1.0 - see comments._

This upgrades to TypeScript 7.

Build times on my machine, running `tsc -b --clean` then `time tsc -b`:

```
# Before:
tsc -b  14.86s user 0.70s system 172% cpu 9.020 total
# After:
tsc -b  9.07s user 0.74s system 748% cpu 1.312 total
```

The main changes appear to be from increasing the concurrency, but it does also have a reduction in overall CPU usage.

This also removes nodemon and ts-node, since ts-node is not directly compatible with typescript 7. Possible options:
1. Just remove the `pnpm run watch` command (current PR state).
2. tsx _might_ be compatible with typescript 7 as a drop-in replacement.
3. Install typescript 6.x next to 7.x.
4. Use `node --watch` for restarting the process, along with `tsc -b -w` in a separate tab.

No changeset since this does not change the build output.

## AI Usage

None.

### DIFF
--- a/modules/module-mongodb/tsconfig.scripts.json
+++ b/modules/module-mongodb/tsconfig.scripts.json
@@ -8,5 +8,6 @@
     "composite": false,
     "types": ["node"]
   },
-  "include": ["scripts/**/*.mts"]
+  "include": ["scripts/**/*.mts"],
+  "references": [{ "path": "./" }]
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -16,8 +16,8 @@ catalogs:
       specifier: ^4.1.5
       version: 4.1.5
     typescript:
-      specifier: ~6.0.3
-      version: 6.0.3
+      specifier: ~7.0.2
+      version: 7.0.2
     uuid:
       specifier: ^14.0.0
       version: 14.0.0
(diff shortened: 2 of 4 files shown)


rkistner:
@stevensJourney I believe you originally added the `pnpm watch` command - are you still actively using it? If not, I think the simplest option is to just remove it.

Rentacookie:
> @stevensJourney I believe you originally added the pnpm watch command - are you still actively using it? If not, I think the simplest option is to just remove it.

I still use the watch command to start up the service locally. But if there is an alternative to that I'd gladly use it.

rkistner:
It looks like in theory `node --watch` along with a separate `tsc -b -w` can work quite well. @Rentacookie would that be fine for you (needing two commands instead of one)?

That said, during testing I found that `tsc -b -w` is unusable on my machine right now, blocking this PR. That should be resolved in 7.1.0, so we can wait until then before merging this: https://github.com/microsoft/TypeScript/issues/63867

Rentacookie:
> It looks like in theory `node --watch` along with a separate `tsc -b -w` can work quite well. @Rentacookie would that be fine for you (needing two commands instead of one)?
> 
> That said, during testing I found that `tsc -b -w` is unusable on my machine right now, blocking this PR. That should be resolved in 7.1.0, so we can wait until then before merging this: [microsoft/TypeScript#63867](https://github.com/microsoft/TypeScript/issues/63867)

That should be fine then. I am not stressed about auto triggering a rebuild for file changes while running the service. 🙇

stevensJourney:
> @stevensJourney I believe you originally added the `pnpm watch` command - are you still actively using it? If not, I think the simplest option is to just remove it.

I do still use the watch command in my dev setup - it's quite convenient, but I'm happy to work around it. 

I think we already use `concurrently` in the `watch:service` command, so I _think_ we could run both `node --watch` and `tsc -b -w` in the single script. I'm also happy if I have to run two commands.

PR #746, opened 2026-08-04
Improve `Flushed X + Y + Z` log message

LucDeCaf:
Updates the `Flushed X + Y + Z updates` log message in `PersistedBatch.ts` to include better labelling of data. Will be accompanied by a PR in `powersync-docs` to elucidate what the different types of updates mean.

### Example

Since this log can potentially occur a lot on a high-load instance, I injected some fake data into a local `powersync-dashboard` app to see what it could look like.
If this proves to be too noisy, I'm happy to strip back on the LOD - but the old log message was very obscure.

#### With `Metadata` enabled

Screenshot 2026-08-04 at 11 56 50

#### Without `Metadata` enabled

Screenshot 2026-08-04 at 12 01 58

### AI Usage

I instructed Claude Opus 5 to inject the mock data into the `powersync-dashboard`. The data does not come from any real instances/provisions.


rkistner:
I think the level of detail is good here, but we should tweak the naming.

1. "95 bucket" - could be misinterpreted as "updated 95 buckets". Maybe something like "95 ops"?
2. "8 parameter" - new versions use "parameter indexes" for the overall concept, but even there the "parameter" part can be confusing.
3. "46 current" - in newer storage versions I renamed "current data" to "source records", which more accurately represents what it is used for.

How does this sound?

```
Flushed: 95 ops, 8 index entries, 46 records
```

We can also consider updating the metadata accordingly.

LucDeCaf:
> How does this sound?
> 
> ```
> Flushed: 95 ops, 8 index entries, 46 records
> ```

I like this better too. Does changing the metadata field names have any backwards-compat concerns?

rkistner (dismissed):

rkistner:
Just fix the formatting changes here: `pnpm format`.

You can also configure hooks automatically check formatting on each commit: `pnpm configure-hooks`

rkistner (approved):

PR #750, opened 2026-08-12
[MSSQL] Add custom CA certificate support for MSSQL connections

Rentacookie:
### Summary

This allows users to specify a CA certificate in the connection config, to be used when opening connections to the SQL Server source DB. Encryption on the connection was always enabled, but for non Azure SQL Server DBs this required the `trustServerCertificate` flag to be set to true, which accepts self signed certificates.

#### Changes

Added the `cacert` and `tls_servername` fields to the MSSQLConnectionConfig. If present those values are then used when opening connections to the SQL server database.
`tls_servername` need only be specified when the DB hostname PowerSync connects to differs from the DNS name in the SQL Server certificate.  For example with the connection config:
```
hostname: 10.0.1.25
tls_servername: sql-prod.internal.example
```
PowerSync connects to 10.0.1.25, but validates that the certificate is valid for sql-prod.internal.example.

#### AI Disclaimer

Used GPT 5.6 to investigate feasibility and then add the required cacert support. 
Manually verified functionality by connecting the PowerSync service to a SQL Server with a specified cacert.

### DIFF
--- a/.changeset/mssql-add-cacert-support.md
+++ b/.changeset/mssql-add-cacert-support.md
@@ -1,0 +1,9 @@
+---
+'@powersync/service-module-mssql': minor
+---
+
+Add custom CA certificate support for MSSQL connections.
+
+The new `cacert` option allows PowerSync to validate SQL Server certificates issued by a private CA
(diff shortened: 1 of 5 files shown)


rkistner (approved):
Looks good to me

PR #760, opened 2026-08-20
[MySQL] Fix replication stalls caused by silently dropped idle connections

bean1352:
## Summary

The PowerSync Service keeps two connections open to MySQL: one that streams changes, and a helper connection used for occasional lookups. The helper can sit idle for hours. Some firewalls quietly close connections that have been idle for an hour. The next time the PowerSync Service tries to use that closed connection, it waits ~15 minutes for the operating system to give up. During that wait replication is completely frozen, nothing is logged, and the PowerSync Service still reports itself as healthy. Shutdown has the same problem, because it sends one last query over that same connection and waits for the answer with no time limit. A self-hosted customer hit this issue five times in two days and traced it to their firewall's one hour idle timeout.

## Fix

The MySQL connections now send a `keepalive` signal every 40 seconds. That is enough for a firewall to see the connection as active, so it never gets closed in the first place. Shutdown also gets a 5 second limit: if the final query gets no answer in that time, the service closes the connection itself instead of waiting.

## AI disclaimer

I developed this change using Claude Opus 4.8 for the investigation and Claude Fable 5 for the implementation. I reviewed and tested the changes myself.


rkistner (commented):
To clarify, this specifically enables _TCP_ keepalive, not a mysql-specific keepalive?

Is there an option on MySQL / Zongji to additionally do a mysql-level keepalive? If there is no built-in functionality for that, maybe something along the lines of a `SELECT 1` query every minute?

Your work here is definitely an improvement, but the firewall silently dropping the connection due to being idle is not the only case this can happen, and TCP keepalives typically take very long to detect the issue on their own. A protocol-level keepalive with a socketTimeout or equivalent can be a much stronger check.

bean1352:
> To clarify, this specifically enables _TCP_ keepalive, not a mysql-specific keepalive?
> 
> Is there an option on MySQL / Zongji to additionally do a mysql-level keepalive? If there is no built-in functionality for that, maybe something along the lines of a `SELECT 1` query every minute?
> 
> Your work here is definitely an improvement, but the firewall silently dropping the connection due to being idle is not the only case this can happen, and TCP keepalives typically take very long to detect the issue on their own. A protocol-level keepalive with a socketTimeout or equivalent can be a much stronger check.

Yes the change was TCP `keepalive` only. There doesn't seem to be a built-in mysql-level `keepalive` in `zongji` or `@vlasky/mysql`, so I added one along the lines you suggested.

The service now creates the control connection itself and passes it into `zongji`, then runs `SELECT 1 `on it once a minute with a 5 second timeout. If the probe gets no answer, the listener destroys the connection and stops with an error, which goes through the normal restart path.

I'll also open a small types PR to `powersync-mysql-zongji`. PR: https://github.com/powersync-ja/powersync-mysql-zongji/pull/19

modules/module-mysql/src/replication/zongji/BinLogListener.ts
```diff
@@ -243,16 +286,61 @@ export class BinLogListener {
...
+
+  private async probeControlConnection(): Promise<void> {
+    const controlConnection = this.controlConnection;
+    if (this.zongji.stopped) {
+      return;
+    }
+    const timeout = this.options.ctrlConnectionKeepAliveTimeoutMs ?? CTRL_CONNECTION_KEEPALIVE_TIMEOUT;
+    const responded = await new Promise<boolean>((resolve) => {
+      controlConnection.query('SELECT 1', (error) => resolve(!error));
+      timers.setTimeout(timeout, undefined, { ref: false }).then(() => resolve(false));
```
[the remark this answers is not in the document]
rkistner: Using a separate dedicated connection would defeat the purpose of the probe. But apart from that this makes sense - need to take into account the case where the connection is already waiting on another query.

modules/module-mysql/src/replication/zongji/BinLogListener.ts
```diff
@@ -243,16 +286,61 @@ export class BinLogListener {
...
+    let idleTime = 0;
+    while (!this.isStopped) {
+      await timers.setTimeout(1_000);
+      idleTime += 1_000;
+      if (idleTime >= interval && !(this.isStopped || this.isStopping)) {
+        idleTime = 0;
+        await this.probeControlConnection();
+      }
+    }
+  }
```
rkistner: Instead of having a separate mechanism here, can this use the `async keepAlive()` trigger in `BinLogReplicationJob`?

bean1352: Yes, thanks for that. I removed the interval loop, the probe runs off of `keepAlive()`.

rkistner (commented):
Looks good to me; just one comment on connection error handling.

It would be nice to get the relevant types upstream in our `@powersync/mysql-zongji` package, but it's not a blocker here.

modules/module-mysql/src/replication/MySQLConnectionManager.ts
```diff
@@ -46,22 +58,33 @@ export class MySQLConnectionManager extends BaseObserver<MySQLConnectionManagerL
...
+      // connection can be silently dropped by stateful firewalls, freezing replication on the
+      // next table metadata query until the TCP retransmission timeout (~950s).
+      enableKeepAlive: true,
+      keepAliveInitialDelay: mysql_utils.TCP_KEEPALIVE_INITIAL_DELAY,
       // We want to avoid parsing date/time values to Date, because that drops sub-millisecond precision.
       dateStrings: true,
       timeZone: 'Z'
     });
+    // The published ZongjiOptions type does not cover passing in an existing connection yet.
+    const listener = new ZongJi(controlConnection as unknown as ZongjiOptions);
```
rkistner: From Codex:
> external control-connection errors can crash the process: modules/module-mysql/src/replication/MySQLConnectionManager.ts:82 passes an already-created connection to Zongji. In that code path, Zongji does not attach its normal error / unhandledError forwarding listeners to the control connection. If the idle control socket is closed and emits an error while no query is pending—the failure this PR targets—Node treats the unhandled error event as fatal. Attach handlers to controlConnection that forward to listener.emit('error', error) (or handle/restart it directly) before returning it.

Adding these two lines should cover it, but please confirm:
```js
controlConnection.on('error', listener.emit.bind(listener, 'error'));
controlConnection.on('unhandledError', listener.emit.bind(listener, 'error'));
```

.changeset/mysql-tcp-keepalive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/service-module-mysql': patch
+---
+
+Enable TCP keepalive on the Zongji binlog and control connections and on the connection pool, so that idle connections are no longer silently dropped by stateful firewalls, which froze replication for ~950 seconds per occurrence with no error or health signal. Also add a periodic liveness probe on the control connection that restarts replication within about a minute when the connection stops responding, and bound the BinLog listener stop sequence so that a dead control connection cannot stall shutdown.
```
rkistner:
```suggestion
Enable TCP keepalive on MySQL connections, as well as a periodic liveness probe on the control connection, so that idle connections are no longer silently dropped by stateful firewalls.
```

Rentacookie (approved):
LGTM - thank you for handling the cleanups! 🙏

PR #762, opened 2026-08-21
feat: Use Compiler for Event Definitions

stevensJourney:
**Part 1 of 3**

## Summary

Replication events defined through `event_definitions` are currently parsed and evaluated by a separate legacy implementation. Sync streams and buckets already use the sync-plan compiler.

This PR compiles replication events into the sync plan and evaluates them through the same row-projection machinery used by streams. It also adds behavioral equality for persisted serialized event plans, which the next PR will use to decide whether an existing event mapping can be reused during incremental reprocessing.

This PR does not assign persisted event ids or change incremental-reprocessing behavior. Those storage concerns remain in part 2.

## Motivation

Replication events map rows from a source table to payloads consumed by the service. Custom checkpoint requests are one use of these events:

```yaml
event_definitions:
  checkpoint_requests:
    payloads:
      - SELECT user_id, checkpoint, client_id FROM checkpoints
```

Incremental reprocessing needs to determine whether a newly deployed event definition is compatible with the event plan that produced existing stored data. Comparing raw SQL would treat formatting-only edits as changes. For example, these definitions serialize to the same event behavior:

```sql
SELECT user_id, checkpoint, client_id
FROM checkpoints
WHERE active = TRUE AND checkpoint > 0
```

```sql
select "user_id", "checkpoint", "client_id"
from "checkpoints"
where "active" = true and "checkpoint" > 0
```

The comparison therefore uses the persisted serialized event plans rather than raw SQL or newly compiled JavaScript models. This compares the evaluator that produced the existing data with the evaluator from the new deployment, so a later compiler change cannot be hidden by recompiling the retained SQL with the current compiler.

## What changed

`event_definitions` are now compiled for every supported sync-config edition and stored in an additive `events` field in the serialized sync plan. Compiled events use the shared row-projection implementation, allowing the separate `SqlEventDescriptor` and `SqlEventSourceQuery` evaluators to be removed.

Serialized event equality includes the event name, source tables, filters, projected payload columns, table-valued functions and partitioning data. Raw SQL is excluded because it is retained only as a rolling-upgrade compatibility mirror. Cached evaluator hashes are also excluded because they are not complete equality checks.

Payload-query order is ignored because runtime selects a query by its source table. Normalized variants are also compared without regard to order because they are alternative matching branches with the same projection.

Comparison within each serialized evaluator remains deliberately conservative. Reordering expression operands, filters or projected columns may cause an event to be treated as changed even when some of those changes are semantically equivalent. This may trigger unnecessary reprocessing, but it cannot incorrectly reuse incompatible data and matches the conservative compatibility behavior used for stream plans.

The existing event-query restrictions remain. Each payload query must read from one physical source table and cannot depend on request parameters. An event can define at most one payload query for a given source table.

## Persisted-plan compatibility

Compiled events are an additive sync-plan field and do not require a plan-version bump.

The service continues to dual-write the raw event SQL alongside the compiled plan so older service versions can ignore the new field and use their legacy event evaluator. When a new service loads an older persisted plan without compiled events, it compiles that sidecar SQL at the loading boundary and uses the compiled evaluator in memory.

## Follow-up

Part 2 uses serialized event-plan equality to assign or reuse stream-local event ids, store those ids as source-table memberships, and resnapshot only the source tables associated with new or changed event definitions.

## Validation

Tests cover event compilation, serialization, deserialization, hydration and row evaluation. They verify equality across formatting and payload-query ordering, while confirming that changes to event names, source bindings, literals, projected payload aliases and serialized evaluator behavior are detected. They also document the conservative behavior for reordered operands, filters and projected columns.

Compatibility tests cover both newly compiled plans and older plans containing only raw event descriptors. Validation tests cover unsupported joins, request parameters and duplicate source tables.

---


packages/sync-rules/src/compiler/compiler.ts
```diff
@@ -34,21 +37,84 @@ export interface ParseStreamOptions extends StreamOptions {
...
+        JSON.stringify([
+          tablePatternIdentity(query.sourceTable),
+          query.variants.map(compiledEventVariantIdentity).sort()
+        ])
+      )
+      .sort()
+  ]);
+}
+
+function compiledEventVariantIdentity(variant: EventRowEvaluator): string {
```
simolus3: This is only used to derive a hash code and equality, which `EventRowEvaluator` can already do natively without requiring JSON serialization.

I think it would be easier to express `CompiledEvent` and `CompiledEventSourceQueryModel` as classes implementing `Equatable` by comparing members (and using a `listEquality` / `unorderedEquality` for nested fields).

packages/sync-rules/src/compiler/compiler.ts
```diff
@@ -124,11 +190,173 @@ export class SyncStreamsCompiler {
...
+  options: SyncStreamsCompilerOptions
+): { events: CompiledEventDescriptor[]; errors: SqlRuleError[] } {
+  const { compiler, errors } = compileEventDefinitionsIntoCompiler(definitions, options);
+  return { events: compiler.toSyncPlan().events, errors };
+}
+
+/**
+ * Compiles raw event SQL and retains the compiler's JavaScript model for behavioral compatibility checks.
+ */
+export function compileEventDefinitionsToCompilerModel(
```
simolus3: This is only used in tests, so it should be a test utility? Or do we intend to use it internally for PowerSync cloud?

packages/sync-rules/src/compiler/expression.ts
```diff
@@ -77,6 +77,100 @@ export class SyncExpression implements EqualsIgnoringResultSet {
   }
 }
 
+/**
+ * A stable identity for expression behavior across syntactic operand ordering changes.
+ *
+ * Boolean conjunction/disjunction and equality are commutative. Comparisons are normalized to one direction, so
+ * `a > b` and `b < a` also match. Operators and constructs where order affects behavior retain their original order.
+ */
+export function expressionBehaviorIdentity(expression: SqlExpression<ExpressionInput>): string {
```
simolus3: It would be nice if we didn't need this. If we do, using `ExpressionToSqlite.toSqlite` to get a unique string for the AST shape and then stringifying data instantiation separately (see also `SyncExpression.equalsAssumingSameResultSet`) avoids having to branch for every expression type again.

simolus3 (commented):
It's exciting to see the compiler used for event definitions as well :)

I have left some minor drive-by comments. Another concern I remember for bucket definitions (there was an idea to use the new compiler for that too) is that the compiler/sync plan doesn't behave in exactly the same way as the legacy system. I believe this was primarily a concern about bucket parameters which cannot change after a service upgrade, something that is irrelevant for events. But handling event definitions is largely untested in the sync rules package itself, so it's difficult to spot behavioral differences.

I don't think opting in to this should require a sync config option or a new edition, but it's something to be aware of. Maybe it makes sense to also test this with events configured on PS cloud instances, to make sure it at least parses them without errors?

packages/sync-rules/src/compiler/compiler.ts
stevensJourney: Thanks! I agree that implementing `Equatable` would be cleaner if we needed to compare the compiled event models here. After looking into it, I realised that this is not actually needed in the follow-up PRs.

The equality needed by the follow-up incremental-events PR is instead between the persisted serialized event plans, similar to [serializedStreamBucketDataSourceEquality](https://github.com/powersync-ja/powersync-service/blob/d3216be9f1f285fd4f2622a7d987ebaf4c2b8e0c/packages/sync-rules/src/sync_plan/plan_equality_serialized.ts#L24-L40) used for incremental bucket mappings. Comparing the serialized plans ensures we compare the evaluator that produced the existing stored data, rather than recompiling the SQL with the current compiler.

I’ve therefore removed the compiled-event equality and kept `CompiledEvent` and `CompiledEventSourceQueryModel` as interfaces for now

packages/sync-rules/src/compiler/compiler.ts
stevensJourney: Good point! This has been removed, since it was not needed.

packages/sync-rules/src/compiler/expression.ts
stevensJourney: Thanks, agreed. I think I was a bit overzealous here. The goal was to avoid reprocessing when the compiled plan changed only because semantically equivalent operands had been reordered, such as `a = b` becoming `b = a`.
That is an unnecessary optimization rather than a correctness requirement. We can compare the persisted serialized plans directly and conservatively treat some ordering changes as incompatible. The worst case is that we assign a new event mapping and reprocess when it may not have been strictly necessary, which is the same conservative matching behaviour we already accept for stream plans.
I’ve removed this additional expression-normalization layer in favour of the serialized plan comparison.

stevensJourney:
> It's exciting to see the compiler used for event definitions as well :)
> 
> I have left some minor drive-by comments. Another concern I remember for bucket definitions (there was an idea to use the new compiler for that too) is that the compiler/sync plan doesn't behave in exactly the same way as the legacy system. I believe this was primarily a concern about bucket parameters which cannot change after a service upgrade, something that is irrelevant for events. But handling event definitions is largely untested in the sync rules package itself, so it's difficult to spot behavioral differences.
> 
> I don't think opting in to this should require a sync config option or a new edition, but it's something to be aware of. Maybe it makes sense to also test this with events configured on PS cloud instances, to make sure it at least parses them without errors?

Thanks! There are some compatibility differences between the legacy implementation and the compiler. In practice, though, events are currently only used for custom write checkpoints by a small number of customers, and the queries we recommend are simple projections that don’t encounter those differences.

I checked all custom write checkpoint event definitions currently deployed in PowerSync Cloud. They use our documented examples exactly, so they parse successfully and are safe with the compiler.

One large difference is that the legacy implementation ignored `WHERE` filters. Supporting filters could be useful in future. I don’t think this warrants a new edition or an explicit opt-in. To avoid breaking the (very) unlikely case where someone relied on a filter being ignored, existing deployments retain the legacy behavior and emit a warning. Deploying a new sync config opts into the compiled behavior, where the filter is applied.

packages/sync-rules/src/sync_plan/serialize.ts
```diff
@@ -193,8 +241,16 @@ export function serializeSyncPlan(plan: SyncPlan): SerializedSyncPlan {
...
       queriers: s.queriers.map(serializeStreamQuerier)
     })),
-    version: usesRowMetadataSqlValue ? 2 : 1
+    version: tableProcessorSerializer.usesRowMetadataSqlValue ? 2 : 1
   };
+
+  // Compiled events are intentionally additive to plan versions 1 and 2. The service also persists their raw SQL in
+  // the legacy eventDescriptors field, so older readers can ignore this field and retain equivalent event behavior.
+  if (events.length != 0) {
+    serialized.events = events;
```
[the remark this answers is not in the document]
stevensJourney: I agree the raw SQL mirror isn’t equivalent for WHERE filters. Given the very narrow use-case, I’d prefer to document this rollback limitation rather than introduce a new plan version. The changeset now notes that event `WHERE` clauses should be removed before rolling back to an older service version.

simolus3 (approved):
I am not too familiar with the events system so it might be good to also let Ralf take a look. But this makes sense to me and I'm happy with the compiler changes.

rkistner (approved):
I'm happy with the event system changes.

PR #764, opened 2026-08-21
feat: Incremental Custom Checkpoints

stevensJourney:
**Part 3 of 3**

## Summary

Part 2 gives each replication event a stream-local id and ties that id to the sync configs and source tables that use it. This PR uses that association to scope MongoDB storage v3 custom checkpoint requests to the exact event definition that produced them.

Custom checkpoints from active and processing sync configs can now coexist during an incremental deployment. Client reads continue to use the collection associated with the active sync config until the replacement config becomes active.

## Motivation

MongoDB previously stored custom checkpoints for all replication streams in one shared collection, with records distinguished by replication stream id. That is not sufficient when an active sync config and its replacement are being processed in the same stream at the same time.

If the checkpoint event changes, Part 2 assigns the replacement definition a new event id and snapshots it separately. The processing config may begin producing custom checkpoints before it is active. Reading the greatest checkpoint across the stream would expose that processing state to clients still served by the old config.

The read must instead follow the active sync config's mapping. An unchanged event can safely share its existing collection, while a changed event needs a separate collection until activation.

## Selecting the checkpoint event

Custom checkpoint mode now accepts an event name through `setWriteCheckpointMode({ mode: CUSTOM, eventName })`. The integration supplies the configured event name rather than an event id or resolver. MongoDB storage v3 resolves that name through the active sync config's persisted mapping when it reads custom checkpoint state.

This keeps storage-assigned ids internal. Integrations do not derive ids from SQL or compiled event content, and a later change to event comparison does not change the external configuration contract.

MongoDB storage v3 requires an event name in custom mode. Missing event names are rejected when the mode is configured, while unknown names are rejected when checkpoint state is read. One event owns custom checkpoints for a storage instance, so an integration using this mode must use the same configured event for its custom checkpoint flow.

## Writing scoped checkpoints

Incremental replication event payloads now include the assigned `event_id`. An event handler copies that value to `addCustomWriteCheckpoint()`, allowing the write path to route the checkpoint to the event version that actually fired. MongoDB storage v3 rejects missing or unknown event ids.

Each event stores its records in `custom_checkpoint_requests_${stream_id}_${event_id}`. A collection contains the latest custom checkpoint for each user for that event definition. Collections and their user, operation-id, and retention indexes are created lazily, so unrelated events do not create empty checkpoint collections.

Compatible event definitions reuse the same assigned id and collection across deployments. A changed definition receives a new id and writes to a separate collection while both sync configs coexist.

## Reading through the active sync config

Read-side storage instances are constructed for the active sync config only. MongoDB storage v3 lazily resolves the configured event name through that instance's persisted single-config mapping and reads only the corresponding event collection.

This means checkpoints written by a changed processing event remain invisible to clients until the new sync config is activated. After activation, new reads resolve the same configured event name through the replacement mapping and begin reading its collection. When the event is unchanged, both mappings resolve to the reused id and existing checkpoint state remains available without a transition.

## Lifecycle and compatibility

Request-derived custom checkpoints retain their `checkpoint_requested_at` metadata. MongoDB v3 compaction applies the existing retention policy across the per-event collections, while persistent source-owned checkpoints remain unaffected.

Stopped-config cleanup drops a custom checkpoint collection when no live sync config uses its event id. Clearing a replication stream also removes all of its scoped custom checkpoint collections.

The per-event collection model applies only to MongoDB storage v3. Legacy MongoDB storage versions and Postgres continue using their existing unscoped custom checkpoint storage. Managed checkpoint mode is unchanged.

## Validation

Tests verify independent writes and reads for multiple event definitions, lazy collection and index creation, missing and unknown event validation, and idempotent setup after a service restart. They also verify that an unchanged event retains its checkpoint state across deployment, while a changed processing event remains isolated until activation.

Additional coverage verifies checkpoint-change streaming, request-retention compaction, stopped-config cleanup, full storage clearing, and unchanged behavior for legacy storage implementations.

---


modules/module-mongodb-storage/src/storage/implementation/v3/MongoWriteCheckpointAPIV3.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  protected resolveEventId(syncConfig: HydratedSyncConfig): EventDefinitionId {
+    const eventName = this._customWriteCheckpointEventName;
+    if (!eventName) {
+      throw new ServiceAssertionError(`No eventName has been supplied via setWriteCheckpointMode.`);
+    }
+    if (!syncConfig.eventDescriptors.some((event) => event.name == eventName)) {
+      throw new ServiceAssertionError(`Unknown custom checkpoint event definition ${eventName}`);
+    }
```
rkistner: `syncConfig` is now added on `WatchWriteCheckpointOptions`, but we should already have the data we need on `MongoSyncBucketStorage`. The only place I can see the `syncConfig` arg being used is here, but the checks appear redundant with the checks already present in `this.syncConfigMapping().eventDefinitionIdByName`

rkistner (commented):
The new storage structure and queries look good to me

stevensJourney:
Scoping custom checkpoint requests by replication stream and event ID has one caveat: each sync config can use only one event to produce custom checkpoints.

This affects applications migrating from legacy custom write checkpoints, currently identified by a `write_checkpoints` event, to custom checkpoint requests, currently identified by a `checkpoint_requests` event. During a rolling rollout, older app versions may still use legacy checkpoints while newer versions use checkpoint requests.

Legacy checkpoint records must not expire because older clients do not retry them automatically. The external custom write checkpoints module currently determines retention from the event name, expiring records produced by `checkpoint_requests` but retaining those from `write_checkpoints`. Since the updated storage model supports only one configured checkpoint event, using separate events makes it difficult to support both flows during a rollout.

The updated recommendation is therefore to use a single `checkpoint_requests` event and include an `is_legacy` field in each payload. The module can then use that field to determine whether the record may expire.

One additional detail is that this introduces an `is_legacy` field to identify the type of checkpoint record. We could infer this from `checkpoint_requested_at`, but an explicit field communicates the retention semantics more clearly.

For example, records from one table could be distinguished using `checkpoint_requested_at`:

```yaml
checkpoint_requests:
  payloads:
    - SELECT user_id, checkpoint, client_id,
        CASE WHEN checkpoint_requested_at IS NULL
          THEN true
          ELSE false
        END AS is_legacy
      FROM checkpoints
```

Alternatively, separate source tables could set the field explicitly:

```yaml
checkpoint_requests:
  payloads:
    - SELECT user_id, checkpoint, client_id, true AS is_legacy
      FROM legacy_checkpoints
    - SELECT user_id, checkpoint, client_id, false AS is_legacy
      FROM checkpoint_requests
```

modules/module-mongodb-storage/src/storage/implementation/v3/MongoWriteCheckpointAPIV3.ts
stevensJourney: Good catch! I've remove this.

stevensJourney:
I've tested this by locally building the external Custom checkpoints module and temporarily updating the custom checkpoints self hosted demo to use MonogDB V3 incremental reprocessing. 

I verified both that a combination of legacy and checkpoint requests can function correctly simultaneously and that incremental reprocessing does not reprocess unchanged checkpoints events.

image

rkistner (approved):

rkistner (approved):

PR #770, opened 2026-08-25
S3 storage stats

rkistner:
This records storage stats for S3.

With MongoDB and Postgres, we can cheaply get stats by querying the collection/table, getting metrics with the same granularity as the collection/table. This now exploits it for MongoDB v3 storage: This now reports storage sizes per sync config, in addition to the global metrics. This makes it easy to view metrics for configs that are active, replicating, or deleting.

For S3, it's more tricky. There is no native S3 API that gives aggregated storage size. AWS does provide:
1. ListObject: Iterate through all S3 entries. Does not scale well.
2. CloudWatch metrics for total object size and count per bucket, updated daily. No way to split based on prefix; no way to get more recent data; no way to limit access to specific CloudWatch metrics via policies.
3. S3 Storage Lens: Can give aggregated metrics per bucket or per prefix. Requires extra work to configure.
4. S3 Inventory: Provides a summary of metadata over all S3 files in a bucket, updated daily or hourly. Slightly easier and faster than ListObject calls, but still does not scale well, and requires extra work to configure.

We could also iterate through bucket_data documents, or bucket_state documents, but neither scales well to hundreds of millions of buckets.

This instead updates aggregate counters when writing. By using the same transaction that performs the write, we ensure the numbers stay accurate. As a core requirement, we split the numbers by replication stream and by bucket definition id.

We could store the numbers on the existing `sync_rules` document, or an equivalent document per replication stream or bucket definition. The core issue that has is write contention between the replication process and compact workers. Since both would update the same document, it can easily lead to high write contention, completely stalling replication in the worst case.

This implementation works around it by "sharding" the counters based on the writer: Each writer generates a random writer id, and scopes all writes to that, completely avoiding write contention. As long as the number of writers stay small, the number of documents stay small, and reads stay cheap. But since every process restart or compact run can generate a new id, we do fold up these stats after each compact job.

## Metrics

Example Prometheus metrics below. We now produce metrics for each individual sync config. Note that if they share storage (shared definitions; shared source tables), the storage will be double-counted. The goal here for these metrics is to get a stable value per sync config, not to track overall usage.

```
# Old storage metrics: unchanged
powersync_replication_storage_size_bytes 7340032
powersync_operation_storage_size_bytes 2097152
powersync_parameter_storage_size_bytes 524288

# Added: total object storage size
powersync_object_storage_size_bytes 1048576

# Added: Metrics attributed to sync configs
# Note that these may add up to less than the totals if V1 storage is still used, and more than the totals where sync configs share storage. This is intentional.
powersync_attributed_source_records_bytes{sync_config_id="65f...",sync_config_state="ACTIVE"} 5242880
powersync_attributed_source_records_bytes{sync_config_id="65a...",sync_config_state="PROCESSING"}
4194304

powersync_attributed_bucket_data_bytes{sync_config_id="65f...",sync_config_state="ACTIVE"} 1572864
powersync_attributed_bucket_data_bytes{sync_config_id="65a...",sync_config_state="PROCESSING"} 1310720

powersync_attributed_parameter_indexes_bytes{sync_config_id="65f...",sync_config_state="ACTIVE"} 262144
powersync_attributed_parameter_indexes_bytes{sync_config_id="65a...",sync_config_state="PROCESSING"}
131072

powersync_attributed_object_storage_bytes{sync_config_id="65f...",sync_config_state="ACTIVE"} 786432
powersync_attributed_object_storage_bytes{sync_config_id="65a...",sync_config_state="PROCESSING"} 524288
```

## AI Usage

Planned and implemented using Codex gpt-5.6. Manually reviewed and tweaked.

### DIFF
--- a/modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
+++ b/modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
@@ -783,7 +783,7 @@ export class MongoBucketStorage extends storage.BucketStorageFactory {
 
   async getStorageMetrics(): Promise<storage.StorageMetrics> {
     const ignoreNotExisting = (e: unknown) => {
-      if (lib_mongo.isMongoServerError(e) && e.codeName == 'NamespaceNotFound') {
+      if (lib_mongo.isMongoNamespaceNotFoundError(e)) {
         // Collection doesn't exist - return 0
         return [{ storageStats: { size: 0 } }];
       } else {
--- a/modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
+++ b/modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
@@ -794,9 +794,6 @@ export class MongoBucketStorage extends storage.BucketStorageFactory {
       }
     };
 
-    // For now, we get storage metrics over all v1 and v3 collections.
-    // In the future, we may split these metrics to report separately for active replication streams versus processing streams.
-
     const aggregateStaticCollection = async <T extends mongo.Document>(collection: mongo.Collection<T>) => {
       // We check whether the collection exists before getting the statistics. This avoids repeated
       // errors in the MongoDB logs if the collection hasn't been created yet.
@@ -820,8 +817,9 @@ export class MongoBucketStorage extends storage.BucketStorageFactory {
     };
 
     const operations_aggregate = await aggregateStaticCollection(this.db.bucket_data);
+    const v3OperationCollections = await this.db.listBucketDataCollectionsV3();
     const v3_operation_aggregates = await Promise.all(
-      (await this.db.listBucketDataCollectionsV3()).map((collection) =>
+      v3OperationCollections.map((collection) =>
         collection
           .aggregate([
             {
@@ -837,8 +835,9 @@ export class MongoBucketStorage extends storage.BucketStorageFactory {
 
     const parameters_aggregate = await aggregateStaticCollection(this.db.bucket_parameters);
 
+    const v3ParameterCollections = await this.db.listAllParameterIndexCollectionsV3();
     const v3_parameter_aggregates = await Promise.all(
-      (await this.db.listAllParameterIndexCollectionsV3()).map((collection) =>
+      v3ParameterCollections.map((collection) =>
         collection
           .aggregate([
             {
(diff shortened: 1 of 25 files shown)


modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
```diff
@@ -867,6 +868,119 @@ export class MongoBucketStorage extends storage.BucketStorageFactory {
           .catch(ignoreNotExisting)
       )
     );
+
+    const v3StorageConfig = getMongoStorageConfig(storage.STORAGE_VERSION_3) as StorageConfig & {
+      incrementalReprocessing: true;
+    };
+    const v3Db = this.db.versioned(v3StorageConfig) as VersionedPowerSyncMongoV3;
+    const objectStorageDefinitionUsage = await ObjectStorageUsage.readAllDefinitionUsage(v3Db);
```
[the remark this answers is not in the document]
rkistner: Ignoring existing V3 storage - not used in production. It's even mentioned in AGENTS.md now...

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/ObjectStorageUsage.ts
```diff
@@ -0,0 +1,230 @@
...
+          if (lib_mongo.isMongoNamespaceNotFoundError(error)) {
+            return [];
+          }
+          throw error;
+        });
+
+      return entries.map((entry) => {
+        const activeBytes = BigInt(entry.active_bytes ?? 0);
+        if (activeBytes < 0n) {
+          return {
```
stevensJourney, line 81: It seems like this could only happen under edge cases (like the baseline being missing). I might be missing something here though. 

Should we perhaps log a warning if we reach this case?

stevensJourney (approved):
The changes look good to me

modules/module-mongodb-storage/src/storage/implementation/v3/object-storage/ObjectStorageUsage.ts
rkistner: Yeah, it shouldn't generally happen unless the stats were corrupted somehow. It's just a sanity-check to prevent reporting negative values. Can consider adding logs later, but will merge as-is for now.

PR #777, opened 2026-08-27
fix: use unordered equality in sync compiler

stevensJourney:
When working on https://github.com/powersync-ja/powersync-service/pull/762, I noticed the `equalsIgnoringResultSetUnordered` should probably be using `unorderedEquality` instead of `listEquality`.

This allows the Sync Streams compiler to reuse equivalent query components when independent filters or attached table-valued functions are discovered in a different order.

Incremental reprocessing compares serialized JSON, with bucket data sources sorted, so this does not change how reordered queries are compared between deployments. The serialized JSON includes the compiler hashes though, so upgrading across this change will cause a one-time failure to reuse existing definitions.

AI usage: Manually implemented the change. Reviewed with Codex 5.6

### DIFF
--- a/.changeset/tidy-filters-agree.md
+++ b/.changeset/tidy-filters-agree.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Correctly reuse equivalent Sync Streams query components when independent filters or attached table-valued functions appear in a different order.
--- a/packages/sync-rules/src/compiler/compatibility.ts
+++ b/packages/sync-rules/src/compiler/compatibility.ts
@@ -1,4 +1,4 @@
-import { Equality, listEquality, StableHasher } from './equality.js';
+import { Equality, listEquality, StableHasher, unorderedEquality } from './equality.js';
 
 /**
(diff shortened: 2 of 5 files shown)


simolus3:
The issue found by Codex is technically already present, but this makes it more likely to be observed.

I think the solution is to, when comparing two `BaseSourceRowProcessor` instances `a` and `b`: 

1. Compare the structure of table-valued functions added to the processors (they must be equal either way).
2. Come up with two mappings `identity_a` and `identity_b` so that `identity_a(sourceResultSetMentionedInA) == identity_b(sourceResultSetMentionedInB)` iff the two result sets are structurally the same.
3. Pass that to `equalsAssumingSameResultSet` to compare filter and partition expressions.

We can't really do this for hash codes, but I think it's acceptable for `ja.value = 'x' AND jb.value = 'y'` and `jb.value = 'x' AND ja.value = 'y'` to have the same hash code, we'll catch them not being equal with the full check. Another option might be to embed hash codes of the table valued function when we want to hash references to it.

simolus3:
I'll open a fresh PR with a new description given much needs to change to fix that.

PR #780, opened 2026-09-01
Fix provenance for lookups with multiple inputs

simolus3:
This applies the patch contributed through [this analysis repo](https://github.com/benkraus/powersync-parameter-provenance-repro/tree/main): For lookups with multiple inputs, we'd previously generate a fresh symbol in every iteration calling `setOrUpdate`. This is incorrect, the outputs are part of the same materialized result set and should not be combined later.

The patch was generated by Codex, I manually reviewed changes and made some slight test simplifications.

I want to take a look at removing the provenance system altogether in favor of something that is actually capable of performing joins. That is a bigger change though, so this is another quick fix for the current approach.

### DIFF
--- a/.changeset/calm-lookups-correlate.md
+++ b/.changeset/calm-lookups-correlate.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-sync-rules': patch
+---
+
+Preserve multi-column parameter correlation when a deduplicated lookup is reached through multiple provenance paths.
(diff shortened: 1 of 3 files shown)


rkistner (approved):

benkraus:
Thanks for reviewing my clankers’ fix! Hope some of the docs at least were helpful

simolus3:
Thanks for sharing the patch! We'll also use this opportunity to replace the whole provenance thing with something that is much easier to understand: https://github.com/powersync-ja/powersync-service/pull/782.

PR #782, opened 2026-09-01
Use proper result set when evaluating parameters

simolus3:
When querying buckets for Sync Streams, we generally try to resolve parameters independently to form a cartesian product in the end. This is correct for most streams, but goes wrong when a parameter index has more than one column. For example, in `SELECT a.* FROM a, b WHERE a.c1 = b.c1 AND a.c2 = b.c2 AND b.u = auth.user_id()`, the two parameters are `b.c1` and `b.c2`. If we encounter multiple rows of `b` through a lookup result, we can't assume those to be independent parameters though! We can only pair parameters that originate from the same row.

This is currently implemented by tracking provenance for each parameter value back to the lookup this originally came from. When we build the cartesian product in the end, we ignore values with incompatible provenance from different rows. Unfortunately, tracking provenance is both kind of expensive and _very_ tricky to get right.

Semantically, evaluating bucket parameters involves:

1. Forming a result set by adding lookups through joins or a cartesian product.
2. Once we have the full result set, bucket parameters are a projection.

This replaces the previous querier logic with an actual result set implementation: We start out with a unit set of one row without columns, then go through added lookups that are cross-joined (table-valued functions) or inner-joined (parameter lookups). If we end up with an empty intermediate result set at any point, we know there won't be any buckets and bail out early.

Intersection parameters require special consideration now, but can be implemented by going through the result set and deleting rows where the columns don't match.

AI use: The approach is manual, most tests and some implementation details are generated with Claude Code.

### DIFF
--- a/packages/sync-rules/src/sync_plan/evaluator/parameter_evaluator.ts
+++ b/packages/sync-rules/src/sync_plan/evaluator/parameter_evaluator.ts
@@ -276,12 +276,10 @@ export class RequestParameterEvaluators {
     }
 
     for (const stage of lookupStages) {
-      const stageIndex = mappedStages.length;
       const mappedStage: PreparedExpandingLookup[] = [];
       mappedStages.push(mappedStage);
 
       for (const lookup of stage) {
-        const index = mappedStage.length;
         let resolved: PreparedExpandingLookup;
 
         if (lookup.type == 'parameter') {
@@ -316,6 +314,7 @@ export class RequestParameterEvaluators {
         }
 
         lookupToStage.set(lookup, resolved);
+        mappedStage.push(resolved);
       }
     }
 
--- a/packages/sync-rules/src/sync_plan/evaluator/parameter_evaluator.ts
+++ b/packages/sync-rules/src/sync_plan/evaluator/parameter_evaluator.ts
@@ -1,6 +1,5 @@
 import { ParameterLookupSource, ScopedParameterLookup, UnscopedParameterLookup } from '../../BucketParameterQuerier.js';
 import { ParameterIndexLookupCreator } from '../../BucketSource.js';
-import { HashMap, listEquality, StableHasher } from '../../compiler/equality.js';
 import { HydrationState } from '../../HydrationState.js';
 import { RequestParameters, SqliteParameterValue, SqliteValue } from '../../types.js';
 import { isValidParameterValue } from '../../utils.js';
@@ -14,6 +13,7 @@ import { MapSourceVisitor, visitExpr } from '../expression_visitor.js';
 import * as plan from '../plan.js';
 import { StreamInput } from './bucket_source.js';
 import { PreparedParameterIndexLookupCreator } from './parameter_index_lookup_creator.js';
+import { AsyncJoinLookup, ResultSet, ResultSetColumn, ResultSetElement } from './result_set.js';
 
 /**
  * Finds bucket parameters for a given request or subscription.
@@ -63,7 +63,12 @@ export class RequestParameterEvaluators {
     /**
      * Pending parameter values, or their cached outputs.
      */
-    readonly parameterValues: PreparedParameterValue[]
+    readonly parameterValues: PreparedParameterValue[],
+
+    /**
+     * The materialized result set into which
+     */
+    readonly resultSet: ResultSet
   ) {}
 
   /**
@@ -77,33 +82,10 @@ export class RequestParameterEvaluators {
    * instead of re-evaluating them on every parameter lookup change.
    */
   clone(): RequestParameterEvaluators {
-    function cloneValue(value: PreparedParameterValue): PreparedParameterValue {
-      switch (value.type) {
-        case 'intersection':
-          return { type: 'intersection', values: value.values.map(cloneValue) };
-        case 'request':
-        case 'lookup':
-        case 'cached':
-          return value;
-      }
-    }
+    const copiedStages = this.lookupStages.map((s) => s.map((e) => e.clone()));
+    const outputValues = this.parameterValues.map((v) => v.clone());
 
-    function cloneLookup(lookup: PreparedExpandingLookup): PreparedExpandingLookup {
-      switch (lookup.type) {
-        case 'parameter':
-          // We need to clone the instantiation array as well.
-          return { type: 'parameter', lookup: lookup.lookup, instantiation: lookup.instantiation.map(cloneValue) };
-        case 'table_valued':
-        case 'cached':
-          return lookup;
-      }
-    }
-
-    return new RequestParameterEvaluators(
-      this.stream,
-      this.lookupStages.map((stage) => stage.map(cloneLookup)),
-      this.parameterValues.map(cloneValue)
-    );
+    return new RequestParameterEvaluators(this.stream, copiedStages, outputValues, this.resultSet.clone());
   }
 
   /**
@@ -115,13 +97,34 @@ export class RequestParameterEvaluators {
    * If dynamic lookups are required to resolve parameters, returns `undefined`.
    */
   partiallyInstantiate(input: PartialInstantiationInput): SqliteParameterValue[][] | undefined {
-    const helper = new PartialInstantiator(input, this);
+    try {
+      // At this point, we can resolve table-valued lookups and parameter values based only on request data.
+      for (const stage of this.lookupStages) {
+        for (const element of stage) {
+          if (element instanceof TableValuedExpandingLookup) {
+            const outputs = element.read(input.request);
+            element.wasResolved = true;
+            this.resultSet.multiply(element.resultSetIndex, outputs);
 
-    this.lookupStages.forEach((stage, stageIndex) => {
-      stage.forEach((_, indexInStage) => helper.expandingLookupSync(stageIndex, indexInStage));
-    });
+            this.#checkInstantiable();
+          } else {
+            for (const instantiation of element.instantiation) {
+              if (instantiation instanceof RequestParameterValue) instantiation.resolveWith(input);
+            }
+          }
+        }
+      }
 
-    return helper.tryResolveInstantiation(this.parameterValues)?.map(withoutProvenance);
+      for (const parameter of this.parameterValues) {
+        if (parameter instanceof RequestParameterValue) parameter.resolveWith(input);
+      }
+
+      return this.#readParameters();
+    } catch (e) {
+      if (e === uninstantiableException) return [];
+
+      throw e;
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/sync-rules/src/sync_plan/evaluator/parameter_evaluator.ts
```diff
@@ -130,16 +158,131 @@ export class RequestParameterEvaluators {
...
+          }
+        }
+
+        for (const intersection of intersections) {
+          this.#applyIntersectionConstraint(intersection);
+        }
+      }
 
-    for (let i = 0; i < this.lookupStages.length; i++) {
-      // Within a stage, we can resolve lookups concurrently.
```
simolus3, line 136: With the implementation from this PR, joins are no longer processed concurrently. It's possible to add that back with minor added complexity, but:

- this is only relevant for complex sync streams
- we already evaluate other users / queriers concurrently, to the point where bucket storage is likely the bottleneck and not JS

So I don't think this is necessarily something worth doing, but I can change this here / in a follow-up PR if needed.

simolus3:
@codex review

packages/sync-rules/src/sync_plan/evaluator/result_set.ts
```diff
@@ -0,0 +1,212 @@
...
+        if (existingGroup != null) {
+          yield { index: i, group: existingGroup, first: false };
+        } else {
+          const group = generateGroup([value]);
+          foundValues.set(value, group);
+          yield { index: i, group, first: true };
+        }
+      }
+    } else {
+      const foundValues = new HashMap<SqliteParameterValue[], T>(parameterArrayEquality);
```
simolus3, line 166: It's worth noting that plain `Map`s treat a `NaN` values as equal to other `NaN` values, `parameterArrayEquality` doesn't.

We had this bug before as well, I think we should at least be consistent. I don't know what the intended behavior for `NaN` bucket parameters is, should they match buckets?

simolus3: I actually think this is fine, this only de-duplicates `NaN` values but they'd still compare as distinct parameters on buckets.

rkistner (commented):
Codex picked up a performance regression here with intersecting results. Example queries:

```sql
-- In this case we can intersect results
SELECT * FROM issues
  WHERE id IN auth.parameter('allowed_issue_ids')
    AND id IN subscription.parameter('visible_issue_ids')
    AND id IN subscription.parameter('selected_issue_ids');

-- In this case we can compute the checks independently
SELECT * FROM issues
  WHERE 'issues' IN auth.parameter('tables')
    AND 'read' IN auth.parameter('permissions');
```

With the old implementation, the intersection was computed first. The new implementation builds up the Cartesian product of the arrays before filtering them. In a cases like the above, the number of results can explode even with modestly-sized arrays, leading to slow performance and/or OOM-crashes.

simolus3:
That is a good point. Since the columns of intersections are known statically, we can evaluate them for each row added to the result set instead of removing rows after building the cartesian product.

packages/sync-rules/src/sync_plan/evaluator/result_set.ts
```diff
@@ -0,0 +1,319 @@
...
+    rowIndex: number,
+    rows: SqliteParameterValue[][]
+  ): boolean {
+    const originalRow = this.#rows[rowIndex];
+
+    let isFirst = true;
+    for (const row of rows) {
+      if (!filter(originalRow, row)) {
+        continue;
+      }
```
rkistner, line 223: The memory usage is now only `O(N)` instead of the `O(N^2)` before, but the comparisons are still `O(N^2)` here.

This affects the same query example mentioned before:

```sql
SELECT * FROM issues
  WHERE id IN auth.parameter('allowed_issue_ids')
    AND id IN subscription.parameter('visible_issue_ids')
    AND id IN subscription.parameter('selected_issue_ids');
```

packages/sync-rules/src/sync_plan/evaluator/result_set.ts
```diff
@@ -0,0 +1,319 @@
...
+
+    return {
+      deletedRows,
+      filter,
+      [Symbol.dispose]: () => {
+        let offset = 0;
+        for (const toDelete of deletedRows) {
+          this.#rows.splice(toDelete - offset, 1);
+          offset++;
+        }
```
rkistner, line 198: `Array.splice()` is `O(N)` on the array size, making this loop `O(N^2)`.

rkistner (commented):
I like the approach of using a ResultSet here, but Codex picked up another couple of performance regressions with the implementation (manually confirmed the findings).

packages/sync-rules/src/sync_plan/evaluator/result_set.ts
simolus3: This is tricky to optimize in `ResultSet` because it also needs to work for async joins from parameters (where we at least know we have a constant upper bound due to the paramer limit...).

To avoid pathological cases for parameter-based intersections, I'll look into optimizing those specifically in the evaluator.

packages/sync-rules/src/sync_plan/evaluator/result_set.ts
simolus3: Replaced with an `O(|deletedRows|)` implementation

packages/sync-rules/src/sync_plan/evaluator/result_set.ts
simolus3: I have added this as an optimization to `partiallyInstantiate`.

packages/sync-rules/src/sync_plan/evaluator/parameter_evaluator.ts
```diff
@@ -130,16 +247,128 @@ export class RequestParameterEvaluators {
    * Because this needs to lookup parameter indexes, it is asynchronous.
    */
   async instantiate(input: InstantiationInput): Promise<SqliteParameterValue[][]> {
-    const helper = new FullInstantiator(input, this);
+    try {
+      for (const { lookups } of this.lookupStages) {
+        for (const lookup of lookups) {
+          if (lookup instanceof ParameterIndexExpandingLookup) {
+            await this.#instantiateLookup(lookup, input);
```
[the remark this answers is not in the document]
simolus3: See https://github.com/powersync-ja/powersync-service/pull/782#discussion_r3905756619.

packages/sync-rules/src/sync_plan/evaluator/result_set.ts
```diff
@@ -0,0 +1,328 @@
...
+    const affectedColumnsInAddedResultSet = intersection.columns.filter(
+      ({ lookup }) => lookup.resultSetIndex === addedResultSetIndex
+    );
+
+    if (intersection.fixedValue !== undefined) {
+      // All rows already in the result set satisfy the intersection and must match the fixed value in relevant columns.
+      // So when checking a new row, we just need to check columns there.
+      return function (_existingRow: ResultSetRow, added: SqliteParameterValue[]): boolean {
+        for (const { outputIndex } of affectedColumnsInAddedResultSet) {
+          if (added[outputIndex] !== intersection.fixedValue) return false;
```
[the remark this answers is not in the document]
simolus3: We alrady have this problem on `main`, e.g. here where we're using a map which also treats `1` and `1n` as distinct:

https://github.com/powersync-ja/powersync-service/blob/424bbc010841677c63b91df1cc4c78c94808bee6/packages/sync-rules/src/sync_plan/evaluator/parameter_evaluator.ts#L373-L376

This is a legitimate problem, but a fix is unrelated to this. We would have to do this consistently, or might just get away with parsing parameters with jsonbig as well.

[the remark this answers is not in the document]
simolus3: I have opened #816 as a follow-up PR to fix this.

packages/sync-rules/src/sync_plan/evaluator/parameter_evaluator.ts
```diff
@@ -115,13 +118,127 @@ export class RequestParameterEvaluators {
...
+        for (const element of stage.lookups) {
+          if (element instanceof TableValuedExpandingLookup) {
+            const outputs = element.read(input.request);
+            const intersections = parameterToIntersection.get(element);
+            if (intersections == null || intersections.length == 0) {
+              this.resultSet.multiply(element.resultSetIndex, outputs);
+              this.#checkInstantiable();
+            } else {
+              // For this parameter to be part of the intersection optimization, it must return exactly one column.
+              for (const [value] of new Set(outputs)) {
```
rkistner: Outputs is `SqliteParameterValue[][]`, so this creates a `Set<SqliteParameterValue[]>`. That will compare arrays by identity, rather than value, so de-duplication doesn't work properly.

Since you're only using one array element, it should be sufficient to just change the order:
```ts
for (const value of new Set(outputs.map(([value]) => value))) {
```

A regression test covering this case would also be good - see this suggestion from Codex:
```
For a IN subscription.parameter('x') AND a IN auth.parameter('y'):

  - x = ['unauthorized', 'unauthorized'], y = ['allowed'] → no buckets.
  - x = ['allowed', 'allowed'], y = ['allowed'] → exactly the allowed bucket.
 ```

File: packages/sync-rules/src/sync_plan/evaluator/parameter_evaluator.ts [hunk 8]
```diff
@@ -115,13 +118,127 @@ export class RequestParameterEvaluators {
...
+              return intersections.every((intersection) => {
+                const matchingCount = intersection.rows.get(value);
+                return matchingCount === intersection.amountOfRequestParameters;
+              });
+            })
+          );
+        }
+
+        this.#checkInstantiable();
+      }
```
rkistner: If I understand it correctly, the optimization above now removes values that don't intersect. But any remaining values are still multiplied, which can result in exponential computations and memory.

Maybe it can work to have a multiply function that takes into account the intersection, something like this?

```ts
resultSet.multiplyCorrelated(
    parameters.map((p) => p.resultSetIndex),
    intersection.materializedRows
  );
```

packages/sync-rules/src/sync_plan/evaluator/result_set.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    const originalLength = this.#rows.length;
+    for (let i = 0; i < originalLength; i++) {
+      const lookup = lookupsByRow[i];
+      if (lookup.foundRows.length === 0 || !this.#multiplyAtRow(resultSetIndex, add.filter, i, lookup.foundRows)) {
+        // The row has no matching join partner, so remove it. We can't split it immediately because #multiplyAtRow is
+        // still iterating through rows.
+        add.deletedRows.push(i);
+      }
+    }
```
rkistner: This still does `O(N*M)` comparisons, even though the results are `O(N+M)`. An in-memory index (Map) should be able to reduce the comparisons to `O(N+M)` as well.

Here it appears that the previous implementation did cover intersection on a _single_ key efficiently. A stretch goal would be to cover multiple intersection parameters, such as x and y in this example:

```sql
SELECT items.*
  FROM items
  JOIN allowed a ON items.x = a.x AND items.y = a.y
  JOIN enabled e ON items.x = e.x AND items.y = e.y
  WHERE a.user_id = auth.user_id()
```

I believe the previous implementation would efficiently intersect on `x`, then multiply on filter on `y`. We could make that even more efficient by intersecting on `(x, y)`, but that's a further optimization rather than covering a performance regression.

rkistner (commented):
The further optimizations did help, but there are still some further optimizations needed to avoid performance regressions. See the individual comments, analyzed using Codex.

PR #790, opened 2026-09-09
fix(postgres): skip replication-byte recorder in test-connection

### LINKED ISSUE #751: Postgres test-connection fails because replication-byte metric is not registered
## Summary

The `test-connection` command fails for an otherwise-valid PostgreSQL replication configuration before it can report the connection result.

`test-connection` registers a `ReplicationEngine`, but `TEST_CONNECTION` mode receives no replication metrics. `PostgresModule` then installs a pgwire byte recorder that accesses an unregistered `DATA_REPLICATED_BYTES` counter.

## Environment

- Runtime failure reproduced with PowerSync Service v1.23.3, pinned to OCI digest `sha256:b6b22fa7d0d862f04bdff62846e656756d17bcf3dd6eca399a0633671051438b`
- PostgreSQL replication source
- The five causal files are unchanged in stable v1.24.0, prerelease v1.25.0, and current `main` at `d3216be9f1f285fd4f2622a7d987ebaf4c2b8e0c`

## Reproduction

Use an otherwise-valid self-hosted configuration containing a PostgreSQL replication connection:

```bash
docker run --rm \
  -v "$PWD/powersync.yaml:/config/powersync.yaml:ro" \
  journeyapps/powersync-service:1.23.3@sha256:b6b22fa7d0d862f04bdff62846e656756d17bcf3dd6eca399a0633671051438b \
  test-connection --config-path /config/powersync.yaml
```

## Actual result

The command exits nonzero with:

```text
PSYNC_S0001: Counter powersync_data_replicated_bytes_total has not been created and registered yet.
```

## Expected result

The command should complete its PostgreSQL connection and permission checks and report their actual result.

## Source trail in v1.24.0

- [`test-connection` selects `TEST_CONNECTION` mode and registers `ReplicationEngine`](https://github.com/powersync-ja/powersync-service/blob/v1.24.0/packages/service-core/src/entry/commands/test-connection-action.ts#L17-L36)
- [`TEST_CONNECTION` is absent from the metrics-mode map](https://github.com/powersync-ja/powersync-service/blob/v1.24.0/modules/module-core/src/CoreModule.ts#L160-L172)
- [Replication counters are created only for `MetricModes.REPLICATION`](https://github.com/powersync-ja/powersync-service/blob/v1.24.0/packages/service-core/src/metrics/register-metrics.ts#L20-L36)
- [The PostgreSQL recorder accesses `DATA_REPLICATED_BYTES` whenever a replication engine exists](https://github.com/powersync-ja/powersync-service/blob/v1.24.0/modules/module-postgres/src/module/PostgresModule.ts#L32-L40)
- [Missing counters throw `ServiceAssertionError`](https://github.com/powersync-ja/powersync-service/blob/v1.24.0/packages/service-core/src/metrics/MetricsEngine.ts#L56-L60)

## Impact

This makes the service CLI unsuitable as a PostgreSQL readiness or deployment gate, even when the database configuration itself is valid.

Possible fixes appear to be either registering replication metrics for `TEST_CONNECTION` or avoiding the PostgreSQL replication-byte recorder in this mode. Which behavior is intended for `test-connection`?

Tyagiquamar:
## Summary
- `test-connection` runs in `TEST_CONNECTION` mode, which never registers replication metrics.
- `PostgresModule` still installed a jpgwire recorder that reads `DATA_REPLICATED_BYTES`, so the CLI failed with `PSYNC_S0001` before the connection check ran.
- Skip that recorder in `TEST_CONNECTION` mode. Maintainer confirmed on #751 that metrics are not needed for this command.

Fixes #751

## Test plan
- [x] `pnpm --filter @powersync/service-module-postgres exec vitest run test/src/metrics_recorder.test.ts` (2 passed)
- [ ] `test-connection` against a valid PostgreSQL replication config reports the connection result instead of the missing-counter error

### DIFF
--- a/.changeset/skip-test-connection-replication-metrics.md
+++ b/.changeset/skip-test-connection-replication-metrics.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-module-postgres': patch
+---
+
+Skip the PostgreSQL replication-byte metrics recorder in `test-connection` mode so the command can run without `DATA_REPLICATED_BYTES`.
--- a/modules/module-postgres/src/module/PostgresModule.ts
+++ b/modules/module-postgres/src/module/PostgresModule.ts
@@ -31,7 +31,7 @@ export class PostgresModule extends replication.ReplicationModule<types.Postgres
 
   async onInitialized(context: system.ServiceContextContainer): Promise<void> {
     // Record replicated bytes using global jpgwire metrics. Only registered if this module is replicating
-    if (context.replicationEngine) {
+    if (context.replicationEngine && context.serviceMode !== system.ServiceContextMode.TEST_CONNECTION) {
       jpgwire.setMetricsRecorder({
         addBytesRead(bytes) {
           context.metricsEngine.getCounter(ReplicationMetric.DATA_REPLICATED_BYTES).add(bytes);
--- a/modules/module-postgres/test/src/metrics_recorder.test.ts
+++ b/modules/module-postgres/test/src/metrics_recorder.test.ts
@@ -1,0 +1,40 @@
+import { system } from '@powersync/service-core';
+import * as jpgwire from '@powersync/service-jpgwire';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { PostgresModule } from '../../src/module/PostgresModule.js';
+
+describe('PostgresModule metrics recorder', () => {
+  afterEach(() => {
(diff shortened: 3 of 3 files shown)


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

PR #792, opened 2026-09-10
Use MongoDB client-level bulkWrite where available

rkistner:
This uses the new client-level [bulkWrite](https://www.mongodb.com/docs/v8.0/reference/command/bulkWrite/#mongodb-dbcommand-dbcmd.bulkWrite) command for storage operations on MongoDB 8.0+.

This is different from the connection-level bulkWrite command:
1. The same command can be used for writes across multiple collections.
2. Different operation types such as insertOne and updateMany can be used in the same wire command, even with `ordered: true`.

The collection-level bulkWrite command could do only one operation type at a time. So if you had an ordered bulkWrite with say `insertOne, updateOne, insertOne, insertOne`, it would send 3 wire-level commands instead of 1, which can add significant latency.

In addition to using the client-level bulkWrite when available, this now de-duplicates all records client-side, so we never need a collection-level bulkWrite with `ordered: true`, even on older MongoDB versions.

We still keep our implementation compatible with MongoDB 6 and 7, by falling back to individual operations and collection-level bulkWrite.

We can expect some replication latency improvements here, especially with `storage_version: 4` on many different bucket definitions. We can also expect minor replication throughput improvements due to the decrease in latency.

## Expected changes on MongoDB 8.0

_V1 = `storage_version: 2`; V3 = `storage_version: 4`, we match the implementation names here._

| Modified operation                                                                                           | Before → after                                |
| ------------------------------------------------------------------------------------------------------------ | --------------------------------------------- |
| V1 replication flush, including transaction tail                                                             | **5–6 → 2**                                   |
| V3 inline replication flush, including transaction tail                                                      | **6–8 → 2**                                   |
| V3 with S3 replication flush, including transaction tail                                                          | **7–9 → 2**                                   |
| V1/V3 mark tables’ snapshots complete, including checkpoint barrier                                          | **2 → 1**                                     |
| V3 replicated custom checkpoints across E event collections                                                  | **E → 1**                                     |
| V1 replicated custom checkpoints                                                                             | **1 → 1**, fewer operations when users repeat |
| Managed-checkpoint creation, generated or supplied IDs                                                       | **1 → 1**, now unordered                      |
| V1 compaction: flush data changes and bucket state                                                           | **2–3 → 1**                                   |
| V1 compaction: replace operations with CLEAR                                                                 | **2 → 1**                                     |
| V3 inline compaction: replacement, leading CLEAR, or boundary CLEAR                                          | **2 → 1**                                     |
| V3 with S3 compaction: those same three operations, including publication/retirement markers and usage accounting | **Up to 5 → 1**                               |

The replication figures assume one persisted-data flush, one bucket definition,
and one source table. The ranges reflect optional parameter writes and V3
delete-watermark updates. Additional V3 collections increase the old count while
still fitting into one new data bulk.

Previously, V1 ordered source-record bulks could split into many commands when
updates and deletes alternated. This is the main case affected by previously-ordered collection-level bulkWrite. Client-side de-duplication now eliminate that extra fragmentation.

MongoDB 6/7 retain separate collection commands, but still benefit from deduplication and unordered writes. Large requests can require multiple commands;
client-bulk result cursors can also require additional `getMore` commands.

## Implementation

A new `MongoWriteBatch` class handles the client-level bulkWrite command with fallback support.

No new tests added - the implementation is covered by existing regression tests.

For source-record writes and write checkpoints, we now de-duplicate operations before doing a bulkWrite, allowing us to use unordered writes. For the rest, we mostly preserve the previous behavior.

## AI Usage

Implemented using Codex gpt-6-astra. Manually checked all changes. The above table is also generated using Codex.

### DIFF
--- a/.changeset/wide-feet-reply.md
+++ b/.changeset/wide-feet-reply.md
@@ -1,0 +1,5 @@
+---
+'@powersync/service-module-mongodb-storage': minor
+---
+
+Use MongoDB client-level bulkWrite where available.
--- a/modules/module-mongodb-storage/src/storage/implementation/MongoBucketBatch.ts
+++ b/modules/module-mongodb-storage/src/storage/implementation/MongoBucketBatch.ts
@@ -36,6 +36,7 @@ import type { VersionedPowerSyncMongo } from './db.js';
 import { MAX_ROW_SIZE } from './MongoBucketBatchShared.js';
 import { MongoIdSequence } from './MongoIdSequence.js';
 import { MongoParsedSyncConfigSet } from './MongoParsedSyncConfigSet.js';
+import { MongoWriteBatch } from './MongoWriteBatch.js';
 import { OperationBatch, RecordOperation } from './OperationBatch.js';
 import { ObjectStorage } from './v3/object-storage/ObjectStorage.js';
 import { createObjectStorageUsageWriterId } from './v3/object-storage/ObjectStorageUsage.js';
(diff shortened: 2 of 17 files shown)


stevensJourney (approved):
This looks like a solid improvement to me.

PR #795, opened 2026-09-10
fix (MongoDB): Improve Schema detection

stevensJourney:
Previously, schema detection fetched full MongoDB documents and inspected them in the service.

With large documents, even a small sample could use a lot of memory just to work out the field names and types.

This moves schema detection into a MongoDB aggregation query, so the service only receives the field names and detected types. Memory usage in the service now scales with the inferred schema rather than the size of the sampled documents.

AI Usage: Implemented with Codex gpt-6. Manually tested and verified.

### DIFF
--- a/modules/module-mongodb/src/api/infer-collection-schema.ts
+++ b/modules/module-mongodb/src/api/infer-collection-schema.ts
@@ -32,7 +32,10 @@ const BSON_TYPES: Record<string, { name: string; sqliteType: ExpressionType }> =
  * Infer top-level fields without transferring sampled document values to the service.
  * Memory on the service scales with the inferred schema, not the size of those values.
  */
-export async function inferCollectionSchema(collection: mongo.Collection): Promise<TableSchema['columns']> {
+export async function inferCollectionSchema(
+  collection: mongo.Collection,
+  isDocumentDb: boolean
+): Promise<TableSchema['columns']> {
   const fields = await collection
     .aggregate<{ _id: string; types: string[] }>(
       [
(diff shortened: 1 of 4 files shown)


modules/module-mongodb/src/api/infer-collection-schema.ts
```diff
@@ -0,0 +1,127 @@
...
+        { $group: { _id: '$fields.name', types: { $addToSet: '$fields.type' } } }
+      ],
+      {
+        // Bound execution per collection below the default 60-second socket timeout
+        // so MongoDB can return a query timeout before the connection times out.
+        maxTimeMS: 30_000,
+        // When $sample can't use a random cursor, it sorts full documents before the
+        // projection. Allow that sort to spill to disk if large documents exceed its
+        // memory limit, instead of failing schema inference.
+        allowDiskUse: true,
```
rkistner: When would $sample not be able to use a random cursor?

It might be safer to just fail in that case, rather than risking overloading the source database with 4x concurrent queries spilling to disk.

rkistner (dismissed):
Added a nitpick on disk use, but it's not a blocker: this already appears significantly better than the current approach.

modules/module-mongodb/src/api/infer-collection-schema.ts
stevensJourney: For our 50-document sample, collections with 1,000 documents or fewer fall outside MongoDB’s documented random-cursor conditions and require a scan and random sort ([MongoDB $sample documentation](https://www.mongodb.com/docs/manual/reference/operator/aggregation/sample/)). Large sampled documents could maybe push that sort over its memory limit and cause disk spill. I’ve now set `allowDiskUse: false` so it fails instead.

rkistner: Ok, so if I understand correctly, MongoDB would only hold 50 documents at a time in memory. But if those 50 documents exceed 100MB in size, it could trigger "spill to disk" (or fail if allowDiskUse is disabled).

I think it's fine to fail in that case - it should be fairly rare.

rkistner (approved):

Issue #801, opened 2026-09-14
Issues running PowerSync Service (Open Edition) natively on Windows


T0b1i:
## Context

At my company, we are building an on-premise application that our customers host themselves. Some of them run
exclusively Windows Server (no Linux, no Docker), and all data has to stay on their own infrastructure.

I asked on Discord whether Windows is supported. The answer was that there are no plans to support it for the
Open Edition, but that the PowerSync Team would be happy to investigate issues found when running it as a native Node.js process.
This issue lists what I found.

## Issues

I built the service from source with Node.js 24 on Windows 11 and on the GitHub Actions `windows-2025` runner.
Three small issues block Windows today.

**1. Storage migrations fail to load on Windows**

`AbstractPowerSyncMigrationAgent.loadMigrations` (`packages/service-core/src/migrations/PowerSyncMigrationManager.ts`)
imports migration scripts with `import(path.resolve(migrationsDir, migration))`. On Windows this is an absolute
path like `C:\...`, which the ESM loader rejects:

```
ERR_UNSUPPORTED_ESM_URL_SCHEME: Only URLs with a scheme in: file, data, and node are sup
ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
```

This affects `migrate up` and automatic migrations on startup. Fix: `import(pathToFileURL(path.resolve(...)).href)`.

**2. `pnpm test:core` silently runs no tests on Windows**

pnpm runs package.json scripts with `cmd.exe` on Windows, which does not treat single quotes as quotes.
`--filter '!./modules/*'` then matches nothing (`No projects matched the filters "'!./mo
command exits with code 0. Double quotes work on Windows and Linux.

**3. `BroadcastIterable` tests depend on the timer resolution**

Three tests assume ~1 ms timer resolution. Windows timers resolve at ~15 ms, so they fail there. The fix runs these
tests on fake timers (`vi.useFakeTimers()`) with the assertions unchanged, which makes them deterministic on every
platform.

## Results

With these fixes, `pnpm install`, `pnpm build` and all core tests pass on Windows. I also ran all module tests
locally on Windows 11, including the slow tests, and they all pass: Postgres, Postgres storage, MongoDB, MongoDB
storage, MySQL, SQL Server and Convex. The databases ran in Docker, Convex as the local development backend.
Apart from the issues above, I found no Linux-specific dependencies.

## Next steps

[...]

PR #806, opened 2026-09-15
Add a Windows CI job for build and core tests

T0b1i:
Follow-up to #801 / #802. Not a bug fix, this adds the CI coverage @rkistner suggested there:

> I think having one Windows CI job covering build and core tests make sense. If there is a specific need we can consider running some of the database-specific tests on Windows as well, but I wouldn't want to run the full matrix we have on Linux currently.

## What this adds

A single `Core Test (Windows)` job on `windows-latest`. It is the existing `Core Test` job with a different `runs-on`, and it reuses `./.github/actions/setup-node-build` unchanged, so there is one place to keep in sync. Linux CI is untouched.

## Follow-up: Postgres tests on Windows

To answer "if there is a specific need": we ship an on-premise product that uses PowerSync as its sync service, and some customers host it on Windows Server only - no Linux, no Docker. It uses Postgres both as the replication source and as the storage backend, so `module-postgres` and `module-postgres-storage` are the two that matter for us.

There is also a coverage argument. The migration bug from #801 - `ERR_UNSUPPORTED_ESM_URL_SCHEME` when importing migration scripts by absolute Windows path - only surfaces when migrations are loaded from disk, and `pnpm test:core` never reaches that code. The job in this PR would not have caught it. The Postgres storage tests would have.

If you are open to it I would send that as a separate PR: one job, no matrix. Happy to drop it if that is more CI than you want to maintain.

## Unrelated question: build artifacts on releases

Different topic, and happy to move it to its own issue if you prefer. `packages_release`. GitHub release per service version tag, and `@powersync/service-image` is `private`, so the Docker image is the only released artifact of the service itself - which means building from source for anyone who cannot run Docker.

Would you consider attaching a prebuilt bundle of the service to that release as an asset? That would be great for us, but we completely understand if it is not something you want to take on. We are already very grateful for the response on #801.

## AI disclaimer

This change was made with the help of an AI coding assistant. I reviewed it, and the Windows job ran green on my fork before opening this PR.


rkistner (approved):
Thanks, happy with these changes, and happy to add a Postgres job as well.

For the service build, would that essentially just be publishing `@powersync/service-image` publicly? That also sounds good to me

T0b1i:
@rkistner Thanks! I'll send the Postgres job as a follow-up PR so this one can be merged as is, unless you'd rather have it added here. 
On the service build: just to make sure I understand, do you mean publishing `@powersync/service-image` to npm, like the other `@powersync/*` packages? If so, that would work well for us, and I'm happy to send a PR for it.

PR #807, opened 2026-09-15
Add a Windows CI job for Postgres tests

T0b1i:
Follow-up to #806, where @rkistner was happy to add a Postgres job on Windows as well.

## What this adds

A single `Postgres Test (Windows)` job on `windows-latest`, running after `Core Test (Windows)`, with no matrix. It runs the `module-postgres` and `module-postgres-storage` tests, the two modules that matter for our Windows Server deployments. The job takes about 7-9 minutes.

- Windows runners cannot run Linux containers, so the job uses the PostgreSQL version preinstalled on the runner image (currently 17) instead of Docker. This is also closer to our deployments: running Linux containers on Windows Server is not really supported for production, so PostgreSQL runs natively there as well. The job enables `wal_level = logical` and creates both test databases on one instance, matching the defaults in the test `env.ts` files, so no connection URLs are needed.
- It sets `fsync = off` for this throwaway instance, so the tests run faster and do not run into timeouts. My assumption is that either the disk on the Windows runners is slower than on the Linux runners, or PostgreSQL writes data to disk in a different way on Windows, which makes it slower there. With `fsync` on, 9 of 10 runs on my fork failed because `replication slot lost` exceeded its 20 s timeout. `fsync` only skips forcing writes to disk. WAL contents, LSNs and the flush position should stay the same, so replication behaves the same.
- It reuses `./.github/actions/setup-node-build` unchanged.

## Test changes

- **Validation tests:** The tests in `modules/module-postgres/test/src/validation.test.ts` always used MongoDB storage, even with `TEST_MONGO_STORAGE=false`, so they failed without MongoDB. They do not depend on the storage implementation, so `util.ts` now exports `INITIALIZED_STORAGE_FACTORY`, which uses MongoDB storage when it is enabled and Postgres storage otherwise. The Linux jobs keep using MongoDB storage for these tests. I wasn't sure whether using MongoDB storage here was intended, or whether I just spotted an oversight. Happy to change this to whatever approach you prefer.
- **`withMaxWalSize`:** `pg_reload_conf()` returns before the new `max_slot_wal_keep_size` is applied, and the helper only waited a fixed 100 ms. On Windows that was not always enough: `returns WAL budget with configured limit` failed because a pooled connection still saw the old value, so `safe_wal_size` was null. The helper now polls until the setting is visible, following the loop in `getClientCheckpoint`, and the test waits with `vi.waitFor` until the budget reports the limit.

## Stability

I ran the job 10 times in parallel on my fork (with a temporary matrix, not part of this PR). All 10 runs passed, each with 180 replication and 172 storage tests.
image

---

The npm publishing of `@powersync/service-image` will follow as a separate PR.

## AI disclaimer

This change was made with the help of an AI coding assistant. I reviewed it, and the Windows job ran green on my fork (10 of 10 runs) before opening this PR.


rkistner (approved):
Thanks, this looks good for a start.

We'll see if we need to restructure to make these faster to run in the future, for example by splitting storage and replication test jobs, but for now it seems like these are still faster than the Linux-based Postgres replication jobs. That could be from disabling fsync, so that's also something to investigate for the Linux jobs in the future.

PR #808, opened 2026-09-16
Publish the service as an npm package

T0b1i:
Follow-up to #806, where @rkistner asked: "For the service build, would that essentially just be publishing `@powersync/service-image` publicly? That also sounds good to me". This is that change.

We run PowerSync on Windows Server, where Linux containers are not really supported for production, so the Docker image is not an option for us. Today the service itself is not published anywhere else, which means building from source.

## Before this can be merged

I think two things have to happen on npm before this is merged, but I have never published an npm package, so I would rather ask than assume:

1. Does the package have to exist on npm first? The [`npm trust` docs](https://docs.npmjs.com/cli/v11/commands/npm-trust/) say *"The package you're configuring must already exist on the npm registry"*, which I read as: a trusted publisher can only be set up after a first, manual publish.
2. Does the trusted publishing configuration need direct publishing enabled? Since 3 September 2026, *"Direct publishing is opt-in per configuration"* ([changelog](https://github.blog/changelog/2026-09-03-multiple-trusted-publishing-configurations-for-npm/)), and the release workflow publishes directly through `pnpm changeset publish`.

If both are true, the next release would fail when `changeset publish` reaches this package unless that setup is done first. Is there anything else left, I can test to make sure that this step does not fail?

## What this changes

`service/package.json`:

- `repository`, same value as the other packages
- `publishConfig.access: public`, because `.changeset/config.json` sets `"access": "restricted"` globally
- `files: ["lib/**/*"]`, without it ships `src/`, `Dockerfile`, `local-dev/` and `tsconfig.tsbuildinfo`
- `private: true` removed, so `changeset publish` picks the package up

Two optional commits on top, easy to drop if you do not want them:

- **`Add a powersync-service executable`**: a `bin` entry plus a shebang in `src/entry.ts`, so `npx powersync-service start -c config.yaml` works. `tsc` keeps the shebang, and the Docker entrypoint is unaffected.
- **`Document running the service from npm`**: a short section in `service/README.md`, next to the image instructions. That file is also the npm package page.

## AI disclaimer

This change was made with the help of an AI coding assistant. I reviewed it and verified as far as I could. Is there anything else you would like me to check?


T0b1i:
I don't think the MongoDB Replication (8.0) failure on `storage v4 > replicating renameCollection` is related to this PR.

rkistner (approved):

rkistner:
Thanks for the change. We'll handle the trusted publishing config for the new package when we do the next release

T0b1i:
Thanks for the quick review and the merge, and for being open to these Windows-related changes in the first place.

@rkistner, @simolus3 the feedback was fast and clear across all four PRs, and making these changes was a genuinely good experience. The Discord support has been just as helpful whenever I had questions, so a thank you to the whole team as well.

Is there a rough timeframe for the next release? Once `@powersync/service-image` is published, I'm happy to verify the install and startup path on Windows in a real environment.

PR #814, opened 2026-09-22
fix: Merge static and dynamic bucket descriptions

### LINKED ISSUE #813: Postgres storage returns a bucket's operations twice when a stream reaches the bucket through a static and a dynamic parameter (client checksum failure loop)
## Summary

With **Postgres bucket storage**, the service returns a bucket's operations **twice in one `data` line** when a Sync Stream reaches the same bucket through both a static parameter (`id = auth.user_id()`) and a dynamic one (`id IN (SELECT …)`). The checkpoint line lists the bucket once with the correct `count` and `checksum`, so every client fails the bucket checksum on every connection (`expected X, got 2X = X (op) + X (add)` in the Rust core) and reconnects for ever. The same configuration on **MongoDB bucket storage** returns the operation once.

- Service: `journeyapps/powersync-service:1.26.1` (`start -r unified`), Sync Streams (`config.edition: 3`)
- Source: PostgreSQL 16 (`wal_level=logical`)
- Bucket storage: PostgreSQL (`storage.type: postgresql`) → duplicate; MongoDB 7 → correct
- Clients that hit it: powersync-swift 1.16.0 / powersync-sqlite-core 0.5.2 (the checksum-failure loop); a raw `POST /sync/stream` shows the duplicate without any client

## Minimal reproduction

`init.sql` (mounted into `/docker-entrypoint-initdb.d/`):

```sql
create table profiles (id uuid primary key, name text not null);
create table memberships (id uuid primary key, owner_id uuid not null, member_id uuid not null);
create publication powersync for all tables;
insert into profiles values ('11111111-1111-1111-1111-111111111111', 'alice');
-- alice is a member of her own group, so the dynamic parameter yields her own id as well
insert into memberships values ('22222222-2222-2222-2222-222222222222',
  '11111111-1111-1111-1111-111111111111', '11111111-1111-1111-1111-111111111111');
create database powersync;
```

`sync.yaml`:

```yaml
config:
  edition: 3
streams:
  profiles:
    auto_subscribe: true
    query: |
      SELECT id, name FROM profiles
      WHERE id = auth.user_id()
         OR id IN (SELECT member_id FROM memberships WHERE owner_id = auth.user_id())
```

`powersync.yaml`:

```yaml
replication:
  connections:
    - type: postgresql
      uri: postgresql://postgres:postgres@db:5432/postgres
      sslmode: disable
storage:
  type: postgresql
  uri: postgresql://postgres:postgres@db:5432/powersync
  sslmode: disable
port: 8080
sync_rules:
  path: /config/sync.yaml
client_auth:
  jwks:
    keys:
      - kty: oct
        alg: HS256
        kid: repro
        k: <base64url secret>
  audience: ["powersync"]
```

`compose.yml`:

```yaml
services:
  db:
    image: postgres:16
    command: ["postgres", "-c", "wal_level=logical"]
    environment: { POSTGRES_PASSWORD: postgres }
    volumes: ["./init.sql:/docker-entrypoint-initdb.d/init.sql:ro"]
    healthcheck: { test: ["CMD-SHELL", "pg_isready -U postgres"], interval: 2s, timeout: 3s, retries: 30 }
  powersync:
    image: journeyapps/powersync-service:1.26.1
    command: ["start", "-r", "unified"]
    depends_on: { db: { condition: service_healthy } }
    environment: { POWERSYNC_CONFIG_PATH: /config/powersync.yaml }
    volumes: ["./powersync.yaml:/config/powersync.yaml:ro", "./sync.yaml:/config/sync.yaml:ro"]
    ports: ["8082:8080"]
```

Request, with an HS256 token for `sub = 11111111-1111-1111-1111-111111111111`, `aud = powersync`:

```sh
curl -N -X POST -H "Authorization: Token $TOKEN" -H 'Content-Type: application/json' \
  --data '{"buckets":[],"include_checksum":true,"raw_data":true,"client_id":"repro","streams":{"include_defaults":true,"subscriptions":[]}}' \
  http://127.0.0.1:8082/sync/stream
```

### Observed (Postgres storage)

```json
{"checkpoint":{"last_op_id":"2","buckets":[{"bucket":"1#profiles|0[\"11111111-1111-1111-1111-111111111111\"]","checksum":-1933772616,"count":1,"priority":3,"subscriptions":[{"default":0}]}],"streams":[{"name":"profiles","is_default":true,"errors":[]}]}}
{"data":{"bucket":"1#profiles|0[\"11111111-1111-1111-1111-111111111111\"]","after":"0","has_more":false,"data":[
  {"op_id":"1","op":"PUT","object_type":"profiles","object_id":"11111111-1111-1111-1111-111111111111","checksum":2361194680,"subkey":"…","data":"{\"id\":\"11111111-1111-1111-1111-111111111111\",\"name\":\"alice\"}"},
  {"op_id":"1","op":"PUT","object_type":"profiles","object_id":"11111111-1111-1111-1111-111111111111","checksum":2361194680,"subkey":"…","data":"{\"id\":\"11111111-1111-1111-1111-111111111111\",\"name\":\"alice\"}"}
],"next_after":"1"}}
```

The checkpoint says `count: 1`; the data line carries `op_id 1` twice. `bucket_data` in the storage database holds the row once.

### Expected

One operation in the data line, as with `storage.type: mongodb` on the same source, config and request:

```json
{"data":{"bucket":"1#profiles|0[\"11111111-1111-1111-1111-111111111111\"]","after":"0","has_more":false,"data":[{"op_id":"1","op":"PUT", … }],"next_after":"1"}}
```

Removing either side of the `OR`, or the membership row that makes the subquery return the user's own id, also makes the duplicate disappear.

## Client impact

The Rust core supersedes the first copy when it inserts the second (same bucket and key), so the bucket ends up with `op_checksum = X` and `add_checksum = X`. `sync_local` then logs `Checksums didn't match, failed for: 1#profiles|0[…] (expected X, got 2X = X (op) + X (add))`, drops the bucket and reconnects; the next connection delivers the pair again, so the client never reaches a consistent checkpoint. In our app this happens for every user who owns a project, because the stream that syncs user profiles selects `user_id = auth.user_id() OR user_id IN (<members of the user's projects>)` and the second branch includes the user themself.

## Where it seems to come from

- `BucketChecksumState.getCheckpointUpdateDynamic` builds `const allBuckets = [...staticBuckets, ...mergeBuckets(dynamicBuckets)]` (https://github.com/powersync-ja/powersync-service/blob/v1.26.1/packages/service-core/src/sync/BucketChecksumState.ts#L750). Static and dynamic buckets are each merged on their own, but a bucket present in both lists stays twice.
- The checkpoint line dedupes through `bucketDescriptionMap` (a `Map` keyed by bucket name), but on a full sync `bucketsToFetch = allBuckets`, and `getFilteredBucketPositions` turns that into two `BucketDataRequest`s for the same bucket.
- `PostgresSyncRulesStorage.getBucketDataBatch` (https://github.com/powersync-ja/powersync-service/blob/v1.26.1/modules/module-postgres-storage/src/storage/PostgresSyncRulesStorage.ts#L337) puts every request into a `VALUES` list and `CROSS JOIN LATERAL`s `bucket_data` per request row, so each op comes back once per request; since both requests share the bucket name, the chunk builder appends both to the same `SyncBucketData`. The MongoDB storage filters with one `$or` over the requests, which returns each document once, hence the difference.

Deduplicating `allBuckets` by bucket name (merging `inclusion_reasons`) or deduplicating `dataBuckets` at the start of the Postgres `getBucketDataBatch` would both avoid it.

stevensJourney:
Fixes #813.

#721 significantly improved bucket data query performance for Postgres storage by using a `CROSS JOIN` over the supplied bucket requests. This works like a for-each loop, returning bucket operations for each request.

When a Sync Stream selects the same bucket through both static and dynamic parameters, service-core can produce duplicate bucket requests. For example:

    WHERE id = auth.user_id()
       OR id IN (SELECT member_id FROM memberships WHERE owner_id = auth.user_id())

The first branch selects a bucket directly from the authenticated user's ID (static). The second looks up membership data to determine which buckets to select (dynamic). If the user belongs to their own group, both branches select the same profile bucket.

With the updated Postgres query, those duplicate requests return the same operations twice, causing the client checksum failures reported in #813.

This PR fixes the issue in service-core by merging static and dynamic bucket descriptions together before fetching data. Each bucket is requested once, while preserving its subscription metadata and priority.

AI Usage: This fix was implemented with the help of Codex 6-Astra. Reviewed manually and with Claude Sonnet 5.


simolus3 (approved):

packages/service-core/src/sync/BucketChecksumState.ts
```diff
@@ -747,7 +747,8 @@ export class BucketParameterState {
         }
       }
     }
-    const allBuckets = [...staticBuckets, ...mergeBuckets(dynamicBuckets)];
+    // A stream can reach the same bucket through both static and dynamic parameters.
+    const allBuckets = mergeBuckets([...staticBuckets, ...dynamicBuckets]);
```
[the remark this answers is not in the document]
simolus3: fwiw the core extension can handle duplicate entries, but I agree it's better to de-duplicate.

packages/sync-rules/src/utils.ts
```diff
@@ -65,7 +65,17 @@ export function mergeBuckets(buckets: ResolvedBucket[]): ResolvedBucket[] {
     const existing = byBucketId[bucket.bucket];
 
     if (existing != null) {
-      existing.inclusion_reasons.push(...bucket.inclusion_reasons);
+      // A stream can reach the same bucket through more than one branch (e.g. overlapping
+      // subscriptions, or both a static and a dynamic parameter query) with the same inclusion
+      // reason each time, so de-duplicate as we merge.
+      const seenReasons = new Set(existing.inclusion_reasons.map((reason) => JSON.stringify(reason)));
```
simolus3: To me, this feels potentially expensive. I looked at ways to improve this and came up with this (using a hash map instead of an object for lookups, only cloning what we mutate and avoiding string allocations and comparisons):

```TypeScript
export function mergeBuckets(buckets: ResolvedBucket[]): ResolvedBucket[] {
  interface PendingResolvedBucket {
    resolved: ResolvedBucket;
    coveredByDefaultSubscription: boolean;
    explicitSubscriptionReasons: number[];
  }

  const pending = new Map<string, PendingResolvedBucket>();
  const allBuckets: ResolvedBucket[] = [];

  for (const bucket of buckets) {
    const existing = pending.get(bucket.bucket);
    let resolved: PendingResolvedBucket;

    if (existing != null) {
      resolved = existing;
      existing.resolved.priority = Math.min(existing.resolved.priority, bucket.priority) as BucketPriority;
    } else {
      // Clone so that we can modify the merged value without affecting the input value
      const clone: ResolvedBucket = withBucketSource(
        {
          definition: bucket.definition,
          bucket: bucket.bucket,
          priority: bucket.priority,
          inclusion_reasons: [] // Set while iterating over all soures
        },
        bucket.source
      );
      resolved = {
        resolved: clone,
        coveredByDefaultSubscription: false,
        explicitSubscriptionReasons: []
      };

      allBuckets.push(clone);
      pending.set(bucket.bucket, resolved);
    }

    // A stream can reach the same bucket through more than one branch (e.g. overlapping
    // subscriptions, or both a static and a dynamic parameter query) with the same inclusion
    // reason each time, so de-duplicate as we merge.
    for (const reason of bucket.inclusion_reasons) {
      if (reason === 'default') {
        if (!resolved.coveredByDefaultSubscription) {
          resolved.coveredByDefaultSubscription = true;
          resolved.resolved.inclusion_reasons.push(reason);
        }
      } else {
        // explicitSubscriptionReasons is a small array in practice, use a linear scan instead of a set 
        if (!resolved.explicitSubscriptionReasons.includes(reason.subscription)) {
          resolved.explicitSubscriptionReasons.push(reason.subscription);
          resolved.resolved.inclusion_reasons.push(reason);
        }
      }
    }
  }

  return allBuckets;
}
```

I'm not 100% sure if this is worth the extra lines though, so I'm happy with this either way.

simolus3 (dismissed):

packages/sync-rules/src/utils.ts
stevensJourney: Fair enough. I also wondered about potentially optimizing the function, but initially thought it might not be worth it. These optimizations do assume the structure of the inclusion reasons will be constant over time - but I suppose an AI review will spot any issues if that does change in future for some reason. 

I've updated the method to use your suggestion :)

simolus3 (approved):
LGTM 👍 

Image

PR #815, opened 2026-09-22
Deprecate sync rules in code - phase 1

benitav:
Phase 1 contains mainly cosmetic/label updates with no changes to functionality aside from some updated messaging.

- Agent guidance: new "Sync Rules Deprecation" section in AGENTS.md with terminology, casing rules, and the list of names that must never be renamed.
- Legacy code: every export under packages/sync-rules/src/legacy/ carries an @deprecated tag, one message for legacy Sync Rules and one for the alpha Sync Streams implementation, both pointing to src/legacy/README.md.
- Docs and README: the sync-rules package README now documents Sync Streams first with legacy Sync Rules under a deprecated heading; internal docs under docs/ and the Convex README say "sync config" instead of "sync rules"; the legacy grammar page opens with a deprecation callout.
- Templates and examples: the local-dev and MySQL dev templates and the service-client validate() example use streams: with config: edition: 3.
- Casing: "Sync Streams" and "Sync Rules" are capitalized in comments, JSDoc, log messages and the too-many-buckets error breakdown; the replicator logs "No sync config configured".
- Stale messages fixed: the edition 1 and 2 parser says CTEs and multiple queries "require edition 3" instead of naming a non-existent option; internal errors no longer reference the removed HydratedSyncRules class; a CLI help typo and a stale type name in a comment are corrected.

🤖 Claude Code made the updates based on [this plan ](https://docs.google.com/document/d/16wBnfZkALkfn2JaSw3qaxMiSRIiB0K_OFPfrQa86V8k/edit?tab=t.0#heading=h.ujpr7pgf0nii)(internal doc) and generated the above summary. I manually reviewed and simplified a few things.


modules/module-mongodb-storage/src/storage/implementation/db.ts
```diff
@@ -81,6 +81,7 @@ export class PowerSyncMongo {
     this.bucket_data = db.collection('bucket_data');
     this.bucket_parameters = db.collection('bucket_parameters');
     this.op_id_sequence = db.collection('op_id_sequence');
+    // Persisted name, do not rename. See AGENTS.md backwards-compatibility.
```
simolus3: These comments are hard to understand without context (do not rename _to what_?)

```suggestion
    // Uses legacy terminology instead of "sync config" for backwards-compatibility
```

modules/module-mongodb-storage/src/storage/MongoBucketStorage.ts
```diff
@@ -378,9 +379,9 @@ export class MongoBucketStorage extends storage.BucketStorageFactory {
 
     const mapping =
       options.config.plan == null
-        ? // For legacy sync rules and streams, use the parsed config directly to create a mapping
+        ? // No serialized plan (legacy Sync Rules, or Sync Streams parsed in-process): use the parsed config directly to create a mapping
```
simolus3: > Sync Streams parsed in-process

These are legacy too, the old comment is correct.

modules/module-mysql/dev/config/sync_rules.yaml
```diff
@@ -1,10 +1,15 @@
...
-    data:
-      - SELECT * FROM lists
-      - SELECT * FROM todos
+config:
+  edition: 3
+
+streams:
+  lists:
+    auto_subscribe: true
+    query: SELECT * FROM lists
```
simolus3: Should we make this a single stream with two queries to preserve the exact partitioning?

packages/sync-rules/src/compiler/sqlite.ts
```diff
@@ -255,7 +255,7 @@ export class PostgresToSqlite {
         const left = this.translateNodeWithLocation(expr.left);
         const right = this.translateNodeWithLocation(expr.right);
         if (expr.op === 'LIKE') {
-          // We don't support LIKE in the old bucket definition system, and want to make sure we're clear about ICU,
+          // We don't support LIKE in legacy Sync Rules bucket definitions, and want to make sure we're clear about ICU,
           // case sensitivity and changing the escape character first. TODO: Support later.
```
simolus3: Now that Sync Streams are the default, this is no longer a good excuse :D

```suggestion
          // TODO: Support later. We want to make sure we're clear about ICU, case sensitivity and changing the escape character first. 
```

service/local-dev/sync-rules-template.yaml
```diff
@@ -1,7 +1,12 @@
...
+config:
+  edition: 3
+
+streams:
+  lists:
+    auto_subscribe: true
+    query: SELECT * FROM lists
+  todos:
+    auto_subscribe: true
+    query: SELECT * FROM todos
```
simolus3: We might also want to merge these

modules/module-mongodb-storage/src/storage/implementation/db.ts
simolus3: Could you update the other comments referencing agents.md as well? The rest of this LGTM

